PR #109, opened 2018-01-27
Feature/cognito credential provider

jeskew:
This PR adds a browser-specific credential provider based on the "Simplified Flow" described in [the Cognito developer guide](https://docs.aws.amazon.com/cognito/latest/developerguide/authentication-flow.html). The `@aws/credential-provider-cognito-identity` package provides two credential provider functions, one of which (`fromCognitoIdentity`) takes an identity ID and calls `GetCredentialsForIdentity`, while the other (`fromCognitoIdentityPool`) takes an identity pool ID, calls `GetId` on the first invocation, and then calls `fromCognitoIdentity`. Subsequent invocations of the latter do not re-invoke `GetId`.

Some differences with V2's [`AWS.CognitoIdentityCredentials`](https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/CognitoIdentityCredentials.html):

* Users are responsible for instantiating the Cognito client
* There is no support for Cognito's "Classic Flow," which involves calling a Cognito's `GetOpenIdToken` and then calling STS's `AssumeRoleWithWebIdentity`. Cognito identifies this in their developer guide as only relevant in a particular case, so we could offer a separate credential provider if we see any customer demand for that flow. There appears to be some overlap with the `CustomRoleARN` parameter added to `GetId`'s input shape, so I would want to talk to the service team or a PM about whether classic flow is still necessary.
* Values supplied for the `Logins` parameter object may be either strings or functions that return promises for strings. This allows customers to provide self-refreshing token provider functions to solve issues like https://github.com/aws/aws-sdk-js/issues/1685
* Caching is left as an exercise to the reader. We can wrap the provider with [`memoize`](https://github.com/aws/aws-sdk-js-staging/blob/f965045bac4a62605f5aa9aa305e5ab7d0b202d6/packages/property-provider/src/memoize.ts#L41) when it's passed to a client constructor, but customers who want to cache the credentials somewhere durable (e.g., [session storage](https://github.com/aws/aws-sdk-js/issues/1097)) should have access to the raw provider function. Before GA, we should check with AppSec to see if we can cache credentials in session storage directly.

### DIFF
--- a/packages/credential-provider-cognito-identity/src/IndexedDbStorage.ts
+++ b/packages/credential-provider-cognito-identity/src/IndexedDbStorage.ts
@@ -4,20 +4,62 @@ import { ProviderError } from '@aws/property-provider';
 const STORE_NAME = 'IdentityIds';
 
 export class IndexedDbStorage implements Storage {
-    private readonly db: Promise<IDBDatabase>;
-
     constructor(
-        dbName: string = 'aws:cognito-identity-ids',
-        version: number = 1
-    ) {
-        const openDbRequest = self.indexedDB.open(dbName, version);
-        this.db = new Promise((resolve, reject) => {
+        private readonly dbName: string = 'aws:cognito-identity-ids'
+    ) {}
+
+    getItem(key: string): Promise<string|null> {
+        return this.withObjectStore('readonly', store => {
+            const req = store.get(key);
+
+            return new Promise<string|null>(resolve => {
+                req.onerror = () => resolve(null);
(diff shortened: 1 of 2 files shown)


nihakue:
In aws sdk v3, the CognitoIdentityClient can only be constructed with valid credentials. Am I missing something here? It feels like a chicken and egg scenario for the federated identity use case (e.g. exchanging id_token for temporary credentials)

PR #152, opened 2018-11-22
Added missing "await" in README example

eladb:
Operation methods return a `Promise` since they are asynchronous.


chrisradek:
Thank you for the correction @eladb, good catch!

PR #163, opened 2018-12-10
Move jest, lerna and ts to devDependencies

trivikr:
Moving jest, lerna and typescript under devDependencies where they belong.

As described in Contributing.md, the following commands were successfully run:
* npm install
* npm run bootstrap
* npm test


trivikr:
Any update?

PR #189, opened 2019-01-28
Added pinpoint client and smoke test

c4k9c0:
In the smoke tests, is there a way to do multiple operations per test?

I'd like to create an app, and then add an event to that app.
I verified this worked by manually writing a test, but since the smoke tests are automatically generated, I did not include it. Would be happy to ad it back, or use the `smoke.json` file to generate it if possible


c4k9c0:
Closing in favor of a PR with a cleaner diff

PR #388, opened 2019-10-08
fix: do not append to undefined/null header values

Amplifiyer:
*Issue #, if available:* N/A

*Description of changes:*
Avoid appending to user agent header when the existing value is undefined or null to avoid getting header values as "undefined myUserAgent"

### DIFF
--- a/packages/util-user-agent-browser/src/index.ts
+++ b/packages/util-user-agent-browser/src/index.ts
@@ -15,5 +15,9 @@ export function appendToUserAgent(
   request: HttpRequest,
   userAgentPartial: string
 ): void {
-  request.headers["X-Amz-User-Agent"] += ` ${userAgentPartial}`;
+  if(request.headers["X-Amz-User-Agent"]) {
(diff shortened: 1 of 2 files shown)


packages/util-user-agent-browser/src/index.ts
```diff
@@ -15,5 +15,9 @@ export function appendToUserAgent(
   request: HttpRequest,
   userAgentPartial: string
 ): void {
-  request.headers["X-Amz-User-Agent"] += ` ${userAgentPartial}`;
+  if(request.headers["X-Amz-User-Agent"]) {
+    request.headers["X-Amz-User-Agent"] += ` ${userAgentPartial}`;
+  } else {
+    request.headers["X-Amz-User-Agent"] = `${userAgentPartial}`;
```
trivikr: Quotes are not required for assignment in else block
```suggestion
    request.headers["X-Amz-User-Agent"] = userAgentPartial;
```

PR #420, opened 2019-10-25
feat(signature-v4): add support to override the set of unsignableHeaders

ronak2121:
*Issue #, if available:*
Add support to override the set of unsignableHeaders; in cases where signing those headers (such as user-agent) would be desirable.

*Description of changes:*
Added a new set of signableHeaders, which is checked for inclusion before the specified header is removed for being an unsignable one.

### DIFF
--- a/packages/signature-v4/src/SignatureV4.spec.ts
+++ b/packages/signature-v4/src/SignatureV4.spec.ts
@@ -469,7 +469,7 @@ describe("SignatureV4", () => {
         }
       );
       expect(headers[AUTH_HEADER]).toMatch(
-        /^AWS4-HMAC-SHA256 Credential=foo\/20000101\/us-bar-1\/foo\/aws4_request, SignedHeaders=host;foo;user-agent;x-amz-content-sha256;x-amz-date, Signature=/
+        /^AWS4-HMAC-SHA256 Credential=foo\/20000101\/us-bar-1\/foo\/aws4_request, SignedHeaders=foo;host;user-agent;x-amz-content-sha256;x-amz-date, Signature=/
       );
     });
 
--- a/packages/signature-v4/src/SignatureV4.spec.ts
+++ b/packages/signature-v4/src/SignatureV4.spec.ts
@@ -452,6 +452,27 @@ describe("SignatureV4", () => {
       );
     });
 
+    it("should allow specifying custom signable headers to override custom and always unsignable ones", async () => {
+      const { headers } = await signer.sign(
+        {
+          ...minimalRequest,
+          headers: {
+            host: "foo.us-bar-1.amazonaws.com",
+            foo: "bar",
+            "user-agent": "baz"
+          }
+        },
+        {
+          signingDate: new Date("2000-01-01T00:00:00.000Z"),
+          unsignableHeaders: new Set(["foo"]),
+          signableHeaders: new Set(["foo", "user-agent"])
+        }
+      );
+      expect(headers[AUTH_HEADER]).toMatch(
+        /^AWS4-HMAC-SHA256 Credential=foo\/20000101\/us-bar-1\/foo\/aws4_request, SignedHeaders=host;foo;user-agent;x-amz-content-sha256;x-amz-date, Signature=/
+      );
+    });
+
     it("should support signing with asynchronously resolved credentials", async () => {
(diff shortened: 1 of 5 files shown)


trivikr:
Relevant discussions from internal issue tracker below.

Findings:
* signature-v4-node extends from signature-v4 ([code](https://github.com/aws/aws-sdk-js-v3/blob/226e54407c4979be935d44fb3afdc32fb4154683/packages/signature-v4-node/src/index.ts#L2-L7))
* presignRequest function is defined in SignatureV4.ts ([code](https://github.com/aws/aws-sdk-js-v3/blob/e6fa8b1949eb6bcf9050614e2bb0fad835505e8b/packages/signature-v4/src/SignatureV4.ts#L126))
* The header "X-Amz-SignedHeaders" is set in SignatureV4.ts ([code](https://github.com/aws/aws-sdk-js-v3/blob/e6fa8b1949eb6bcf9050614e2bb0fad835505e8b/packages/signature-v4/src/SignatureV4.ts#L162-L164))
* The method `getCanonicalHeaderList` sorts the headers and joins them with `;` ([code](https://github.com/aws/aws-sdk-js-v3/blob/e6fa8b1949eb6bcf9050614e2bb0fad835505e8b/packages/signature-v4/src/SignatureV4.ts#L352-L356))
* The method `getCanonicalHeaders` returns sign-able headers which are not proxy/sec headers ([code](https://github.com/aws/aws-sdk-js-v3/blob/04592b775d715c1f0313954a3f09a01ba5d229bd/packages/signature-v4/src/getCanonicalHeaders.ts#L11-L33))

Proposed accepted solution:
* Pass `"user-agent"` as a new option in RequestSigningArguments ([code](https://github.com/aws/aws-sdk-js-v3/blob/04592b775d715c1f0313954a3f09a01ba5d229bd/packages/types/src/signature.ts#L19-L26))?
* It can be a set, with a new parameter `signableHeaders` in SigningArguments.
```javascript
interface RequestSigningArguments extends SigningArguments {
  /**
   * A set of strings whose members represents headers that cannot be signed.
   * All headers in the provided request will have their names converted to
   * lower case and then checked for existence in the unsignableHeaders set.
   */
  unsignableHeaders?: Set<string>;

  /**
   * A set of strings whose members represents headers that should be signed.
   * All headers in the provided request will have their names converted to
   * lower case before signing.
   */
  signableHeaders?: Set<string>;
}
```

trivikr:
Other findings:
* user-agent is skipped by default because of [RFC 2616](https://tools.ietf.org/html/rfc2616)
* it's legal for a proxy to modify the user agent header, so default signing it is risky

This PR provides a way for customers to explicitly opt-in to signing the user agent header

packages/signature-v4/src/SignatureV4.spec.ts
```diff
@@ -452,6 +452,27 @@ describe("SignatureV4", () => {
...
+          }
+        },
+        {
+          signingDate: new Date("2000-01-01T00:00:00.000Z"),
+          unsignableHeaders: new Set(["foo"]),
+          signableHeaders: new Set(["foo", "user-agent"])
+        }
+      );
+      expect(headers[AUTH_HEADER]).toMatch(
+        /^AWS4-HMAC-SHA256 Credential=foo\/20000101\/us-bar-1\/foo\/aws4_request, SignedHeaders=host;foo;user-agent;x-amz-content-sha256;x-amz-date, Signature=/
```
trivikr: Edit to fix [CI failure](https://github.com/aws/aws-sdk-js-v3/pull/420#issuecomment-546516264):
```suggestion
        /^AWS4-HMAC-SHA256 Credential=foo\/20000101\/us-bar-1\/foo\/aws4_request, SignedHeaders=foo;host;user-agent;x-amz-content-sha256;x-amz-date, Signature=/
```

trivikr: I'm not able to commit this suggestion to repository.
* It could be because PR was created from master branch?
* The option ["allow edits from maintainers"](https://github.blog/2016-09-07-improving-collaboration-with-forks/) should have been set by default

Screen Shot 2019-10-28 at 12 06 47 PM

Screen Shot 2019-10-28 at 12 06 50 PM

PR #445, opened 2019-11-13
fix: bump smithy dependency to 0.9.5

srchase:
bumps smithy dependency to 0.9.5 to match smithy-typescript.


srchase:
keep in sync with https://github.com/awslabs/smithy-typescript/pull/21

PR #448, opened 2019-11-14
chore: use default with destructuring in resolveRetryConfig

trivikr:
*Issue #, if available:*
N/A

*Description of changes:*
* Example code on TypeScript Playground: https://bit.ly/32PawhS
* docs: https://basarat.gitbooks.io/typescript/docs/destructuring.html


trivikr:
The build is failing with following error:
```console
src/configurations.ts(23,52): error TS2454: Variable 'maxRetries' is used before being assigned.
```

Not sure what the issue is, as TypeScript playground transpiles it correctly
Code: https://bit.ly/376TUWo

Closing as this is not an important change, and was done while working on clock skew

PR #483, opened 2019-12-02
AddGlacierApiVersion middleware migration

srchase:
migrates and applies AddGlacierApiVersion middleware

For now, this hard-codes the api version.  This will need to be updated with a way to dynamically pull in the version for the service.


kstich:
I think this PR includes a chunk of commits that are already on the target branch of the repo.

PR #569, opened 2019-12-19
fix: merge plugin predicates

srchase:
The servicePredicate is automatically set to false when the operationPredicate is used: https://github.com/awslabs/smithy-typescript/blob/5dffb8a0060023af4c0e1f809d116ab77492cdde/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/integration/RuntimeClientPlugin.java#L466

This merges the ServiceId check into the operationPredicate to only apply these plugins to specific operations on specific services.


srchase:
Rebased.  Additional plugin added in #567, HostHeader, didn't have a predicate, so no need to fix that one.

PR #582, opened 2019-12-27
fix: apply header only in node environment

srchase:
Only set 100-continue header in node environment.

Uses runtime config added in https://github.com/awslabs/smithy-typescript/pull/64

### DIFF
--- a/packages/middleware-expect-continue/src/index.ts
+++ b/packages/middleware-expect-continue/src/index.ts
@@ -9,14 +9,24 @@ import {
 } from "@aws-sdk/types";
 import { HttpRequest } from "@aws-sdk/protocol-http";
 
-export function addExpectContinueMiddleware(): BuildMiddleware<any, any> {
+interface PreviouslyResolved {
+  runtime: string;
+}
+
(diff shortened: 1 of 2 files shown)


packages/middleware-expect-continue/src/index.ts
```diff
@@ -9,14 +9,24 @@ import {
 } from "@aws-sdk/types";
 import { HttpRequest } from "@aws-sdk/protocol-http";
 
-export function addExpectContinueMiddleware(): BuildMiddleware<any, any> {
+interface PreviouslyResolved {
+  runtime: string;
```
trivikr: use `RuntimeTarget` instead of string
```suggestion
  runtime: RuntimeTarget;
```

where:
```typescript
import {
  RuntimeTarget
} from "@aws-sdk/build-types";
```

We can change Runtime target to [Enum](https://www.typescriptlang.org/docs/handbook/enums.html) if required, in a separate PR

srchase: @aws-sdk/build-types is probably going away, so I'll leave it as-is for now.

PR #641, opened 2020-01-06
fix(client-api-gateway): remove forward slash after *

trivikr:
*Issue #, if available:*
Temporary fix for https://github.com/awslabs/smithy-typescript/issues/74

*Description of changes:*
fix(client-api-gateway): remove backslash after *


trivikr:
This fix will be overwritten by permanent one, when https://github.com/awslabs/smithy-typescript/issues/74 is fixed and CodeGen is run

PR #730, opened 2020-01-14
fix: xml map serializer valueNode

srchase:
Fixes awslabs/smithy-typescript#97

Use valueNode instead of keyNode for value's node.


srchase:
This change is needed for https://github.com/awslabs/smithy-typescript/issues/97

PR #774, opened 2020-01-22
feat: add isValidHostname function 

srchase:
Adds validHostname function for checking hostname validity.  To be used for host prefixing. Ref: awslabs/smithy-typescript/pull/104


packages/util-valid-hostname/src/index.ts
```diff
@@ -0,0 +1,4 @@
+export function validHostname(hostname: string): boolean {
+  const hostPattern = /^[a-zA-Z0-9]{1}$|^[a-zA-Z0-9][a-zA-Z0-9\-]*[a-zA-Z0-9]$/;
```
mtdowling: Shouldn't you compile this once and then reuse the regex?

srchase: Using the literal (vs the RegExp constructor) should cache the regex:
```
const { validHostname } = require("/aws-sdk-js-v3/packages/util-valid-hostname");

console.time('1');validHostname("test1");console.timeEnd('1');
console.time('2');validHostname("test2");console.timeEnd('2');
```
Node returns:
```
1: 0.262ms
2: 0.011ms
```

srchase: Remove standalone package and added function to protocol-http.

packages/protocol-http/src/validHostname.ts
```diff
@@ -0,0 +1,4 @@
+export function validHostname(hostname: string): boolean {
```
trivikr: use `isValidHostname`
```suggestion
export function isValidHostname(hostname: string): boolean {
```

StackOverflow discussion https://softwareengineering.stackexchange.com/questions/348649

srchase: Fixed.

Issue #487, opened 2019-12-03
Node Http2 Handler Spec fails intermittently


srchase:
The abortSignal test intermittently fails CI:

```
PASS  packages/fetch-http-handler/build/fetch-http-handler.spec.js

 FAIL  packages/node-http-handler/build/node-http2-handler.spec.js

  ● NodeHttp2Handler › abortSignal › will close request on session when aborted

    expect(received).rejects.toHaveProperty()

    Received promise resolved instead of rejected
```

smithy-codegen as of b831fe89baf344c2a9cfdadb74147892633a1165

**Is the issue in the browser/Node.js?**
NA: CI

trivikr:
This happened only in Travis

The commit to remove TravisCI was cherry-picked in `smithy-codegen` branch https://github.com/aws/aws-sdk-js-v3/commit/d37c4b3041e644ab0aacf323f429637bb7e714d2

This issue can be resolved if issue doesn't reproduce in `smithy-codegen`

trivikr:
Resolving as the issue hasn't reproduced in `smithy-codegen`

trivikr:
Was closed by mistake

Reopening as the issue did reproduce in `smithy-codegen` in severals PRs.
Latest one https://github.com/aws/aws-sdk-js-v3/pull/782#issuecomment-577922459

trivikr:
Flaky test in https://github.com/aws/aws-sdk-js-v3/pull/787#issuecomment-578260875

ajredniwja:
@srchase is this still a persisting issue?

PR #790, opened 2020-01-24
feat: rename calculateSha256 functions

srchase:
Renames calculateSha256 to fileStreamHasher and blobHasher.  Each function takes a Hash constructor and thus can return other hashes beyond Sha256.

Updates S3 Client where these functions are currently used.


srchase:
CI failure is due to flaky test, and is unrelated #487

PR #814, opened 2020-01-29
feat: Auto-fill idempotency token values

kstich:
This commit adds support for generating idempotency token values
when not set on input. These values are populated with a UUID v4
value provided by the `uuid` package.

This is done through setting the value on the provided input as to
not further complicate protocol specific member serialization.

-----

Related to awslabs/smithy-typescript#109

-----

Sample codegen:

```
import { v4 as generateIdempotencyToken } from "uuid";

if (input.ClientRequestToken === undefined) {
    input.ClientRequestToken = generateIdempotencyToken();
}
```

-----

### DIFF
--- a/codegen/smithy-aws-typescript-codegen/src/main/java/software/amazon/smithy/aws/typescript/codegen/AwsProtocolUtils.java
+++ b/codegen/smithy-aws-typescript-codegen/src/main/java/software/amazon/smithy/aws/typescript/codegen/AwsProtocolUtils.java
@@ -21,9 +21,11 @@ import java.util.TreeSet;
 import software.amazon.smithy.aws.traits.UnsignedPayloadTrait;
 import software.amazon.smithy.model.knowledge.NeighborProviderIndex;
 import software.amazon.smithy.model.neighbor.Walker;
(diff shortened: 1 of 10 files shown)


codegen/smithy-aws-typescript-codegen/src/main/java/software/amazon/smithy/aws/typescript/codegen/AwsProtocolUtils.java
```diff
@@ -156,4 +164,40 @@ static void writeXmlNamespace(GenerationContext context, Shape shape, String nod
             writer.write("$L.addAttribute($S, $S);", nodeName, xmlns, trait.getUri());
         });
     }
+
+    /**
+     * Imports an UUID v4 generating function used for auto-filling idempotency tokens.
+     *
+     * @param context The generation context.
+     */
+    static void generateIdempotencyAutofiller(GenerationContext context) {
```
mtdowling: This method name was confusing to me -- it's not actually generating anything. It's just adding dependencies if needed. Can this name be updated to reflect that?

Issue #763, opened 2020-01-20
The e.Error.Message must be in e.message


trivikr:
**Describe the bug**
The e.Error.Message must be in e.message

**SDK version number**
```console
$ yarn list aws-sdk
└─ aws-sdk@2.605.0
$ yarn list @aws-sdk/client-elastic-beanstalk
└─ @aws-sdk/client-elastic-beanstalk@1.0.0-alpha.2
```

**Is the issue in the browser/Node.js?**
Node.js

**Details of the browser/Node.js version**
```console
$ node -v
v12.14.0
```

**To Reproduce (observed behavior)**
Reproduced using the following code:
```javascript
const AWS = require("aws-sdk");
const { ElasticBeanstalk } = require("@aws-sdk/client-elastic-beanstalk");

(async () => {
  const region = "us-west-2";
  const params = {
    ApplicationName: ""
  };

  const v2Client = new AWS.ElasticBeanstalk({ region });
  try {
    await v2Client.createApplication(params).promise();
  } catch (e) {
    console.log("Error returned by v2:");
    console.log(JSON.stringify(e, null, 2));
  }

  const v3Client = new ElasticBeanstalk({ region });
  try {
    await v3Client.createApplication(params);
  } catch (e) {
    console.log("\nError returned by v3:");
    console.log(JSON.stringify(e, null, 2));
  }
})();
```

Output shows that in v3, there's an additional Error object inside existing Error object
```
Error returned by v2:
{
  "message": "2 validation errors detected: Value '' at 'applicationName' failed to satisfy constraint: Member must satisfy regular expression pattern: [^/]+; Value '' at 'applicationName' failed to satisfy constraint: Member must have length greater than or equal to 1",
  "code": "ValidationError",
  "time": "2020-01-20T23:42:44.536Z",
  "requestId": "f3ca4ab8-896b-47db-937a-961df2986b5d",
  "statusCode": 400,
  "retryable": false,
  "retryDelay": 15.387417546477877
}

Error returned by v3:
{
  "xmlns": "http://elasticbeanstalk.amazonaws.com/docs/2010-12-01/",
  "Error": {
    "Type": "Sender",
    "Code": "ValidationError",
    "Message": "2 validation errors detected: Value '' at 'applicationName' failed to satisfy constraint: Member must satisfy
 regular expression pattern: [^/]+; Value '' at 'applicationName' failed to satisfy constraint: Member must have length great
er than or equal to 1"
  },
[...]

PR #842, opened 2020-02-04
fix: Update AWS protocol error body locations

kstich:
This commit updates some AWS protocols that wrap their error data
in additional elements.

Depends on awslabs/smithy-typescript#117
Fix for #763 

-----

Modeled error sample:

```
const deserializeAws_queryDBClusterEndpointNotFoundFaultResponse = async (
  parsedOutput: any,
  context: __SerdeContext
): Promise<DBClusterEndpointNotFoundFault> => {
  const body = parsedOutput.body
  const deserialized: any = deserializeAws_queryDBClusterEndpointNotFoundFault(body.Error, context); // <-- Note body.Error
  const contents: DBClusterEndpointNotFoundFault = {
    name: "DBClusterEndpointNotFoundFault",
    $fault: "client",
    $metadata: deserializeMetadata(parsedOutput),
    ...deserialized,
  };
  return contents;
};
```

Unmodeled error sample:

```
    default:
      const parsedBody = parsedOutput.body;
      errorCode = errorCode || "UnknownError";
      response = {
        ...parsedBody.Error,
        name: `${errorCode}`,
        message: parsedBody.Error.message || parsedBody.Error.Message || errorCode,
        $fault: "client",
        $metadata: deserializeMetadata(output)
      } as any;
```

-----

### DIFF
--- a/codegen/smithy-aws-typescript-codegen/src/main/java/software/amazon/smithy/aws/typescript/codegen/AwsEc2.java
+++ b/codegen/smithy-aws-typescript-codegen/src/main/java/software/amazon/smithy/aws/typescript/codegen/AwsEc2.java
@@ -87,10 +87,10 @@ final class AwsEc2 extends HttpRpcProtocolGenerator {
                        + "  output: $T,\n"
                        + "  data: any\n"
                        + "): string => {", "};", responseType, () -> {
-
-            // Attempt to fetch the error code from the specific location, including the wrapper.
-            writer.openBlock("if (data.Errors.Error.Code !== undefined) {", "}", () -> {
-                writer.write("return data.Errors.Error.Code;");
+            // Attempt to fetch the error code from the specific location.
+            String errorCodeLocation = getErrorBodyLocation(context, "data") + ".Code";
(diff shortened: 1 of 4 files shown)


codegen/smithy-aws-typescript-codegen/src/main/java/software/amazon/smithy/aws/typescript/codegen/AwsEc2.java
```diff
@@ -87,10 +87,10 @@ public void generateSharedComponents(GenerationContext context) {
                        + "  output: $T,\n"
                        + "  data: any\n"
                        + "): string => {", "};", responseType, () -> {
-
-            // Attempt to fetch the error code from the specific location, including the wrapper.
-            writer.openBlock("if (data.Errors.Error.Code !== undefined) {", "}", () -> {
-                writer.write("return data.Errors.Error.Code;");
+            // Attempt to fetch the error code from the specific location.
```
mtdowling: I think we should take a look at moving as much code as possible into library functions written in TypeScript. These kinds of methods seem like they could be generic methods that are HTTP-specific.

Issue #871, opened 2020-02-06
Lex API: TypeError: Cannot read property 'sentimentLabel' of null


srchase:
**Describe the bug**
Lex API returns `null`, causing deserialization errors:

`TypeError: Cannot read property 'sentimentLabel' of null`

**SDK version number**
1.0.0-alpha.13

**Is the issue in the browser/Node.js?**
Browser and Node.js

**Details of the browser/Node.js version**
all

**To Reproduce (observed behavior)**
```
const { LexRuntimeService } = require("@aws-sdk/client-cognito-sync");

(async () => {
    const params = {
        botAlias: "beta",
        botName: "BookTrip",
        inputText: "Book My Hotel",
        userId: "fooUserId"
    };
    const client = new LexRuntimeService({ region: 'us-east-1' });
  
    try {
        const result = await client.postText(params);
        console.log('success result: ', result);
    } catch (e) {
        console.log('error: ', e);
        process.exit(1);
    }
})();
```

**Expected behavior**
Returns postText response

PR #872, opened 2020-02-06
fix: check null locations

srchase:
Fixes: https://github.com/aws/aws-sdk-js-v3/issues/871

Need to check that locations are not null before attempting to deserialize.

### DIFF
--- a/codegen/smithy-aws-typescript-codegen/src/main/java/software/amazon/smithy/aws/typescript/codegen/JsonShapeDeserVisitor.java
+++ b/codegen/smithy-aws-typescript-codegen/src/main/java/software/amazon/smithy/aws/typescript/codegen/JsonShapeDeserVisitor.java
@@ -104,7 +104,8 @@ final class JsonShapeDeserVisitor extends DocumentShapeDeserVisitor {
             Shape target = context.getModel().expectShape(memberShape.getTarget());
 
             // Generate an if statement to set the bodyParam if the member is set.
-            writer.openBlock("if (output.$L !== undefined) {", "}", locationName, () -> {
+            writer.openBlock("if (output.$L !== undefined && output.$L !== null) {", "}", locationName, locationName,
+                    () -> {
                 writer.write("contents.$L = $L;", memberName,
                         // Dispatch to the output value provider for any additional handling.
                         target.accept(getMemberVisitor("output." + locationName)));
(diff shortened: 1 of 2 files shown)


srchase:
With this fix, responses are deserialized like this:
```
{
  '$metadata': {
    httpStatusCode: 200,
    httpHeaders: {
      'content-type': 'application/json',
      date: 'Thu, 06 Feb 2020 17:26:11 GMT',
      'x-amzn-requestid': '...',
      'content-length': '372',
      connection: 'keep-alive'
    },
    requestId: '...',
    retries: 0,
    totalRetryDelay: 0
  },
  __type: 'PostTextResponse',
  dialogState: 'ElicitSlot',
  intentName: 'BookHotel',
  message: 'What city will you be staying in?',
  messageFormat: 'PlainText',
  responseCard: undefined,
  sentimentResponse: undefined,
  sessionAttributes: undefined,
  sessionId: '2020-02-06T17:22:12.905Z-LzBxvqUs',
  slotToElicit: 'Location',
  slots: { CheckInDate: null, Location: null, Nights: null, RoomType: null }
```

Does it make sense to mix `null` and `undefined`?

kstich:
Do the non-structure member targets need this treatment as well?

From the above:
```
 slots: { CheckInDate: null, Location: null, Nights: null, RoomType: null }
```

trivikr:
> Does it make sense to mix `null` and `undefined`?

Yes, it makes sense to add both `null` and `undefined` checks before deserializing.

TLDR version:
* The value null represents the intentional absence of any object value ([MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/null))
* undefined is a primitive value automatically assigned to variables that have just been declared, or to formal arguments for which there are no actual arguments. ([MDN](https://developer.mozilla.org/en-US/docs/Glossary/Undefined))
* null is an empty or non-existent value, where as undefined typically means a variable has been declared, but not defined. Both null and undefined are primitive values, as well as falsy.

srchase:
```
sentimentResponse: undefined,
sessionAttributes: undefined,
...
slots: { CheckInDate: null, Location: null, Nights: null, RoomType: null }
```

The service explicitly returns `null` values for the `slots` string to string map.  This change results in the correct behavior.

PR #883, opened 2020-02-07
fix: remove duplicate signingName config

srchase:
Currently, the same `signingName` is written to `runtimeConfig.ts` `runtimeConfig.browser.ts`, and `runtimeconfig.shared.ts`.

Since this is the same across runtimes, it only needs to be written to the `runtimeConfig.shared.ts`.

### DIFF
--- a/codegen/smithy-aws-typescript-codegen/src/main/java/software/amazon/smithy/aws/typescript/codegen/AddAwsRuntimeConfig.java
+++ b/codegen/smithy-aws-typescript-codegen/src/main/java/software/amazon/smithy/aws/typescript/codegen/AddAwsRuntimeConfig.java
@@ -95,15 +95,16 @@ public final class AddAwsRuntimeConfig implements TypeScriptIntegration {
             LanguageTarget target
     ) {
         ServiceShape service = settings.getService(model);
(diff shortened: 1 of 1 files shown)


codegen/smithy-aws-typescript-codegen/src/main/java/software/amazon/smithy/aws/typescript/codegen/AddAwsRuntimeConfig.java
```diff
@@ -99,7 +99,7 @@ public void addRuntimeConfigValues(
                 .map(ServiceTrait::getArnNamespace)
                 .orElse(null);
 
-        if (signingName != null) {
+        if (signingName != null && target.equals(LanguageTarget.SHARED)) {
```
kstich: This target check should be a wrapper for the current `if/else`, or there will be log lines for other targets.

srchase: Fixed in 694083cbd.  Moved target check up to wrap entire `signingName` section.

Issue #906, opened 2020-02-10
CodeGen updates in #900 changes the order of dependencies


trivikr:
**Is your feature request related to a problem? Please describe.**
* CodeGen updates in #900 changes the order of dependencies
* This was noticed in commit https://github.com/aws/aws-sdk-js-v3/pull/903/commits/09659dad03a38b0c15665b3279db8fa402789382 of #903

**Describe the solution you'd like**
CodeGen retains the order of dependencies/devDependencies same as that done by lerna

**Describe alternatives you've considered**
changing the order after running codegen as in #900, and it being updated when release is run

trivikr:
Fixed in https://github.com/aws/aws-sdk-js-v3/pull/909

PR #909, opened 2020-02-11
fix: sort deps/devDeps by keys

trivikr:
*Issue #, if available:*
Fixes: https://github.com/aws/aws-sdk-js-v3/issues/906

*Description of changes:*
* sort deps/devDeps by keys
* verified that the changes in https://github.com/aws/aws-sdk-js-v3/commit/09659dad03a38b0c15665b3279db8fa402789382 were undone after this code change, will verify again by running codegen after lerna releases `v1.0.0-alpha.21`


trivikr:
After this fix, codegen needs to be run with Node.js 12+ as `Object.fromEntries` is ES2019 https://tc39.es/proposal-object-from-entries/

trivikr:
Verified after `v1.0.0-alpha.21` release that ordering maintained by this PR is same as that maintained by lerna

PR #915, opened 2020-02-11
feat: extended URI encoding

srchase:
Fixes: #866

Adds extendedEncodeURI function to wrap encodeURIComponent and encode additional characters, !, ', (, ), and *, to adhere to [RFC 3986](https://tools.ietf.org/html/rfc3986).


packages/smithy-client/src/extended-encode-uri.ts
```diff
@@ -0,0 +1,5 @@
+export function extendedEncodeURI(str: string): string {
```
kstich: Can you add a comment explaining why this is necessary (RFC compliance)?

srchase: fixed in 7b93183ab

PR #914, opened 2020-02-11
chore: codegen for return empty instead of undefined

trivikr:
*Issue #, if available:*
Refs: https://github.com/aws/aws-sdk-js-v3/pull/885

*Description of changes:*
codegen for return empty instead of undefined

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #919, opened 2020-02-11
fix: handle flattened xml with self-closed tag

srchase:
Fix for CI failure in https://github.com/aws/aws-sdk-js-v3/pull/914 by using correct location for flattened XML.

`CloudSearch.describeDomains` returns an empty `array` for `DomainStatusList`.  `SQS.listQueues` still returns `undefined` for `QueueUrls` since the  `QueueUrl` is not set in the XML.

### DIFF
--- a/codegen/smithy-aws-typescript-codegen/src/main/java/software/amazon/smithy/aws/typescript/codegen/XmlShapeDeserVisitor.java
+++ b/codegen/smithy-aws-typescript-codegen/src/main/java/software/amazon/smithy/aws/typescript/codegen/XmlShapeDeserVisitor.java
@@ -164,7 +164,7 @@ final class XmlShapeDeserVisitor extends DocumentShapeDeserVisitor {
 
         // Handle self-closed xml parsed as an empty string.
         if (deserializationReturnsArray) {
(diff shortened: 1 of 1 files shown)


codegen/smithy-aws-typescript-codegen/src/main/java/software/amazon/smithy/aws/typescript/codegen/XmlShapeDeserVisitor.java
```diff
@@ -164,16 +164,17 @@ void deserializeNamedMember(
 
         // Handle self-closed xml parsed as an empty string.
         if (deserializationReturnsArray) {
-            writer.openBlock("if ($L.$L === \"\") {", "}", inputLocation, locationName, () -> {
+            writer.openBlock("if ($L['$L'] === \"\") {", "}", inputLocation, locationName, () -> {
```
kstich: `'$L'` should just be `$S` if this change is necessary.

srchase: Not needed.  Reverted in 593308c84.

PR #1077, opened 2020-04-03
fix: XML Namespace Serde issues

kstich:
### Handle extra attrs on value nodes

This handles receiving xmlns attributes in response where that information is unused.

### Update protocol tests with codegen

-----


codegen/smithy-aws-typescript-codegen/src/main/java/software/amazon/smithy/aws/typescript/codegen/XmlShapeSerVisitor.java
```diff
@@ -115,8 +121,10 @@ protected void serializeMap(GenerationContext context, MapShape shape) {
             String keyName = keyMember.getTrait(XmlNameTrait.class)
                     .map(XmlNameTrait::getValue)
                     .orElse("key");
-            writer.write("entryNode.addChildNode($L.withName($S));",
-                    keyTarget.accept(getMemberVisitor("key")), keyName);
+            writer.write("const keyNode = $L.withName($S);", keyTarget.accept(getMemberVisitor("key")), keyName);
+            // Add @xmlNamespace value of the key member.
+            AwsProtocolUtils.writeXmlNamespace(context, keyMember, "workingNode");
```
kuhe, line 126: is this supposed to be keyNode?

PR #1145, opened 2020-05-04
chore: remove redundant return statements from protocol.ts

trivikr:
*Issue #, if available:*
N/A

*Description of changes:*
remove redundant return statements from protocol.ts


trivikr:
This PR was merged as a follow-up to https://github.com/awslabs/smithy-typescript/pull/176, instead of merging https://github.com/aws/aws-sdk-js-v3/pull/1149

But the two PRs are similar, so won't be reverting

PR #1279, opened 2020-06-16
example config to be propagated across packages

alexforsyth:
*Issue #, if available:*

*Description of changes:*

Experimental PR to show how we'd implement a root tsconfig. This works properly for the abort controller package


alexforsyth:
Key points:
1. There is now a root ts-config, we'll put all the common ts options there. 
2. Individual packages no longer have a tsconfig. This is because vscode looks for the nearest tsconfig in the tree. By removing it, vscode jumps to using the root tscofig and provides all sorts of awesome goodies like paths and automatically highlighting places that do not adhere to the root tsconfig options.
3. Individual packages now have a tsconfig.build.json. This is a renamed tsconfig.test.json that inherits from the root tsconfig. Used by lerna when we want to build tsc.

tsconfig.json
```diff
@@ -0,0 +1,32 @@
...
+     * Nice to have:
+     */
+    // "preserveConstEnums": true,
+    // "declaration": true,
+    // "allowUnreachableCode": false,
+    // "pretty": true,
+
+    "module": "commonjs",
+    "moduleResolution": "node",
+    "target": "es5",
```
trivikr, line 22: target should be browser TSConfig target, as that's usually the lower version of ECMAScript.

tsconfig.json
```diff
@@ -0,0 +1,32 @@
...
+    "lib": ["es2015", "dom"],
```
trivikr, line 23: decided to exclude lib, as it's overridden in individual clients

PR #1280, opened 2020-06-17
test(region-provider): add extensive unit tests for region-provider

trivikr:
*Issue #, if available:*
Refactored while investigating Retry Configuration task (internal JS-1813)

*Description of changes:*
add unit tests for region-provider

### DIFF
--- a/packages/region-provider/src/defaultProvider.ts
+++ b/packages/region-provider/src/defaultProvider.ts
@@ -8,10 +8,7 @@ import { Provider } from "@aws-sdk/types";
 
 export type RegionProviderConfiguration = EnvConfiguration & SharedConfigInit;
 
-export function defaultProvider(
+export const defaultProvider = (
   configuration: RegionProviderConfiguration = {}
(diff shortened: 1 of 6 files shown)


packages/region-provider/src/defaultProvider.ts
```diff
@@ -8,10 +8,7 @@ import { Provider } from "@aws-sdk/types";
 
 export type RegionProviderConfiguration = EnvConfiguration & SharedConfigInit;
 
-export function defaultProvider(
+export const defaultProvider = (
   configuration: RegionProviderConfiguration = {}
-): Provider<string> {
-  return memoize(
```
alexforsyth, line 14: QQ: Why did the formatting change here?

trivikr: It's now an arrow function.
We're slowly switching to arrow functions for no binding of `this` https://www.freecodecamp.org/news/3d851d7f0b26/

PR #1282, opened 2020-06-18
feat: add retry-config-provider

trivikr:
*Issue #, if available:*
Internal issue at JS-1813

*Description of changes:*

adds retry-config-provider to read values from environment or config file:

|                     | Env Var          | Config File  |
|---------------------|------------------|--------------|
| maxAttemptsProvider | AWS_MAX_ATTEMPTS | max_attempts |
| retryModeProvider   | AWS_RETRY_MODE   | retry_mode   |

Similar to [region-provider](https://github.com/aws/aws-sdk-js-v3/tree/master/packages/region-provider) but with following differences:
* the defaultProvider passes the values to fromEnv and fromSharedConfigFiles as there are two methods to export
* fromSharedConfigFiles doesn't read for values from credentials file.
* doesn't export fromEnv and fromSharedConfigFiles as they're not required by external code.


alexforsyth:
Can you verify that its a bug that the tests impact packages/middleware-retry/src/defaultStrategy.ts, and that the unit tests do not actually impact it?

packages/retry-config-provider/src/fromSharedConfigFiles.ts
```diff
@@ -0,0 +1,43 @@
...
+  init: SharedConfigInit = {},
+  configKey: string
+): Provider<string> => async () => {
+  const {
+    loadedConfig = loadSharedConfigFiles(init),
+    profile = process.env[ENV_PROFILE] || DEFAULT_PROFILE
+  } = init;
+
+  const { configFile } = await loadedConfig;
+  const { [configKey]: configValue } = configFile[profile] || <any>{};
```
alexforsyth, line 35: dont neeed this any

PR #1285, opened 2020-06-19
chore: faster build time

alexforsyth:
*Issue #, if available:*

*Description of changes:*
Removed --include-dependency to speed up build time. 

Some stats:

Before (clean build, git clean -fxd && yarn)
```
real    46m50.258s
user    57m11.852s
```

After (clean build, git clean -fxd && yarn)
```
real    5m35.860s
user   6m6.196s
```


package.json
```diff
@@ -73,4 +73,4 @@
   "lint-staged": {
     "**/*.{ts,js,md,json}": "prettier --write"
   }
-}
```
alexforsyth, line 76: Not sure what happened here, doesnt show up on the diff in vscode

Issue #1026, opened 2020-03-20
Use compilerOptions.paths to resolve @aws-sdk/ dependencies


trivikr:
**Is your feature request related to a problem? Please describe.**
@aws-sdk dependencies are not navigable in a fresh clone as described in https://github.com/aws/aws-sdk-js-v3/issues/329

**Describe the solution you'd like**
Use compilerOptions.paths to resolve @aws-sdk/ dependencies
* [Module resolution documentation](https://www.typescriptlang.org/docs/handbook/module-resolution.html#path-mapping)
* [Blog post](https://medium.com/@NiGhTTraX/c6acda7d4559#56cf)

**Describe alternatives you've considered**
Running yarn test:all or just the dependencies using lerna --scope command in https://github.com/aws/aws-sdk-js-v3/issues/329#issuecomment-523991081

trivikr:
There's an open issue with TypeScript team to infer project references from common monorepo patterns/tools https://github.com/microsoft/TypeScript/issues/25376

Right now, the dependencies have to be defined in package.json and references need to be defined in tsconfig.json

trivikr:
Fixed in https://github.com/aws/aws-sdk-js-v3/pull/1297

PR #1297, opened 2020-06-25
feat: use a common tsconfig for the monorepo

alexforsyth:
*Issue #, if available:*
https://github.com/aws/aws-sdk-js-v3/issues/1026
https://i.amazon.com/issues/JS-1866

*Description of changes:*
1.  Created a root `tsconfig.json` and root `tsconfig.cjs.json` and `tsconfig.es.json`.
2. Each package extends the root `tsconfig.cjs.json` and root `tsconfig.es.json` in their local version. Those are themselves built from the historical `tsconfig.json + tsconfig.test.json`. The previous `packages/*/tsconfig.json` and `packages/*/tsconfig.test.json` have been removed.
3. Each package also now generates two builds, one common (cjs) in `./build` and one esnext (es) in `./build-es`. Packages have also been updated to provide a `build:es` and a `build:cjs` command.
4. Because we are now generating two builds the time to build should approximately double. This should not be too bad, as build time was greatly reduced in the previous PR: https://github.com/aws/aws-sdk-js-v3/pull/1285

FAQ:
1. **Why not have each package have its own tsconfig?** Its been a huge pain and contributes to non standard packages if they do not all adhere to the same typescript standards. We've chosen to have a rot tsconfig.
1. **Okay, but why get rid of tsconfig in each package?** Vscode looks at the closest tsconfig.json in the tree. In order for vscode to profile the repo correctly and code competition to work well, we need to have vscode profile stuff from the root, that means we need to keep the tsconfig there and remove it from closer places in the code tree. 
1. **It looks like a lot of the packages share many attributes in their tsconfig.(cjs|es).json, is it possible to pull those options out?**  Yes - there are tons of shared options, but not too many that are in each and every single package. I'll be making some nice updates in future PRs that will enforce things like `"noUnusedParameters": true,` everywhere but to keep this PR reasonable i've chosen not to include them here.
1. **I see these changes do not apply to ./clients** Correct. We cannot currently provide a common tsconfig for the generated clients. Smithy generates the tsconfig for us and does not provide a way to modify that tsconfig as part of generation. We also do not want smithy to explicitly extend a higher level tsconfig as that breaks encapsulation. 

Experimental branch reference:
https://github.com/aws/aws-sdk-js-v3/pull/1279


alexforsyth:
@trivikr @AllanFly120  since this is my first major commit and it could cause some problems if I missed something, can you checkout the branch and make sure everything works and is solid on your local?

tsconfig.json
```diff
@@ -0,0 +1,33 @@
...
+    // "pretty": true,
+
+    "module": "commonjs",
+    "moduleResolution": "node",
+    "target": "es5",
+    "lib": ["es2015", "dom"],
+    "baseUrl": ".",
+    "paths": {
+      "@aws-sdk/*": ["packages/*/src"],
+      "@aws-sdk/clients-*": ["clients/clients-*/"]
```
trivikr, line 28: The package names are prefixed with `client-` and not `clients-`
Example: [@aws-sdk/client-dynamodb](https://www.npmjs.com/package/@aws-sdk/client-dynamodb)

```suggestion
      "@aws-sdk/client-*": ["clients/clients-*/"]
```

trivikr:
Verified that browsing across dependencies in `packages/*` is possible without building them

<details>
<summary>Screen recording</summary>

ezgif-6-93076c906776


</details>

tsconfig.json
alexforsyth: Ahh yes, good catch

packages/abort-controller/tsconfig.cjs.json
```diff
@@ -1,16 +1,18 @@
...
-    "module": "commonjs",
-    "target": "es5",
     "lib": ["es5", "es2015.collection"],
     "strict": true,
-    "sourceMap": true,
+    "sourceMap": false,
     "declaration": true,
     "stripInternal": true,
     "rootDir": "./src",
     "outDir": "./build",
```
trivikr, line 9: Standardize outDir across `clients/*` and `packages/*`

Recommending `"dist/cjs"` and `"dist/es"` as:
* It's currently used in clients https://github.com/aws/aws-sdk-js-v3/blob/bddeb3d172f1e26ff9e429140333d6ea624168c2/clients/client-accessanalyzer/tsconfig.json#L15
* It's easy to have transpiled code in `dist` folder as a standard. In future, if we decide to create new bundle for `es5`, it can be put in `dist/es5` folder.

```suggestion
    "outDir": "dist/cjs",
```

alexforsyth: Yeah this is a very good point. I dont really like build and build-es very much

packages/abort-controller/package.json
```diff
@@ -5,8 +5,10 @@
   "main": "./build/index.js",
   "types": "./build/index.d.ts",
   "scripts": {
-    "prepublishOnly": "tsc",
-    "pretest": "tsc -p tsconfig.test.json",
+    "prepublishOnly": "yarn build:cjs",
+    "pretest": "yarn build:cjs && yarn build:es",
```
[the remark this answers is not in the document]
alexforsyth: Yes this is also a very good point, good callout

packages/abort-controller/package.json
```diff
@@ -5,8 +5,10 @@
   "main": "./build/index.js",
   "types": "./build/index.d.ts",
   "scripts": {
-    "prepublishOnly": "tsc",
-    "pretest": "tsc -p tsconfig.test.json",
+    "prepublishOnly": "yarn build:cjs",
```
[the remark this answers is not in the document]
alexforsyth: Makes sense, will do

trivikr:
This PR needs rebase to include `karma-credential-loader` added in https://github.com/aws/aws-sdk-js-v3/pull/1296

packages/abort-controller/package.json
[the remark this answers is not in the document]
alexforsyth: Done, propagated

packages/abort-controller/package.json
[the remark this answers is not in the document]
alexforsyth: done, propagated

packages/abort-controller/tsconfig.cjs.json
alexforsyth: done, everything should be standardized to cjs. All packages should now emit a es and cjs bundle, module will map to es

.gitignore
```diff
@@ -25,7 +25,9 @@ lerna-debug.log
 package-lock.json
 
 !serviceModels/logs
+dist
 build
+build-es
```
trivikr, line 30: The directories `build` and `build-es` can be removed now since we use `dist`
```suggestion
```

alexforsyth:
> One minor comment for removing build and build-es.
> 
> Also, it looks like karma-credential-loader is not rebased from master:
> 
> * https://github.com/aws/aws-sdk-js-v3/tree/aforsyth/root-tsconfig/packages
> * https://github.com/aws/aws-sdk-js-v3/tree/master/packages
> 
> The changes would be required in karma-credential-loader too.

All done. My rebase screwed up so ended up just doing it in a merge. karma-credential-loader has changes propagated

tsconfig.json
```diff
@@ -0,0 +1,33 @@
...
+    "incremental": true,
+
+    /**
+     * Nice to have:
+     */
+    // "preserveConstEnums": true,
+    // "declaration": true,
+    // "allowUnreachableCode": false,
+    // "pretty": true,
+
```
AllanFly120, line 20: I agree to make `preserveConstEnums` true. It make generated code more readable, rather than just number/string without reference to the enum names.
All the other options are just default values, so they can be safely removed.

AllanFly120, line 20: I agree to make `preserveConstEnums` true. It make generated code more readable, rather than just number/string without reference to the enum names.
All the other options are just default values, so they can be safely removed.

tsconfig.json
```diff
@@ -0,0 +1,33 @@
...
+    "module": "commonjs",
+    "moduleResolution": "node",
+    "target": "es5",
+    "lib": ["es2015", "dom"],
+    "baseUrl": ".",
+    "paths": {
+      "@aws-sdk/*": ["packages/*/src"],
+      "@aws-sdk/client-*": ["clients/client-*/"]
```
AllanFly120, line 28: Do we need to add `protocol_tests/*` here?

tsconfig.json
```diff
@@ -0,0 +1,29 @@
...
+    "downlevelIteration": true,
+    "resolveJsonModule": true,
+    "experimentalDecorators": true,
+    "noUnusedParameters": false,
+    "removeComments": false,
+    "incremental": true,
+    /**
+     * Nice to have:
+     */
+    // "preserveConstEnums": true,
```
AllanFly120, line 15: I'm fine if you uncomment it.😊

alexforsyth: We can't uncomment without massive changes across the repo. Lets address this in a followup

PR #1185, opened 2020-05-14
chore: switch to prettier@2 defaults with 120 printWidth

trivikr:
*Issue #, if available:*
Follow-up to https://github.com/aws/aws-sdk-js-v3/pull/1183

*Description of changes:*
switch to prettier@2 defaults with 120 printWidth

**Verified that vscode-prettier formats the page on edit**
<details>
<summary>Screen recording</summary>

vscode-formatFileOnSave

</details>

<details>
<summary>prettier-vscode Output</summary>

```console
["INFO" - 7:56:48 PM] Formatting /home/trivikr/workspace/aws-sdk-js-v3/clients/client-accessanalyzer/protocols/Aws_restJson1_1.ts
["INFO" - 7:56:48 PM] Using ignore file (if present) at /home/trivikr/workspace/aws-sdk-js-v3/.prettierignore
["INFO" - 7:56:48 PM] Loaded module 'prettier@2.0.4' from '/home/trivikr/workspace/aws-sdk-js-v3/node_modules/prettier/index.js'
["INFO" - 7:56:48 PM] File Info:
{
  "ignored": false,
  "inferredParser": "typescript"
}
["INFO" - 7:56:48 PM] Detected local configuration (i.e. .prettierrc or .editorconfig), VS Code configuration will not be used
["INFO" - 7:56:48 PM] Prettier Options:
{
  "filepath": "/home/trivikr/workspace/aws-sdk-js-v3/clients/client-accessanalyzer/protocols/Aws_restJson1_1.ts",
  "parser": "typescript",
  "printWidth": 120
}
["INFO" - 7:56:49 PM] Formatting completed in 597.256223ms.

```

</details>

**Verified that yarn generate-clients use printWidth 120 on CodeGen code**

<details>
<summary>diff</summary>

Screen Shot 2020-05-14 at 1 03 41 PM

</details>

**Verified that husky formats files with printWidth 120 on precommit**

<details>
<summary>terminal output</summary>

```diff
diff --git a/clients/client-accessanalyzer/protocols/Aws_restJson1_1.ts b/clients/client-accessanalyzer/protocols/Aws_restJson1_1.ts
index 683e228413..36ef40cf8d 100644
--- a/clients/client-accessanalyzer/protocols/Aws_restJson1_1.ts
+++ b/clients/client-accessanalyzer/protocols/Aws_restJson1_1.ts
@@ -1,4 +1,5 @@
[...]

PR #1350, opened 2020-07-09
feat: add code linting and prettify

alexforsyth:
*Issue #, if available:* We don't have a linter or prettify function implemented that will easily format your code for you

*Description of changes:*
Added linting with eslint to the codebase. This runs at the root and takes ~5sec.

**yarn run lint**
real    0m4.294s
user    0m5.053s
sys     0m0.560s


.prettierrc.js
```diff
@@ -0,0 +1,7 @@
+module.exports = {
```
alexforsyth: This is added by mistake, removed here: https://github.com/aws/aws-sdk-js-v3/pull/1350/commits/716b2899c852c8f887584ef560f50d737b395022

package.json
```diff
@@ -20,7 +20,9 @@
     "test:integration-legacy": "cucumber-js --fail-fast",
     "test:integration": "jest --config jest.config.integ.js --passWithNoTests",
     "test:protocols": "yarn build:protocols && lerna run test --scope '@aws-sdk/aws-*'",
-    "local-publish": "node ./scripts/verdaccio-publish/index.js"
+    "local-publish": "node ./scripts/verdaccio-publish/index.js",
+    "lint": "eslint 'packages/**/src/*.ts' --fix",
+    "prettify": "prettier --write './packages/**/src/*.ts'"
```
trivikr, line 25:
```suggestion
```

> We don't have prettify function implemented

Although there is not separate prettify script, we prettify on save and on precommit. So, a separate prettify script is not required.
* Code to run prettier on Save https://github.com/aws/aws-sdk-js-v3/blob/e6a4fcff20560d9302032e6e9c6f3ebae91478b2/.vscode/settings.json#L3-L4
* Code to run prettier on precommit https://github.com/aws/aws-sdk-js-v3/blob/e6a4fcff20560d9302032e6e9c6f3ebae91478b2/package.json#L91-L99

.prettierrc
```diff
@@ -2,5 +2,7 @@
   "$schema": "http://json.schemastore.org/prettierrc",
   "arrowParens": "avoid",
   "endOfLine": "auto",
+  "semi": true,
+  "printWidth": 120,
```
trivikr, line 6:
```suggestion
```

This change is going to affect 10K+ files, and should be done at once. It should be done when there are very few open PRs to avoid conflicts.

There's already a existing PR which is pending review from @AllanFly120 https://github.com/aws/aws-sdk-js-v3/pull/1185
Once approved, the last two commits would be removed and recreated after rebase.

package.json
alexforsyth: Okay removed script

alexforsyth: Running prettier on precommit will be another MR

.prettierrc
alexforsyth: This affected 58 files, I reverted the commit regardless

features/extra/hooks.js
```diff
@@ -7,14 +7,7 @@ const isType = (obj, type) => {
   return Object.prototype.toString.call(obj) === "[object " + type + "]";
```
trivikr, line 7: This file needs to be removed from this PR, can be prettified separately if required

package.json
```diff
@@ -66,7 +73,7 @@
     "lerna": "3.22.1",
     "lint-staged": "^10.0.1",
     "mocha": "^8.0.1",
-    "prettier": "2.0.5",
+    "prettier": "^2.0.5",
```
trivikr, line 76: We pin the version of prettier we use, to avoid surprise formatting changes if version get upgraded by dependabot or while running yarn upgrade
```suggestion
    "prettier": "2.0.5",
```

protocol_tests/aws-ec2/EC2ProtocolClient.ts
```diff
@@ -1,65 +1,106 @@
 import {
```
trivikr, line 1: Undo lint fixes in protocol_tests as they're codegen

features/extra/hooks.js
alexforsyth: Good catch, not sure why thats there. Updated

PR #1503, opened 2020-09-08
Compatibility with Webpack 5

rvaronos:
On my Next.js project I am migrating to webpack 5. I get the following error that on my side is not possible to fix:
``Should not import the named export 'name' (imported as 'name') from default-exporting module (only default export is available soon)``

In the documentation page from webpack "To v5 from v4" the following is mentioned:
``Using named exports from JSON modules: this is not supported by the new specification and you will get a warning. Instead of import { version } from './package.json' use import package from './package.json'; const { version } = package;``

This should fix the error not just for me but also for others who are attempting:
https://github.com/vercel/next.js/issues/13341#issuecomment-665761510

### DIFF
--- a/clients/client-cognito-identity/runtimeConfig.browser.ts
+++ b/clients/client-cognito-identity/runtimeConfig.browser.ts
@@ -1,4 +1,4 @@
-import * as packageInfo from './package.json'; 
+import packageInfo from "./package.json"; 
 import { Sha256 } from "@aws-crypto/sha256-browser";
 import { FetchHttpHandler, streamCollector } from "@aws-sdk/fetch-http-handler";
 import { invalidFunction } from "@aws-sdk/invalid-dependency";
--- a/clients/client-cognito-identity/runtimeConfig.browser.ts
+++ b/clients/client-cognito-identity/runtimeConfig.browser.ts
@@ -1,4 +1,4 @@
-import packageInfo from './package.json'; 
+import * as packageInfo from './package.json'; 
 import { Sha256 } from "@aws-crypto/sha256-browser";
 import { FetchHttpHandler, streamCollector } from "@aws-sdk/fetch-http-handler";
 import { invalidFunction } from "@aws-sdk/invalid-dependency";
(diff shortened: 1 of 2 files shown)


trivikr:
Created an issue at https://github.com/aws/aws-sdk-js-v3/issues/1504, as it's a warning and not an error with bundling JS SDK v3

clients/client-cognito-identity/runtimeConfig.browser.ts
```diff
@@ -1,4 +1,4 @@
-import { name, version } from "./package.json";
+import package from './package.json';
```
trivikr, line 1: Fix for build error:
```console
runtimeConfig.browser.ts(1,8): error TS1214: Identifier expected. 'package' is a reserved word in strict mode. Modules are automatically in strict mode.
```

```suggestion
import packageInfo from './package.json'; 
```

clients/client-cognito-identity/runtimeConfig.browser.ts
```diff
@@ -11,6 +11,8 @@ import { fromUtf8, toUtf8 } from "@aws-sdk/util-utf8-browser";
 import { ClientDefaults } from "./CognitoIdentityClient";
 import { ClientSharedValues } from "./runtimeConfig.shared";
 
+const { name, version } = package;
```
trivikr, line 14:
```suggestion
const { name, version } = packageInfo;
```

trivikr (changes requested):

clients/client-cognito-identity/runtimeConfig.browser.ts
```diff
@@ -1,4 +1,4 @@
-import { name, version } from "./package.json";
+import packageInfo from './package.json';
```
trivikr, line 1: Fix for error:
```console
runtimeConfig.browser.ts(1,8): error TS1259: Module '"/codebuild/output/src713163289/src/github.com/aws/aws-sdk-js-v3/clients/client-cognito-identity/package"' can only be default-imported using the 'esModuleInterop' flag
```

```suggestion
import * as packageInfo from './package.json'; 
```

trivikr (changes requested):

trivikr:
Super-ceded by https://github.com/aws/aws-sdk-js-v3/pull/1505

Issue #1223, opened 2020-05-30
DynamoDB documentClient support


fboudra:
**Is your feature request related to a problem? Please describe.**
Migrate existing code using documentClient to v3
https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/DynamoDB/DocumentClient.html

I'm also using dynamodb-toolbox on some projects (it relies on documentClient ):
https://github.com/jeremydaly/dynamodb-toolbox

**Describe the solution you'd like**
ability to use the convenient documentClient API with v3

**Describe alternatives you've considered**
* go back to primitive calls like client putitem/getitem
* stick to v2 and don't migrate

russell-dot-js:
I'm with you @fboudra, looking forward to DocumentClient myself!
However, this is a duplicate of https://github.com/aws/aws-sdk-js-v3/issues/288

Maybe you or I could take a stab at porting over DocumentClient 🤔

mhart:
It's a duplicate, but the other issue was closed so I'm assuming this is the best place to follow progress of this?

@AllanFly120 is there somewhere else tracking this?

antstanley:
Hey folks, for those struggling with this I've created an npm package to convert JSON to DynamoDB AttributeValue type schema JSON and back again.

I'm using it today in a small production project.

I've created two packages, one a CommonJS module, and the other a ES module. You can find them, with docs here. 

It's a fork and update from the DocumentClient converter.js code.

CommonJS module - `dynamodb-type-marshall` [https://www.npmjs.com/package/dynamodb-type-marshall](https://www.npmjs.com/package/dynamodb-type-marshall)

ES module - `@esmodule/dynamodb-type-marshall` [https://www.npmjs.com/package/@esmodule/dynamodb-type-marshall](https://www.npmjs.com/package/@esmodule/dynamodb-type-marshall)

Source code here: https://github.com/senzodev/dynamodb-type-marshall

trivikr:
Thanks @antstanley for creating a package for marshalling and unmarshalling.

We've implemented marshaller (convertToAttr) in https://github.com/aws/aws-sdk-js-v3/pull/1531, and aiming to push it in `gamma.10` launch before 9/25
A PR for unmarshaller (convertToNative) will be posted soon.

trivikr:
[...]

PR #1531, opened 2020-09-18
feat(util-dynamodb): marshall to convert JavaScript object into DynamoDB Record

trivikr:
*Issue #, if available:*
Refs: https://github.com/aws/aws-sdk-js-v3/issues/1223

*Description of changes:*
Adds utility function marshall to convert JavaScript object into DynamoDB Record

```js
const { DynamoDB } = require("@aws-sdk/client-dynamodb");
const { marshall } = require("@aws-sdk/util-dynamodb");

const client = new DynamoDB(clientParams);
const params = {
  TableName: "Table",
  Item: marshall({
    HashKey: "hashKey",
    NumAttribute: 1,
    BoolAttribute: true,
    ListAttribute: [1, "two", false],
    MapAttribute: { foo: "bar" },
    NullAttribute: null,
  }),
};

await client.putItem(params);
```

### DIFF
--- a/packages/util-dynamodb/package.json
+++ b/packages/util-dynamodb/package.json
@@ -1,0 +1,28 @@
+{
+  "name": "@aws-sdk/util-dynamodb",
+  "version": "1.0.0-gamma.0",
+  "scripts": {
+    "prepublishOnly": "yarn build:cjs && yarn build:es",
+    "pretest": "yarn build:cjs",
+    "build:cjs": "tsc -p tsconfig.cjs.json",
+    "build:es": "tsc -p tsconfig.es.json",
+    "build": "yarn build:es && yarn build:cjs",
+    "test": "jest"
+  },
+  "main": "./dist/cjs/index.js",
+  "module": "./dist/es/index.js",
+  "types": "./dist/cjs/index.d.ts",
+  "author": {
+    "name": "AWS SDK for JavaScript Team",
+    "url": "https://aws.amazon.com/javascript/"
+  },
+  "license": "Apache-2.0",
+  "dependencies": {
+    "tslib": "^1.8.0"
+  },
+  "devDependencies": {
+    "@types/jest": "^26.0.4",
+    "jest": "^26.1.0",
+    "typescript": "~4.0.2"
+  }
+}
--- a/packages/util-dynamodb/package.json
+++ b/packages/util-dynamodb/package.json
@@ -18,6 +18,7 @@
   },
   "license": "Apache-2.0",
   "dependencies": {
+    "@aws-sdk/client-dynamodb": "1.0.0-gamma.9",
     "tslib": "^1.8.0"
   },
   "devDependencies": {
--- a/packages/util-dynamodb/package.json
+++ b/packages/util-dynamodb/package.json
@@ -18,10 +18,10 @@
   },
   "license": "Apache-2.0",
   "dependencies": {
-    "@aws-sdk/client-dynamodb": "1.0.0-gamma.9",
     "tslib": "^1.8.0"
   },
   "devDependencies": {
+    "@aws-sdk/client-dynamodb": "1.0.0-gamma.9",
     "@types/jest": "^26.0.4",
     "jest": "^26.1.0",
     "typescript": "~4.0.2"
--- a/packages/util-dynamodb/src/convertToAttr.spec.ts
+++ b/packages/util-dynamodb/src/convertToAttr.spec.ts
@@ -19,6 +19,14 @@ describe("convertToAttr", () => {
     });
   });
 
+  [NaN, Infinity, -Infinity].forEach((num) => {
+    it(`throws for number (special numeric value): ${num}`, () => {
+      expect(() => {
+        convertToAttr(num);
+      }).toThrowError(`Special numeric value ${num} is not allowed`);
+    });
+  });
+
   it("returns for string", () => {
     const str = "str";
     expect(convertToAttr(str)).toEqual({ S: str });
--- a/packages/util-dynamodb/src/convertToAttr.spec.ts
+++ b/packages/util-dynamodb/src/convertToAttr.spec.ts
@@ -1,6 +1,12 @@
 import { convertToAttr } from "./convertToAttr";
 
 describe("convertToAttr", () => {
+  [true, false].forEach((bool) => {
+    it(`test boolean "${bool}"`, () => {
+      expect(convertToAttr(bool)).toEqual({ BOOL: bool });
+    });
+  });
+
   it("returns for number", () => {
     const num = 5;
     expect(convertToAttr(num)).toEqual({ N: num.toString() });
--- a/packages/util-dynamodb/src/convertToAttr.spec.ts
+++ b/packages/util-dynamodb/src/convertToAttr.spec.ts
@@ -1,0 +1,13 @@
+import { convertToAttr } from "./convertToAttr";
+
+describe("convertToAttr", () => {
+  it("returns for number", () => {
+    const num = 5;
+    expect(convertToAttr(num)).toEqual({ N: num.toString() });
+  });
+
+  it("returns for string", () => {
+    const str = "str";
+    expect(convertToAttr(str)).toEqual({ S: str });
+  });
+});
--- a/packages/util-dynamodb/src/convertToAttr.spec.ts
+++ b/packages/util-dynamodb/src/convertToAttr.spec.ts
@@ -144,7 +144,8 @@ describe("convertToAttr", () => {
 
     it("bigint set", () => {
       // @ts-expect-error BigInt literals are not available when targeting lower than ES2020.
-      const set = new Set([1n, 2n, 3n]);
+      const bigNum = BigInt(Number.MAX_SAFE_INTEGER) + 2n;
+      const set = new Set([bigNum, -bigNum]);
       expect(convertToAttr(set)).toEqual({ NS: Array.from(set).map((num) => num.toString()) });
     });
 
--- a/packages/util-dynamodb/src/convertToAttr.spec.ts
+++ b/packages/util-dynamodb/src/convertToAttr.spec.ts
@@ -192,7 +192,15 @@ describe("convertToAttr", () => {
   });
 
   describe(`unsupported type`, () => {
-    [undefined].forEach((data) => {
+    class FooObj {
+      constructor() {
+        // @ts-ignore
+        this.foo = "foo";
+      }
+    }
+
+    // ToDo: Serialize ES6 class objects as string https://github.com/aws/aws-sdk-js-v3/issues/1535
+    [undefined, new Date(), new FooObj()].forEach((data) => {
       it(`throws for: ${String(data)}`, () => {
         expect(() => {
           // @ts-ignore Argument is not assignable to parameter of type 'NativeAttributeValue'
--- a/packages/util-dynamodb/src/convertToAttr.spec.ts
+++ b/packages/util-dynamodb/src/convertToAttr.spec.ts
@@ -189,4 +189,15 @@ describe("convertToAttr", () => {
       expect(convertToAttr(str)).toEqual({ S: str });
     });
   });
+
+  describe(`unsupported type`, () => {
+    [undefined].forEach((data) => {
+      it(`throws for: ${String(data)}`, () => {
+        expect(() => {
+          // @ts-ignore Argument is not assignable to parameter of type 'NativeAttributeValue'
+          convertToAttr(data);
+        }).toThrowError(`Unsupported type passed: ${String(data)}`);
+      });
+    });
+  });
 });
--- a/packages/util-dynamodb/src/convertToAttr.spec.ts
+++ b/packages/util-dynamodb/src/convertToAttr.spec.ts
@@ -120,6 +120,16 @@ describe("convertToAttr", () => {
         [uint8Arr, biguintArr],
         [{ B: uint8Arr }, { B: biguintArr }],
       ],
+      [
+        [
+          { nullKey: null, boolKey: false },
+          { stringKey: "one", numberKey: 1.01, bigintKey: BigInt(9007199254740996) },
+        ],
+        [
+          { M: { nullKey: { NULL: true }, boolKey: { BOOL: false } } },
+          { M: { stringKey: { S: "one" }, numberKey: { N: "1.01" }, bigintKey: { N: "9007199254740996" } } },
+        ],
+      ],
     ].forEach(([input, output]) => {
       it(`testing list: ${input}`, () => {
         // @ts-ignore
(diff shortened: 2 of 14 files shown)


russell-dot-js:
It would be really nice to expose some way to expose custom marshalling/unmarshalling config. Even if it's just a function for custom marshalling. One of the big drawbacks of the current DocumentClient implementations is that JS Dates serialize to maps, and the data gets completely hosed (saves as {} IIRC). What if we had something like this?

```
premarshall: ({key: string, value: any}) => any;

new DocumentClient({
  premarshall: ({key, value}) => {
    if(v instanceof Date) {
      return v.toIsoString(); // or convert to number depending on use case
    }
    if(v instanceof moment) {
      return v.format();
    }

    if(key === 'someKeyIDontWantInDynamo') {
      return undefined;
    }
    return v;
  }
})
```

Then the library would call premarshall prior to converting to AttributeValues

IMO it's more important to have this on the way in to dynamo on the way out, because there's always some casting from DB values to runtime values anyway (e.g. converting iso string back to date, stripping database concerns, etc). But some way to inject some casting logic that could be applied to Item, Attributes, Items, etc based on endpoint is a secondary item on my wishlist.

What are your thoughts?

trivikr:
@russell-dot-js Custom premarshaller is a good idea. Can you create a feature-request for it?
Link: https://github.com/aws/aws-sdk-js-v3/issues/new?assignees=&labels=feature-request&template=---feature-request.md&title=

Since v3 is is prerelease, we can still modify the marshaller (convertToAttr).
Is Date the only type you have concern about? Or are there generic JavaScript types which should be included in default implementation?

russell-dot-js:
> @russell-dot-js Custom premarshaller is a good idea. Can you create a feature-request for it?
> Link: https://github.com/aws/aws-sdk-js-v3/issues/new?assignees=&labels=feature-request&template=---feature-request.md&title=
> 
> Since v3 is is prerelease, we can still modify the marshaller (convertToAttr).
> Is Date the only type you have concern about? Or are there generic JavaScript types which should be included in default implementation?

@trivikr I'll take the discussion to the issue, and I'm happy to implement the feature, but should there be a default implementation? I wouldn't want to break consumers who pass dates blindly and expect a number or a string... then again, right now, dates don't work at all. Maybe even logging a warning when an unmarshalled date makes it to convertToAttr would be good enough...

My concern about default implementation is that everyone has a different way of using dynamo. Some people may cram 4 attributes into a sort key, then convert back to whatever those fields were to begin with on the way out. So it might be tough for a default implementation to make an assumption on how other native JS types should be handled - but we can at least identify when we receive those and don't know how to marshall them.

packages/util-dynamodb/src/convertToAttr.spec.ts
```diff
@@ -0,0 +1,220 @@
...
+  describe("string", () => {
+    ["", "string", "'single-quote'", '"double-quote"'].forEach((str) => {
+      it(`returns for string: ${str}`, () => {
+        expect(convertToAttr(str)).toEqual({ S: str });
+      });
+    });
+  });
+
+  describe(`unsupported type`, () => {
+    [undefined].forEach((data) => {
```
russell-dot-js, line 195: curious -- why throw errors for undefined? 

I would hope we can borrow the rules from JSON serialization... meaning, undefined never makes it to dynamo

packages/util-dynamodb/src/convertToAttr.ts
```diff
@@ -0,0 +1,117 @@
...
+ *
+ * @param {NativeAttributeValue} data The data to convert to a DynamoDB AttributeValue
+ * @param {Object} options
+ * @param {boolean} options.convertEmptyValues Whether to automatically
+ *                      convert empty strings, blobs, and sets to `null`
+ */
+export const convertToAttr = (
+  data: NativeAttributeValue,
+  options?: {
+    convertEmptyValues: boolean;
```
russell-dot-js, line 16: just a note that this config is a bit confusing in the current document client, and might be better handled with custom premarshall https://github.com/aws/aws-sdk-js-v3/issues/1533

It's confusing because `""` is considered an empty value, but `undefined` is not

packages/util-dynamodb/src/convertToAttr.ts
```diff
@@ -0,0 +1,117 @@
...
+  }
+): AttributeValue => {
+  if (Array.isArray(data)) {
+    return { L: data.map((item) => convertToAttr(item)) };
+  } else if (data?.constructor?.name === "Set") {
+    return convertToSetAttr(data as Set<any>, options?.convertEmptyValues);
+  } else if (data?.constructor?.name === "Object") {
```
russell-dot-js, line 23: is this code borrowed from the current (v2) DocumentClient? or is this a new implementation?

if I'm reading correctly, this will only serialize pure JS Objects to JSON.
Anyone who's using ES6 classes or other easily serializable objects would be hosed without custom premarshall?

Again, borrowing the rules from JSON serialization would be $

russell-dot-js (commented):
overall LGTM, but may be regressions from current DocumentClient Convereter's handling of objects

other misc. comments added

packages/util-dynamodb/src/convertToAttr.ts
trivikr: The parameter `convertEmptyValues` is added for backward compatibility with v2 DocumentClient
API Reference: https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/DynamoDB/DocumentClient.html

packages/util-dynamodb/src/convertToAttr.ts
trivikr: > is this code borrowed from the current (v2) DocumentClient? or is this a new implementation?

This PR is new implementation in TypeScript based on the behavior of v2 DocumentClient.

> Anyone who's using ES6 classes or other easily serializable objects would be hosed without custom premarshall?

Can you provide an example?

packages/util-dynamodb/src/convertToAttr.spec.ts
trivikr: convertToAttr returns `AttributeValue` to be populated in DynamoDB. If it returns `undefined`, there would be a ValidationException as can be seen in the code below:

<details>
<summary>Code</summary>

```js
const { DynamoDB } = require("@aws-sdk/client-dynamodb");

(async () => {
  const region = "us-west-2";

  const TableName = `test-doc-client`;
  const id = { S: "id1" };
  const undefinedString = { S: undefined };

  const client = new DynamoDB({ region });
  await client.putItem({ TableName, Item: { id, undefinedString } });
  console.log(await client.getItem({ TableName, Key: { id } }));
})();

```

</details>

<details>
<summary>Output</summary>

```console
(node:4711) UnhandledPromiseRejectionWarning: ValidationException: Supplied AttributeValue is empty, must contain exactly one of the supported datatypes
    at deserializeAws_json1_0PutItemCommandError (/Users/trivikr/workspace/ddb-doc-client-test/node_modules/@aws-sdk/client-dynamodb/dist/cjs/protocols/Aws_json1_0.js:1420:41)
    at processTicksAndRejections (internal/process/task_queues.js:93:5)
(Use `node --trace-warnings ...` to show where the warning was created)
(node:4711) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 6)
(node:4711) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

</details>

convertToAttr ensures that the error is thrown during conversion itself.

trivikr:
> should there be a default implementation for Date?

The recommended way in DynamoDB is to store dates is ISO 8601 strings.
Documentation: https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/HowItWorks.NamingRulesDataTypes.html#HowItWorks.DataTypes.String

I'll store date by converting it to ISO 8601 string using `toISOString()` in convertToAttr function
MDN: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toISOString

packages/util-dynamodb/src/convertToAttr.ts
russell-dot-js: For sure
Here is the source, btw:
https://github.com/aws/aws-sdk-js/blob/5155fde557ecc1d166bb5040823289254bea1641/lib/dynamodb/converter.js#L27

The old DocumentClient defaulted to treating objects as map. Thus

```
class FooObj {
  constructor() {
    this.foo = 'foo';
  }
}

JSON.stringify(new FooObj())
"{"foo":"foo"}"

// new implementation needs constructor name to be Object or it errors
(new FooObj()).constructor.name
"FooObj"

// old implementation checked typeof
typeof new FooObj()
"object"
```

See https://github.com/aws/aws-sdk-js/blob/cc29728c1c4178969ebabe3bbe6b6f3159436394/lib/dynamodb/types.js#L3

russell-dot-js:
> > should there be a default implementation for Date?
> 
> The recommended way in DynamoDB is to store dates is ISO 8601 strings.
> Documentation: https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/HowItWorks.NamingRulesDataTypes.html#HowItWorks.DataTypes.String
> 
> I'll store date by converting it to ISO 8601 string using `toISOString()` in convertToAttr function
> MDN: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toISOString

Reading those docs, it doesn't say the recommended way is ISO 8601. It says you "can use ISO 8601". Immediately before that, in the numbers docs, it says you can use numbers also to represent date. It does not make a recommendation.

We use ISO 8601, but I don't think the default sdk should make assumptions on what you want to do. In other languages (e.g. Go, .NET, Java), the dynamo client uses JSON serialization by default.

trivikr:
I created a feature request for supporting date https://github.com/aws/aws-sdk-js-v3/issues/1534

This PR will continue throwing error for date - as it fixes the bug with v2 DynamoDB client which converts Date to an empty map as follows:

<details>
<summary>Code</summary>

```js
const AWS = require("aws-sdk");

(async () => {
  const region = "us-west-2";

  const TableName = `test-doc-client`;
  const id = "id1";
  const date = new Date();

  const client = new AWS.DynamoDB.DocumentClient({ region });
  await client.put({ TableName, Item: { id, date } }).promise();
  console.log(await client.get({ TableName, Key: { id } }).promise());
})();

```

</details>

<details>
<summary>Output</summary>

```console
{ Item: { id: 'id1', date: {} } }
```

</details>

packages/util-dynamodb/src/convertToAttr.ts
trivikr: This is a miss.

However, using `JSON.stringify()` for serializing and `JSON.parse()` for deserializing is a norm in JavaScript world.
* Example StackOverflow discussion: https://stackoverflow.com/a/40201783
* MDN: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify

I created an issue to discuss this internally, and will post updates in https://github.com/aws/aws-sdk-js-v3/issues/1535

packages/util-dynamodb/src/convertToAttr.ts
```diff
@@ -0,0 +1,117 @@
...
+      M: Object.entries(data).reduce(
+        (acc: { [key: string]: AttributeValue }, [key, value]: [string, NativeAttributeValue]) => ({
+          ...acc,
+          [key]: convertToAttr(value),
+        }),
+        {}
+      ),
+    };
+  } else {
+    if (data === null && typeof data === "object") {
```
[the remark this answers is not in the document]
trivikr: The else was added to separate collection types (List, Set, Map) with scalar types (boolean, number, string, etc)

I can add a comment about it, if it appears to be a better organization.
If that happens, the throw Error on line 55 needs to be moved out of else.

What are your thoughts? Should I create a new function `convertScalar` instead?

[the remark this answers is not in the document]
AllanFly120: You can wrap all these into `convertScalarAttrs()`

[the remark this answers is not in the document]
trivikr: Updated in [`e3ba847` (#1531)](https://github.com/aws/aws-sdk-js-v3/pull/1531/commits/e3ba84779287a351c5b8bc67dbca68093503d2ba)

trivikr:
The build failure is related to intermittent test failure of NodeHttpHandler
Created issue at https://github.com/aws/aws-sdk-js-v3/issues/1539

packages/util-dynamodb/package.json
```diff
@@ -0,0 +1,29 @@
...
+  "main": "./dist/cjs/index.js",
+  "module": "./dist/es/index.js",
+  "types": "./dist/cjs/index.d.ts",
+  "author": {
+    "name": "AWS SDK for JavaScript Team",
+    "url": "https://aws.amazon.com/javascript/"
+  },
+  "license": "Apache-2.0",
+  "dependencies": {
+    "@aws-sdk/client-dynamodb": "1.0.0-gamma.9",
```
AllanFly120, line 21: If only TS interface is imported from `@aws-sdk/client-dynamodb`, I think we can move it to dev dependency

AllanFly120 (commented):
It's not a blocker, but having cross-platform test cases here is necessary. It will validate this works for `Buffer` as well as `Blob` and `File`

packages/util-dynamodb/package.json
trivikr: Fixed in [`2e12934` (#1531)](https://github.com/aws/aws-sdk-js-v3/pull/1531/commits/2e12934268857e9b19b81c20eb41f25962c2221d)

trivikr:
> It's not a blocker, but having cross-platform test cases here is necessary. It will validate this works for `Buffer` as well as `Blob` and `File`

Created an issue for follow-up https://github.com/aws/aws-sdk-js-v3/issues/1542

packages/util-dynamodb/src/convertToAttr.ts
trivikr: We're not using `data instanceof Object`, as it applies to binary data as shown in code below:

```js
const data = new ArrayBuffer();
data instanceof Object;
```

AllanFly120: Would checking in order of `set`, `list`, `map`, `scalars` help?

trivikr: > Would checking in order of `set`, `list`, `map`, `scalars` help?

Did you mean moving check for `data instanceof Object` in `scalars` after String?

trivikr: > Would checking in order of `set`, `list`, `map`, `scalars` help?

The check for binary happens under `scalars`, so map check should be moved after it.
Currently, check for scalars is in an else block, so check for `map` will have to be moved inside scalars.

trivikr: Can you share an example where `data?.constructor?.name === "Object"` will fail for JS for maps because of prototype chain?

AllanFly120 (approved):
Awesome! Thanks for putting this together!

Issue #1543, opened 2020-09-29
bug(util-dynamodb): BigInt value increases by 1n after marshall+unmarshall


trivikr:
**Describe the bug**
BigInt value increases by 1n after marshall+unmarshall

**SDK version number**
`@aws-sdk/util-dynamodb@1.0.0-gamma.1`

**Is the issue in the browser/Node.js/ReactNative?**
Node.js

**Details of the browser/Node.js/ReactNative version**
`v12.18.3`

**To Reproduce (observed behavior)**

<details>
<summary>Code</summary>

```js
const { deepStrictEqual } = require("assert");
const { marshall, unmarshall } = require("@aws-sdk/util-dynamodb");

const bigInt = 9007199254740995n;
deepStrictEqual(bigInt, unmarshall(marshall({ bigInt })));
```

</details>

<details>
<summary>Output</summary>

```console
assert.js:100
  throw new AssertionError(obj);
  ^

AssertionError [ERR_ASSERTION]: Expected values to be strictly deep-equal:
+ actual - expected

+ 9007199254740995n
- {
-   bigInt: 9007199254740996n
- }
    at Object.<anonymous> (/local/home/trivikr/workspace/ddb-test/bigint.js:5:1)
    at Module._compile (internal/modules/cjs/loader.js:1137:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1157:10)
    at Module.load (internal/modules/cjs/loader.js:985:32)
    at Function.Module._load (internal/modules/cjs/loader.js:878:14)
    at Function.executeUserEntryPoint [as runMain] (internal/modules/run_main.js:71:12)
    at internal/main/run_main_module.js:17:47 {
  generatedMessage: true,
  code: 'ERR_ASSERTION',
  actual: 9007199254740995n,
  expected: { bigInt: 9007199254740996n },
  operator: 'deepStrictEqual'
}
```

</details>

**Expected behavior**
Value of BigInt doesn't change while marshalling/unmarshalling. i.e. No AssertionError

trivikr:
This happens because Number value for a BigInt string if one more than what's passed to it.

```node
> Number("9007199254740995")
9007199254740996
> BigInt(Number("9007199254740995"))
9007199254740996n
```

This conversion is used in code at https://github.com/aws/aws-sdk-js-v3/blob/1b1e40d82da60a98228467662840e4bf9086683e/packages/util-dynamodb/src/convertToNative.ts#L49-L53

[...]

PR #1544, opened 2020-09-29
fix(util-dynamodb): create BigInt from string value instead of number

trivikr:
*Issue #, if available:*
Fixes: https://github.com/aws/aws-sdk-js-v3/issues/1543

*Description of changes:*
creates BigInt from string value instead of number


alexforsyth (approved):
Looks good to me!

PR #1579, opened 2020-10-16
chore: update smithy models as of 10/16

trivikr:
*Issue #, if available:*
Internal issue JS-2156

*Description of changes:*
update smithy models as of 10/16


alexforsyth (approved):
model updates LGTM

Issue #1695, opened 2020-11-18
Make handlerOptions optional in RequestHandler


trivikr:
**Is your feature request related to a problem? Please describe.**
Not a problem, but an empty hash needs to be passed for httpOptions when calling handle function.

Code links:
* RequestHandler: https://github.com/aws/aws-sdk-js-v3/blob/9e58bbb94ad89ade79def2cb808510b8d02e23cb/packages/types/src/transfer.ts#L11
* The keys in HttpHandlerOptions are optional https://github.com/aws/aws-sdk-js-v3/blob/96f61bbead0773384d0f63966eda3c5f80ba4c41/packages/types/src/http.ts#L96-L101
* node-http-handler https://github.com/aws/aws-sdk-js-v3/blob/9e58bbb94ad89ade79def2cb808510b8d02e23cb/packages/node-http-handler/src/node-http-handler.ts#L53
* node-http2-handler https://github.com/aws/aws-sdk-js-v3/blob/9e58bbb94ad89ade79def2cb808510b8d02e23cb/packages/node-http-handler/src/node-http2-handler.ts#L43
* fetch-http-handler https://github.com/aws/aws-sdk-js-v3/blob/9e58bbb94ad89ade79def2cb808510b8d02e23cb/packages/fetch-http-handler/src/fetch-http-handler.ts#L28

**Describe the solution you'd like**
Make handlerOptions optional in RequestHandler

**Describe alternatives you've considered**
Stick with required handlerOptions, as it's used internally by client code and not directly called by customers.

**Additional context**
This was noticed while writing new unit tests for node-http-handler in https://github.com/aws/aws-sdk-js-v3/pull/1693

PR #1697, opened 2020-11-18
chore(handler): update Options name to match that of class

trivikr:
*Issue #, if available:*
Noticed while working on https://github.com/aws/aws-sdk-js-v3/issues/1695

*Description of changes:*
* Updates Options name to match that of class.
* Creates new private variables in class to store options values instead of entire object.

### DIFF
--- a/packages/fetch-http-handler/src/fetch-http-handler.ts
+++ b/packages/fetch-http-handler/src/fetch-http-handler.ts
@@ -18,7 +18,11 @@ export interface FetchHttpHandlerOptions {
 }
 
 export class FetchHttpHandler implements HttpHandler {
-  constructor(private readonly httpOptions: FetchHttpHandlerOptions = {}) {}
+  private readonly requestTimeout?: number;
+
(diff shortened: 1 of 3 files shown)


packages/fetch-http-handler/src/fetch-http-handler.ts
```diff
@@ -9,7 +9,7 @@ declare let AbortController: any;
 /**
  * Represents the http options that can be passed to a browser http client.
  */
-export interface BrowserHttpOptions {
+export interface FetchHttpHandlerOptions {
```
alexforsyth, line 12: Are we renaming this just to match the file?

alexforsyth (approved):
LGTM

packages/fetch-http-handler/src/fetch-http-handler.ts
trivikr: > Are we renaming this just to match the file?

To match the class name where options interface is passed to.

PR #1707, opened 2020-11-20
refactor(s3-presigned-post): use virtual hosted style URLs

kripod:
*Issue #, if available:* N/A

*Description of changes:* According [to the AWS docs](https://docs.aws.amazon.com/AmazonS3/latest/dev/UsingBucket.html#virtual-host-style-url-ex), path-style URLs will eventually be deprecated. Thus, virtual hosted style access should be preferred.

### DIFF
--- a/packages/s3-presigned-post/src/createPresignedPost.spec.ts
+++ b/packages/s3-presigned-post/src/createPresignedPost.spec.ts
@@ -78,7 +78,7 @@ describe("createPresignedPost", () => {
       Bucket,
       Key,
     });
-    expect(url).toBe("https://s3.us-foo-1.amazonaws.com/bucket"),
+    expect(url).toBe("https://bucket.s3.us-foo-1.amazonaws.com"),
       expect(fields).toEqual({
         bucket: Bucket,
         key: Key,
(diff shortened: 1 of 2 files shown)


okofish:
I ran into this issue too - surprising to see this part of the new SDK defaulting to path-style URLs...

Based on working around this in my own code, I suggest that it might be better to go through the [`middleware-bucket-endpoint`](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/modules/_aws_sdk_middleware_bucket_endpoint.html) package rather than reimplementing the logic from scratch. That way it's a bit easier to support things like Transfer Acceleration and dual-stack endpoints. I think the function to use is [`bucketHostname`](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/modules/_aws_sdk_middleware_bucket_endpoint.html#buckethostname-3), similar to what's done [here](https://github.com/aws/aws-sdk-js-v3/blob/2165d4a5c2aec2945900bbe8e79d497da8a80cfa/packages/middleware-bucket-endpoint/src/bucketEndpointMiddleware.ts#L35).

trivikr:
ToDo: Since this PR is old, the action item is to create a feature request based on the fix suggests and close this PR.

RanVaknin:
Since this PR is so old, I created a [feature request](https://github.com/aws/aws-sdk-js-v3/issues/5257) to track this in our backlog. If someone wants to take another stab at implementing this I'll have the team prioritize this.

Thanks again, and sorry for the inconvenience. 
Ran

Issue #1742, opened 2020-12-04
SQS sendMessage sends messages twice


rahulswarnkar:
**Describe the bug**
I am using `SQSClient.send(new SendMessageCommand(...))` to send messages to a queue. For each request, the queue ends up with two identical messages with same MD5 (different IDs). The response from the above request shows one of the two IDs. Same behaviour is observed with `SQS.sendMessage(...)` 

**SDK version number**
`1.0.0-rc.7`

**Is the issue in the browser/Node.js/ReactNative?**
Node.js

**Details of the browser/Node.js/ReactNative version**
v12.0.0

**To Reproduce (observed behavior)**
```js
import * as sqs from "@aws-sdk/client-sqs";

const QUEUE_NAME = "my-queue";
const MESSAGE = { data: "my-data" };

(async () => {
  const client = new sqs.SQSClient({ region: "us-east-1" });
  const queueUrl = await client.send(new sqs.GetQueueUrlCommand({ QueueName: QUEUE_NAME }));
  const response = await client.send(
    new sqs.SendMessageCommand({ QueueUrl: queueUrl.QueueUrl, MessageBody: JSON.stringify(MESSAGE) }),
  );
  console.log(response);
})();
```

**Expected behavior**
One message in queue for one message request

**Screenshots**
Response from sendMessage request:
```console
{
  '$metadata': {
    httpStatusCode: 200,
    httpHeaders: {
      'x-amzn-requestid': 'fd3f98cd-883a-5c7d-a000-c0898ce776cd',
      date: 'Fri, 04 Dec 2020 09:34:21 GMT',
      'content-type': 'text/xml',
      'content-length': '378'
    },
    requestId: 'fd3f98cd-883a-5c7d-a000-c0898ce776cd',
    attempts: 1,
    totalRetryDelay: 0
  },
  MD5OfMessageBody: 'a54155364f584b052f1b6733f0ea1343',
  MessageId: 'db7844dc-33a4-405a-9aa8-0f38580e3229',
  MD5OfMessageAttributes: undefined,
  MD5OfMessageSystemAttributes: undefined,
  SequenceNumber: undefined
}
```
Messages as seen on AWS console:
![db7844dc-33a4-405a-9aa8-0f38580e3229](https://user-images.githubusercontent.com/2743012/101150146-be501880-3617-11eb-8875-943c1c05c205.png)
[...]

PR #1752, opened 2020-12-08
fix(middleware-sdk-sqs): call next() exactly once in sendMessageMiddleware

trivikr:
*Issue #, if available:*
Fixes: https://github.com/aws/aws-sdk-js-v3/issues/1742

*Description of changes:*
call next() exactly once in sendMessageMiddleware

Verified that the issue is not reproducible with code from https://github.com/aws/aws-sdk-js-v3/issues/1742#issuecomment-740401577


AllanZhengYP (approved):
🚢, We should've catch it before😔

PR #1761, opened 2020-12-10
fix(codegen): import SENSITIVE_STRING only when used

trivikr:
*Issue #, if available:*
Refs: https://github.com/awslabs/smithy-typescript/pull/244

*Description of changes:*
import SENSITIVE_STRING only when used


alexforsyth (approved):
LGTM

PR #1838, opened 2020-12-23
chore(clients): retain structure members' iteration order

trivikr:
*Issue #, if available:*
Depends on https://github.com/awslabs/smithy-typescript/pull/254

*Description of changes:*
retain structure members' iteration order.

The codegen was run five times, and the Structure members' order was not updated.
```
$ yarn generate-clients
```


alexforsyth (approved):
Interesting that these two are the only files that are impacted.

srchase (approved):

PR #1829, opened 2020-12-23
fix(lib-storage): cleanup stream listeners to prevent memory leak

samsullivan:
**Issue**
Fixes #1728 (#1828)

*Description of changes:*
In `lib-storage`'s helper for chunking a `stream.Readable`, remove event listeners at the end of `_chunkFromStream()`. This prevents memory leaks caused by streaming the data to a growing number of variables as the function runs on longer files.

To verify, I used the example file upload script and uploaded a ~75MB file to S3 while debugging free memory. Before this patch, Node crashed with ~500MB left; afterwards, it finished and never dipped below 2GB free.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #1841, opened 2020-12-24
fix(lib-storage): improve performance with stream's readable event

samsullivan:
I didn't make an issue for this, since it's hard to make a minimal example recreating the problem; however, I noticed this when attempting to test #1829 in a production use case.

*Description of changes:* For performance reasons, switch from `stream.on("data")` to `stream.on("readable")`. Before making this change, I wasn't able to stream from a large/undetermined-length source. Individual uploads would get stalled which slowed everything else down until it eventually was so slow that the client I was streaming from kills the stream's connection. All of this is fine when switching event listeners.

From [Node's `Stream` docs](https://nodejs.org/api/stream.html#stream_event_readable), it says:
> In general, the `readable.pipe()` and `'data'` event mechanisms are easier to understand than the `'readable'` event. However, handling `'readable'` might result in increased throughput.

This is also what the v2 library uses to stream an upload to S3: https://github.com/aws/aws-sdk-js/blob/v2.817.0/lib/s3/managed_upload.js#L188

### DIFF
--- a/lib/storage/src/data-chunk/readable-helper.ts
+++ b/lib/storage/src/data-chunk/readable-helper.ts
@@ -37,12 +37,17 @@ function _chunkFromStream(stream: Readable, chunkSize: number, oldBuffer: Buffer
   let currentChunk = oldBuffer;
   return new Promise((resolve, reject) => {
     const cleanupListeners = () => {
-      stream.removeAllListeners("data");
+      stream.removeAllListeners("readable");
       stream.removeAllListeners("error");
       stream.removeAllListeners("end");
     };
 
-    stream.on("data", (chunk) => {
+    stream.on("readable", () => {
+      const chunk = stream.read();
+      if (!chunk) {
+        return;
+      }
+
       currentChunk = Buffer.concat([currentChunk, Buffer.from(chunk)]);
       if (currentChunk.length >= chunkSize) {
         cleanupListeners();


thomassuckow:
I don't know if my "benchmarking" is better, I updated to v3 and deployed to our testing environment and watched the world end ^_^.

Now I'm using a hybrid setup with v2 for upload and v3 for download (I really wanted the convenience of knowing if the file exists (without a race condition) before piping the stream).

Do you know the easiest way to link to a build of this branch? Presumably if i check it out and run `yarn build` the libraries end up somewhere and I should just need to `yarn someMagicThing` in my project to reference it. If you don't I'll try to figure it out after I get done with some other fires.

samsullivan:
I was building things locally to test and it was a little time intensive; however, I just tried this and it looks like it did what's needed (and should work w/`npm i` as well):

```
yarn add @aws-sdk/lib-storage@git+ssh://git@github.com/samsullivan/aws-sdk-js-v3#fix-lib-upload-readable-stream-use-readable-event
```

alexforsyth:
Hey thanks so much for opening this. This is awesome, let me take a quick look and we'll get this merged

samsullivan:
Appreciate the review (again)! I'm excited to be able to upgrade from v2 in a production use case after this next release :+1:

monken:
There are some additional performance issues. https://github.com/aws/aws-sdk-js-v3/blob/e8322f8035f54307b109668883c4b4e43d34bad7/lib/storage/src/data-chunk/readable-helper.ts#L51 makes a copy of `currentChunk` on every iteration. This is a very slow operation in JavaScript land. Instead, the chunks could be pushed into a `chunks` array and then `concat`ed once the desired length has been reached. 

I'm also wondering if this implementation will handle back-pressure where the network bandwidth for the upload is slower than the read stream (e.g. try writing from `createReadStream('/dev/urandom')` to S3 and you should see the memory usage increase).

The NodeJS documentations says 

> Developers that require more fine-grained control over the transfer and generation of data can use the EventEmitter and readable.on('readable')/readable.read() or the readable.pause()/readable.resume() APIs.

It looks like we'd be mixing `pause()`, `resume()` and `on('readable')` in this PR.

samsullivan:
Thanks @monken; reading some now and I'll try to take another hack at it.

monken:
@samsullivan I created #1955 that implements the "less concat" solution. Could you please give it a go and see if it also fixes your performance issues?

samsullivan:
Closed in favor of #1955.

PR #1886, opened 2021-01-08
fix(lib-storage): chunk from readable only when defined

alexforsyth:
*Issue #, if available:*
Unit tests for `lib/storage` are broken on node 12+, they work in node 10. The root cause is a minor difference in how readable works in node 12+. 

*Description of changes:*
The functionality of readable differs slightly between node 10 and node 12+. This results in ` stream.on("end", () => {` listener never being called in node 12+ but always being called in node 10. Due to this, chunking never actually ends in node 12+. The unit tests catch this and complain when being run in node 12+.

I've fixed the code to rely on `streamreader.readable` for end-state. This should work in both node 10 and node 12+ now.


lib/storage/src/data-chunk/readable-helper.ts
```diff
@@ -4,27 +4,25 @@ import { Readable } from "stream";
...
 export async function* chunkFromReadable(reader: Readable, chunkSize: number): AsyncGenerator<DataPart, void, unknown> {
   let partNumber = DEFAULT.MIN_PART_NUMBER;
   let oldBuffer = Buffer.from("");
   while (partNumber < DEFAULT.MAX_PART_NUMBER) {
-    reader.resume();
-    const result = await _chunkFromStream(reader, chunkSize, oldBuffer);
-    reader.pause();
-
+    let currentBuffer = oldBuffer;
+    if(reader.readable) {
```
trivikr, line 12: Looks like prettier was not run on this code.
Either VSCode was not used, or prettier extension was not installed/disabled. And commit was created with `--no-verify` which didn't run pre-commit hooks to verify.

Do run prettier on this file for next revision, and create an issue if prettier failed to run because of some other reason.
 
```suggestion
    if (reader.readable) {
```

alexforsyth: yeah looks like Prettier isnt running for me. Strange. Created an issue https://github.com/aws/aws-sdk-js-v3/issues/1888

trivikr: reopening as it's not resolved.

trivikr (approved):

AllanZhengYP (approved):

Issue #1909, opened 2021-01-13
Waiter waits before polling for first request


trivikr:
**Describe the bug**
Waiter waits before polling for first request, which it shouldn't.

**SDK version number**
* v3: `"@aws-sdk/client-s3": "3.2.0",`
* v2 for reference: `"aws-sdk": "2.827.0",`

**Is the issue in the browser/Node.js/ReactNative?**
All

**Details of the browser/Node.js/ReactNative version**
N/A

**To Reproduce (observed behavior)**

In v2, the waiter doesn't wait before making the first poll request.

<details>
<summary>v2 Code</summary>

```js
const AWS = require("aws-sdk");
require("console-stamp")(console, "[ss.l]");

(async () => {
  const region = "us-west-2";
  const client = new AWS.S3({ region, logger: console });

  const Bucket = "test-waiters-call-time";

  console.log("BEFORE createBucket");
  await client.createBucket({ Bucket }).promise();

  console.log("BEFORE waitForBucketExists");
  await client.waitFor("bucketExists", { Bucket }).promise();

  console.log("BEFORE deleteBucket");
  await client.deleteBucket({ Bucket }).promise();
})();
```

</details>

<details>
<summary>v2 Output</summary>

```console
[[00.395]] [LOG]    BEFORE createBucket
[[01.106]] [LOG]    [AWS s3 200 0.706s 0 retries] createBucket({
  Bucket: 'test-waiters-call-time',
  CreateBucketConfiguration: { LocationConstraint: 'us-west-2' }
})
[[01.108]] [LOG]    BEFORE waitForBucketExists
[[01.186]] [LOG]    [AWS s3 200 0.077s 0 retries] headBucket({ Bucket: 'test-waiters-call-time' })
[[01.187]] [LOG]    BEFORE deleteBucket
[[01.544]] [LOG]    [AWS s3 204 0.356s 0 retries] deleteBucket({ Bucket: 'test-waiters-call-time' })
```

</details>

Note that:
* waitForBucketExists is called at `01.108`
* headBucket is called at `01.186`

In v3, the waiter waits before making the first poll request.

<details>
<summary>v3 Code</summary>

```js
const { S3, waitForBucketExists } = require("@aws-sdk/client-s3");
require("console-stamp")(console, "[ss.l]");

(async () => {
  const region = "us-west-2";
  const client = new S3({ region, logger: console });

  const Bucket = "test-waiters-call-time";

  console.log("BEFORE createBucket");
[...]

PR #1915, opened 2021-01-14
fix(util-waiters): waiters should call operation once before entering waiting

alexforsyth:
*Issue #, if available:*
https://github.com/aws/aws-sdk-js-v3/issues/1909

*Description of changes:*
Slightly modify waiters so that they call the underlying operation once before entering wait state

### DIFF
--- a/packages/util-waiter/src/poller.ts
+++ b/packages/util-waiter/src/poller.ts
@@ -13,7 +13,8 @@ const exponentialBackoffWithJitter = (minDelay: number, maxDelay: number, attemp
 const randomInRange = (min: number, max: number) => min + Math.random() * (max - min);
 
 /**
- * Function that runs indefinite polling as part of waiters.
+ * Function that runs polling as part of waiters. This will make one inital attempt and then
+ * subsequent attempts with an increasing delay.
  * @param params options passed to the waiter.
  * @param client AWS SDK Client
  * @param input client input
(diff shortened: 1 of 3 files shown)


yarn.lock
```diff
@@ -4,137 +4,117 @@
 
 "@aws-crypto/crc32@^1.0.0":
   version "1.0.0"
-  resolved "https://registry.yarnpkg.com/@aws-crypto/crc32/-/crc32-1.0.0.tgz#6a0164fd92bb365860ba6afb5dfef449701eb8ca"
+  resolved "http://localhost:4873/@aws-crypto%2fcrc32/-/crc32-1.0.0.tgz#6a0164fd92bb365860ba6afb5dfef449701eb8ca"
```
trivikr, line 7: Verdaccio needs to be disabled

trivikr (changes requested):

packages/util-waiter/src/poller.ts
```diff
@@ -24,6 +25,11 @@ export const runPolling = async <Client, Input>(
   input: Input,
   acceptorChecks: (client: Client, input: Input) => Promise<WaiterResult>
 ): Promise<WaiterResult> => {
+  const { state } = await acceptorChecks(client, input);
+  if (state !== WaiterState.RETRY) {
+    return { state };
+  }
```
trivikr, line 31: nit: move to a function so that it can be called from both outside and inside the loop.

trivikr (approved):

Issue #1919, opened 2021-01-15
Typescript 3.7.x breaking @aws-sdk/types


alexforsyth:
**Describe the bug**
`@aws-sdk/types` declares UserAgentPair.
```ts
export declare type UserAgentPair = [name: string, version?: string];
```

Pairs are not supported in Typescript 3.7.x
https://devblogs.microsoft.com/typescript/announcing-typescript-4-0/#labeled-tuple-elements.

Users running typescript < 4.0.0 cannot correctly get types. 

**SDK version number**
`3.3.0`

**Is the issue in the browser/Node.js/ReactNative?**
All

**Details of the browser/Node.js/ReactNative version**
node v14.15.4

**To Reproduce (observed behavior)**

The following error is thrown for the code example give below:
```console
$ yarn build
yarn run v1.22.5
$ tsc
node_modules/@aws-sdk/types/dist/cjs/util.d.ts:92:42 - error TS1005: ',' expected.

92 export declare type UserAgentPair = [name: string, version?: string];
                                            ~

node_modules/@aws-sdk/types/dist/cjs/util.d.ts:92:60 - error TS1005: ',' expected.

92 export declare type UserAgentPair = [name: string, version?: string];
                                                              ~


Found 2 errors.

error Command failed with exit code 2.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

<details>
<summary>package.json</summary>

```json
{
  "name": "test-ts-3.7",
  "version": "1.0.0",
  "main": "index.js",
  "license": "MIT",
  "scripts": {
    "build": "tsc"
  },
  "dependencies": {
    "@aws-sdk/client-s3": "3.3.0"
  },
  "devDependencies": {
    "@types/node": "14.14.22",
    "typescript": "3.7.5"
  }
}
```

</details>

<details>
<summary>index.ts</summary>

```ts
import { S3 } from "@aws-sdk/client-s3";

(async () => {
  const region = "us-west-2";
  const client = new S3({ region });
  await client.listBuckets({});
})();

```

</details>

<details>
<summary>tsconfig.json</summary>

```json
{
  "compilerOptions": {
    "target": "es5",
    "module": "commonjs",
    "strict": true,
    "esModuleInterop": true,
    "forceConsistentCasingInFileNames": true
  }
}
```

</details>

**Expected behavior**
No error is thrown when compiling client code with TypeScript version 3.7.x

trivikr:
[...]

PR #1937, opened 2021-01-21
fix(generate-clients): call mergeManifest when constructor.name is Object

trivikr:
*Issue #, if available:*
Refs: https://github.com/aws/aws-sdk-js-v3/issues/1919#issuecomment-764218895

*Description of changes:*
In https://github.com/aws/aws-sdk-js-v3/issues/1919, we decided to use [downlevel-dts](https://www.npmjs.com/package/downlevel-dts) to create type definition files which support TS 3.4 or later. The configuration for [typesVersions](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-1.html#version-selection-with-typesversions) contains an array which is replaced to an object by out copy-to-clients script as it returns true for typeof operation.

This PR changes the `typeof` check to `constructor.name` so that only objects are replaced in package.json

```console
$ node
Welcome to Node.js v14.15.4.
Type ".help" for more information.
> const arr = ["arr"];
undefined
> typeof arr
'object'
> arr.constructor.name
'Array'
> const obj = {"obj": "obj"};
undefined
> typeof obj
'object'
> obj.constructor.name
'Object'
> const str = "str";
undefined
> typeof str
'string'
> str.constructor.name
'String'
```

*Testing:*
Verified that generate-clients doesn't update any package.json because of this change.

### DIFF
--- a/scripts/generate-clients/copy-to-clients.js
+++ b/scripts/generate-clients/copy-to-clients.js
@@ -41,7 +41,7 @@ const getOverwritablePredicate = (packageName) => (pathName) => {
 const mergeManifest = (fromContent = {}, toContent = {}) => {
   const merged = {};
   for (const name of Object.keys(fromContent)) {
-    if (typeof fromContent[name] === "object") {
+    if (fromContent[name].constructor.name === "Object") {
       merged[name] = mergeManifest(fromContent[name], toContent[name]);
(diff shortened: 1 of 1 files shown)


alexforsyth (approved):
LGTM

PR #1950, opened 2021-01-22
docs(credential-provider-node): behavior when roleAssumer is not provided

dmattia:


trivikr:
Todo: verify if the documentation needs to be updated, and create new PR.

trivikr (approved):

PR #1951, opened 2021-01-23
feat(credential-provider-env): provide option for custom env prefix

alsiola:
Allows for the provision of a custom prefix that will replace `AWS` at the start of each read environment variable.

### Motivation
My use case is creating to separate S3 clients - one for generating pre-signed links, and one for in-service S3 operations. We would like to use separate credentials for each client, to ensure the pre-signed credentials do not expire prior to the link expiry, and to minimise the permissions given to the pre-signed link.
Currently to achieve this, I have to copy/paste the `credential-provider-env` package into my service, and use different env var names. By allowing options to `fromEnv`, I can use it more easily with a different prefix.

This is similar to the usage to the `EnvironmentCredentials` class in the v2 sdk (https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/EnvironmentCredentials.html)

I have not been able to run `generate-clients` locally - the following error:

```
Execution failed for task ':smithy-aws-typescript-codegen:compileJava'.
> Could not resolve all files for configuration ':smithy-aws-typescript-codegen:compileClasspath'.
   > Could not find software.amazon.smithy:smithy-typescript-codegen:0.3.0.
```

### DIFF
--- a/packages/credential-provider-env/README.md
+++ b/packages/credential-provider-env/README.md
@@ -20,3 +20,19 @@ the following environment variables:
 If either the `AWS_ACCESS_KEY_ID` or `AWS_SECRET_ACCESS_KEY` environment
 variable is not set or contains a falsy value, the promise returned by the
 `fromEnv` function will be rejected.
+
+### Environment variable prefixes
+
+An optional options object containing a `prefix` property can be provided, which will
+alter the environment variables from which credentials will be read. For example,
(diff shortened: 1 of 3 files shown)


alsiola:
Seems to be minimal interest in merging this, and I no longer have a personal use case. Closing.

PR #1955, opened 2021-01-24
feat(lib-storage): improve performance by reducing expensive buffer concat operations

monken:
*Issue #1841*

*Description of changes:*

The `for await` syntax reads chunks from a readable stream and is in my opinion more concise and less error prone as NodeJS will take care of event handlers, exception handling, etc.

In addition, `Buffer.concat` will only be called once per chunk which greatly improves performance (uploading a 1GB file would take 23 seconds previously, down to 10 seconds with this patch). The use of `Buffer.slice` is intentional as it does not copy or create a new buffer. Instead, it's basically a view into the buffer it slices over.

I ran a few profiling runs over the code and I think there is more room for improvement. V2 of the SDK still seems to be ~20 percent faster. For example, v3 calls `update` from `internal/crypto/hash.js` twice as many times as v2.

### DIFF
--- a/lib/storage/src/data-chunk/readable-helper.ts
+++ b/lib/storage/src/data-chunk/readable-helper.ts
@@ -4,60 +4,53 @@ import { Readable } from "stream";
 import { DEFAULT } from "../upload/defaults";
 import { DataPart } from "./yield-chunk";
 
+type Buffers = {
+  chunks: Buffer[];
+  length: number;
+};
+
 export async function* chunkFromReadable(reader: Readable, chunkSize: number): AsyncGenerator<DataPart, void, unknown> {
   let partNumber = DEFAULT.MIN_PART_NUMBER;
-  let oldBuffer = Buffer.from("");
-  while (partNumber < DEFAULT.MAX_PART_NUMBER) {
-    let currentBuffer = oldBuffer;
-    if (reader.readable) {
-      reader.resume();
-      currentBuffer = await _chunkFromStream(reader, chunkSize, oldBuffer);
-      reader.pause();
-    }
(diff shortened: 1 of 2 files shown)


samsullivan:
I don't have to dig deeper (yet), but replacing this with my last build (that I tested #1841 with) gave me the following error:
> Invalid client type

Happened with both `S3` and `S3Client` objects from `@aws-sdk/client-s3`. My last build was on `3.0.1`, though.

monken:
@samsullivan yeah make sure you update everything to 3.3.0. I had the same issue...

samsullivan:
Ah, should've been able to figure that out on my own. Looks like it's working now.

I didn't have any more robust performance testing; however, this seems to be more thought out of a fix than my "bolt it on" approach in #1841 so I'm fine with this one being merged instead. Bummer that v2 SDK still had better performance here, but...the current v3 implementation is unusable.

monken:
I have prepared another PR that will improve upload speeds further, even beating v2. It's a little more invasive by allowing passing Buffer arrays all the way down to the http-client and signer. I'll submit it if this PR gets merged.

AllanZhengYP (commented):
@monken
👍👍 Huge thanks to this optimization and the test case. It looks good to me! The formating of the code is a little off. The Prettier should've formatted it but for some reason it doesn't seem the case. Can you try format the code?

monken:
@AllanZhengYP for some reason the pre-commit hook didn't run. Should be fixed now.

alexforsyth:
This PR to rewrite Upload should incorporate these ideas. @monken please let me know if something is missing there.

PR #1990, opened 2021-02-02
fix(lib-storage): S3 Upload can corrupt data from readable stream

xfournet:
### Issue #
In some circumstance Upload miss some data from the readable stream and corrupt the uploaded data. This has been observed in a real situation when serializing data and uploading data on the fly by using a PassThrough between the serialization and the S3 Upload. The issue has been located in `lib/storage/src/data-chunk/readable-helper.ts`

### Description
In `readable-helper.ts` the pause/resume mechanism has a concurrency flaw between line
https://github.com/aws/aws-sdk-js-v3/blob/fa33f0b04b6c61e8305f07107f92da306d67242a/lib/storage/src/data-chunk/readable-helper.ts#L14
and
https://github.com/aws/aws-sdk-js-v3/blob/fa33f0b04b6c61e8305f07107f92da306d67242a/lib/storage/src/data-chunk/readable-helper.ts#L15

It can happen that stream can emit data that won't be process since no 'data' event listener is registered at that time.

Fix is to handle the pause/resume directly in _chunkFromStream in synchronous code, preventing the hole caused by await/aync code part.

PR #1955 also fix this problem.

### Testing
A new unit test has been added in `lib/storage/test/data-chunk/readable-helper.spec.ts` 

---

### DIFF
--- a/lib/storage/src/data-chunk/readable-helper.ts
+++ b/lib/storage/src/data-chunk/readable-helper.ts
@@ -10,9 +10,7 @@ export async function* chunkFromReadable(reader: Readable, chunkSize: number): A
   while (partNumber < DEFAULT.MAX_PART_NUMBER) {
     let currentBuffer = oldBuffer;
     if (reader.readable) {
-      reader.resume();
       currentBuffer = await _chunkFromStream(reader, chunkSize, oldBuffer);
-      reader.pause();
     }
 
     yield {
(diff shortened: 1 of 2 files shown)


alexforsyth:
This issue should be solved with: https://github.com/aws/aws-sdk-js-v3/pull/2039's for...await

alexforsyth:
Also awesome catch!! huge props to @xfournet. Can I ask for your feedback on https://github.com/aws/aws-sdk-js-v3/pull/2039?

xfournet:
Thanks @alexforsyth! #2039 should also fix the problem so it's fine for me!
Let me know if i need to close this PR ? Or you want to integrate it in the meantime until the delivery of #2039 ?

Issue #1959, opened 2021-01-26
Documentation states `maxSockets` is limited to 50


iwt-janbrauer:
**Describe the issue with documentation**
See https://docs.aws.amazon.com/sdk-for-javascript/v3/developer-guide/node-configuring-maxsockets.html.
It states:
> When using the default of https, the SDK takes the maxSockets value from the globalAgent. If the maxSockets value is not defined or is Infinity, the SDK assumes a maxSockets value of 50.

But this is only true for the old version of the SDK. v3 does not have [this logic](https://github.com/aws/aws-sdk-js/blob/master/lib/http/node.js#L156). Thus it will use `Infinity` and quickly exhaust the limit.

**To Reproduce (observed behavior)**
- Write a Lambda func with v3
- Make a lot of requests to DynamoDB (30000 items / PutItem)
- See `Error: connect EMFILE 52.94.17.26:443 - Local (undefined:undefined)` in the log output

**Expected behavior**
This shouldn't have happened, because the DynamoDB client should not run out of file handles.

**Screenshots**
```json
{
    "errorType": "Runtime.UnhandledPromiseRejection",
    "errorMessage": "Error: connect EMFILE 52.94.17.26:443 - Local (undefined:undefined)",
    "reason": {
        "errorType": "Error",
        "errorMessage": "connect EMFILE 52.94.17.26:443 - Local (undefined:undefined)",
        "code": "EMFILE",
        "errno": "EMFILE",
        "syscall": "connect",
        "address": "52.94.17.26",
        "port": 443,
        "$metadata": {
            "attempts": 1,
            "totalRetryDelay": 0
        },
        "stack": [
            "Error: connect EMFILE 52.94.17.26:443 - Local (undefined:undefined)",
            "    at internalConnect (net.js:923:16)",
            "    at defaultTriggerAsyncIdScope (internal/async_hooks.js:364:12)",
            "    at GetAddrInfoReqWrap.emitLookup [as callback] (net.js:1066:9)",
            "    at GetAddrInfoReqWrap.onlookup [as oncomplete] (dns.js:68:8)"
        ]
    },
    "promise": {},
    "stack": [
        "Runtime.UnhandledPromiseRejection: Error: connect EMFILE 52.94.17.26:443 - Local (undefined:undefined)",
        "    at process.<anonymous> (/var/runtime/index.js:35:15)",
        "    at process.emit (events.js:314:20)",
[...]

PR #2024, opened 2021-02-11
fix(node-http-handler)!: set maxSockets=50 by default

trivikr:
### Issue #
Fixes: https://github.com/aws/aws-sdk-js-v3/issues/1959

### Description
set maxSockets=50 by default in node-http-handler

### Testing
Unit tests

---

### DIFF
--- a/packages/node-http-handler/src/node-http-handler.ts
+++ b/packages/node-http-handler/src/node-http-handler.ts
@@ -42,8 +42,9 @@ export class NodeHttpHandler implements HttpHandler {
     this.connectionTimeout = connectionTimeout;
     this.socketTimeout = socketTimeout;
     const keepAlive = true;
-    this.httpAgent = httpAgent || new hAgent({ keepAlive });
-    this.httpsAgent = httpsAgent || new hsAgent({ keepAlive });
+    const maxSockets = 50;
+    this.httpAgent = httpAgent || new hAgent({ keepAlive, maxSockets });
+    this.httpsAgent = httpsAgent || new hsAgent({ keepAlive, maxSockets });
   }
 
   destroy(): void {
--- a/packages/node-http-handler/src/node-http-handler.spec.ts
+++ b/packages/node-http-handler/src/node-http-handler.spec.ts
@@ -18,15 +18,19 @@ import {
 describe("NodeHttpHandler", () => {
   describe("constructor", () => {
     it("can set httpAgent and httpsAgent", () => {
+      let keepAlive = false;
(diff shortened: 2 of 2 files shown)


packages/node-http-handler/src/node-http-handler.ts
```diff
@@ -42,8 +42,9 @@ export class NodeHttpHandler implements HttpHandler {
     this.connectionTimeout = connectionTimeout;
     this.socketTimeout = socketTimeout;
     const keepAlive = true;
-    this.httpAgent = httpAgent || new hAgent({ keepAlive });
-    this.httpsAgent = httpsAgent || new hsAgent({ keepAlive });
+    const maxSockets = 50;
```
TysonAndre, line 45: Would it make sense to allow passing this option alongside other options in the constructor `{..., maxSockets = 50 }`? AWS.config.update used to allow passing maxSockets if I remembered correctly, and this would help with high-traffic calls for slow API endpoints

It'd make migration from v2 easier, but it could be argued to be redundant if httpsAgent is supported

trivikr: Both the values for `keepAlive` and `maxSockets` can be set by user by passing a custom user agent as follows:

```js
const { DynamoDBClient } = require("@aws-sdk/client-dynamodb");
const { NodeHttpHandler } = require("@aws-sdk/node-http-handler");
var https = require("https");

var dynamodbClient = new DynamoDBClient({
  requestHandler: new NodeHttpHandler({
    httpsAgent: new https.Agent({
      keepAlive: false,
      maxSockets: 25,
    }),
  }),
});
```

This is documented in Developer Guide in [Configuring maxSockets in Node.js](https://docs.aws.amazon.com/sdk-for-javascript/v3/developer-guide/node-configuring-maxsockets.html)

Is this what you're looking for @TysonAndre ?

TysonAndre: >  but it could be argued to be redundant if httpsAgent is supported

I was already aware of that option - this was more of a question of whether the sdk itself should support that (e.g. for applications that used to pass a generic config object to AWS.config.update() and are now passing that to v3 constructors

trivikr: The modular AWS SDK for JavaScript (v3) doesn't support global configuration as documented in README: https://github.com/aws/aws-sdk-js-v3/#configuration

alexforsyth (approved):
LGTM!

Issue #1973, opened 2021-01-29
Upload constructor in @aws-sdk/lib-storage appears to be broken


ffxsam:
## Describe the bug
Using `new Upload()` from `@aws-sdk/lib-storage` throws an error of "Invalid client type" when passed a valid S3 instance. This only happens when `@aws-sdk/client-s3` is also installed.

#### SDK version number
`@aws-sdk/lib-storage@3.4.0
`@aws-sdk/client-s3@3.4.0`

#### Is the issue in the browser/Node.js/ReactNative?
Node.js

#### Details of the browser/Node.js/ReactNative version
v12.20.1

## Steps to reproduce
```js
import { S3 } from '@aws-sdk/client-s3';

// ...

const upload = new Upload({
  client: new S3({}),
  params: {
    Bucket: process.env.STORAGE_BUCKET,
    Key: zipFileKey,
    Body: readableStream,
  },
});
```

#### Observed behavior
An error is thrown: "Invalid client type"

#### Expected behavior
I would expect it to not throw an error.

ffxsam:
Ok, I discovered `@aws-sdk/lib-storage`. However, there's a bug. My code is below:

```js
    console.log('this is true:', s3 instanceof S3);
    const upload = new Upload({
      client: s3,
      params: {
        Bucket: process.env.DEST_BUCKET,
        Key: `${destPrefix}snippet.mp3`,
        Body: ffmpegProc.stdout,
      },
    });
```
So `s3` is definitely an instance of `S3`, and yet I get this error:
```
{
    "errorMessage": "Invalid client type"
}
```

schantaraud:
You may want to uninstall your local version of `@aws-sdk/client-s3` and let `@aws-sdk/lib-storage` install it as a dependency.

ffxsam:
VERY odd. It suddenly works, with no changes on my end. It's possible a package update fixed something.

ffxsam:
I ran into this issue again. Uninstalling the `@aws-sdk/client-s3` package fixed it, but that's a weird solution. These packages shouldn't cause conflicts like that when they're both installed.

@trivikr @AllanZhengYP

trivikr:
@ffxsan Please update the description of this issue with [bug report](https://github.com/aws/aws-sdk-js-v3/blob/master/.github/ISSUE_TEMPLATE/---bug-report.md) template so that it's easy to triage and debug.

ffxsam:
@trivikr Done!

[...]

Issue #1891, opened 2021-01-08
Cannot set Metadata when using lib-storage `Upload`.


peterwmwong:
**Describe the bug**

`@aws-sdk/lib-storage`'s `Upload` ignores `Metadata` parameter.

**SDK version number**: `@aws-sdk/lib-storage = 3.1.0`

**Is the issue in the browser/Node.js/ReactNative?**
Node.js

**Details of the browser/Node.js/ReactNative version**

```
> node -v
v15.4.0
```

**To Reproduce (observed behavior)**

```ts
import { createReadStream } from 'fs';
import { Upload } from '@aws-sdk/lib-storage';
import { S3 } from '@aws-sdk/client-s3';

(async () => {
    const fileStream = createReadStream(__filename);
    try {
        const upload = new Upload({
            client: new S3({}),
            params: {
                Bucket: 'test-bucket',
                Key: 'test-key',
                Body: fileStream,
                // vvv BUG: Currently ignored! vvv
                Metadata: {
                    testMetaData1: 'one',
                    testMetaData2: 'two',
                },
            },
        });

        await upload.done();
    } catch (e) {
        console.log(e);
    }
})();
```

**Actual behavior**

<img width="773" alt="Screen Shot 2021-01-07 at 10 12 25 AM" src="https://user-images.githubusercontent.com/284734/104029256-5ef1a500-518f-11eb-9072-91fd796ffbcd.png">

**Expected behavior**

Specified `Metadata` is added to the uploaded S3 object: (`x-amx-meta-testmetadata1` and `x-amx-meta-testmetadata2`).

<img width="739" alt="Screen Shot 2021-01-07 at 10 15 03 AM" src="https://user-images.githubusercontent.com/284734/104029284-67e27680-518f-11eb-9366-c5e7024131e5.png">

**Additional context**

I believe the problem is `Uploader.intialize()` is currently NOT considering all the properties provided by `PutObjectCommandInput` (in this case `Metadata`).

https://github.com/aws/aws-sdk-js-v3/blob/d4c302b816e1781f8d04bd479cc4e26e0fe4debc/lib/storage/src/upload/service-clients/Uploader.ts#L50-L57

A monkey patch workaround, we've been using is the following

```diff
import { createReadStream } from 'fs';
import { Upload } from '@aws-sdk/lib-storage';
import { S3 } from '@aws-sdk/client-s3';
[...]

Issue #1688, opened 2020-11-15
@aws-sdk/lib-storage - fs, stream resolution in browser


glynd:
**Describe the bug**
Trying to use this in a browser (Angular) test app and getting missing dependencies (appear to be nodejs components)

```console
Error: ./node_modules/@aws-sdk/lib-storage/dist/es/utils/bytelength.js
Module not found: Error: Can't resolve 'fs' in '/XXXXXXXXXXX/node_modules/@aws-sdk/lib-storage/dist/es/utils'

Error: ./node_modules/@aws-sdk/lib-storage/dist/es/data-chunk/yield-chunk.js
Module not found: Error: Can't resolve 'stream' in '/XXXXXXXXXXX/node_modules/@aws-sdk/lib-storage/dist/es/data-chunk'
```

**SDK version number**

1.0.0-rc.6

**Is the issue in the browser/Node.js/ReactNative?**
Browser

**Details of the browser/Node.js/ReactNative version**
  Browsers:
    Chrome: 86.0.4240.198
    Safari: 14.0.1

**To Reproduce (observed behavior)**

I suspect there are easier ways to reproduce... but this will do.

Run:

```consle
ng new uploaderbug
cd uploaderbug
npm i @aws-sdk/lib-storage

edit src/app/app.component.ts 
to:

8<----------8<----------8<----------8<----------8<----------8<----------
import { Component } from '@angular/core';

import * as S3 from '@aws-sdk/client-s3';
import * as LibStorage from '@aws-sdk/lib-storage';


@Component({
  selector: 'app-root',
  templateUrl: './app.component.html',
  styleUrls: ['./app.component.css']
})
export class AppComponent {
  title = 'uploadbug';

  private dummyfunc() {

     const s3Client = new S3.S3Client({region: "eu-west-1"});

     const target :S3.PutObjectCommandInput = {Bucket: "dummybucket", Key: "dummykey"};

     const fileUpload = new LibStorage.Upload({client: s3Client, params: target});

  }
}
8<----------8<----------8<----------8<----------8<----------8<----------
```

Run:
ng serve

Result as above

**Expected behaviour**
Not expecting dependencies of nodejs components - is it currently not browser compatible?

**Screenshots**

**Additional context**
Didn't see any issues like this using the v2 Javascript SDK

alexforsyth:
Interesting, this should be compatible with browsers. Investigating

alexforsyth:
[...]

PR #2039, opened 2021-02-16
feat(lib-storage): rewrite lib-storage upload

alexforsyth:
### Description
As part of a review of our codebase I rewrote Upload(). This PR fixes a range of issues and should address most of the Issues filed (below). This rewrite is backwards compatible.

Upload is a wrapper on s3.multipart upload. This is largely analogous to managed upload in V2 of the JS SDK. 
Here's how you use Upload:
```js
import { Upload } from "@aws-sdk/lib-storage";

(async () => {
  const upload = new Upload({
    params: {
      Bucket: configuration.Bucket,
      Key: configuration.Key,
      Body: String | Buffer | Uint8Array | Stream | ReadableStream,
    },
    client: new S3({}),
    queueSize: 3,
  });

  upload.on("httpUploadProgress", (progress) => {
    console.log(progress);
  });

  await upload.done();
})();

```

### Testing
How was this change tested?
- [x] Test cases enumerating chunking
- [x] Test cases making sure multipart upload is called as expected
- [x] Upload should work if you pass in a version of S3 that is older than the version of S3 that Upload depends on. https://github.com/aws/aws-sdk-js-v3/issues/1973, https://github.com/aws/aws-sdk-js-v3/issues/1729
- [x] Upload should allow you to set MetaData: https://github.com/aws/aws-sdk-js-v3/issues/1891

#### Manual Browser testing
- [x] correctly uploads a string
- [x] correctly uploads an ArrayBuffer
- [x] correctly uploads a stream of unknown length (audio stream)
- [x] correctly provides status updates 

#### Manual Node testing
- [x] correctly uploads a string
- [x] correctly uploads an Buffer
- [x] correctly uploads a stream of unknown length (audio stream)
- [x] correctly provides status updates 

### Additional context

PRs:
Fixes: https://github.com/aws/aws-sdk-js-v3/pull/1990
Fixes: https://github.com/aws/aws-sdk-js-v3/pull/1955

Issues:
Fixes: https://github.com/aws/aws-sdk-js-v3/issues/1973
Fixes: https://github.com/aws/aws-sdk-js-v3/issues/1891
Fixes: https://github.com/aws/aws-sdk-js-v3/issues/1729
Fixes: https://github.com/aws/aws-sdk-js-v3/issues/1688

---


monken:
Hi @alexforsyth, will you incorporate the performance improvements from my PR at https://github.com/aws/aws-sdk-js-v3/pull/1955 ?

alexforsyth:
@monken yes! I'll request you as a reviewer when this PR is in better shape. Everything should be using for...await syntax everywhere here.

alexforsyth:
@monken it looks like I cannot add you as a reviewer to this PR. I'd still love your feedback at any rate

samsullivan:
Thank you so much for your attention here, since I only was able to give partial care previously and we've been waiting for these performance improvements to switch over to the new SDK.

If you care, I'll get this tested with my use case some point this week.

alexforsyth:
@samsullivan that'd be great! We're looking to get this out with the next release

lib/storage/package.json
```diff
@@ -23,11 +23,11 @@
   },
   "license": "Apache-2.0",
   "dependencies": {
+    "@aws-sdk/abort-controller": "3.4.1",
```
xfournet, line 26: shouldn't be 3.5.0 ?

lib/storage/src/chunks/getChunkStream.ts
```diff
@@ -0,0 +1,39 @@
...
+  let partNumber = 1;
+  const currentBuffer: Buffers = { chunks: [], length: 0 };
+
+  for await (const datum of getNextData(data)) {
+    currentBuffer.chunks.push(datum);
+    currentBuffer.length += datum.length;
+
+    while (currentBuffer.length >= partSize) {
+      // concat all the buffers together once.
+      const dataChunk: Buffer = Buffer.concat(currentBuffer.chunks);
```
xfournet, line 23: it look like to me that it should be done before the while loop. Indeed current.chunks is a single element array after the first iteration (line 30) so Buffer.concat seems useless for the 2nd and next iterations

lib/storage/src/chunks/getChunkStream.ts
```diff
@@ -0,0 +1,39 @@
...
+  length: number;
+}
+
+export async function* getChunk<T>(
+  data: T,
+  partSize: number,
+  getNextData: (data: T) => AsyncGenerator<Buffer>
+): AsyncGenerator<RawDataPart> {
+  let partNumber = 1;
+  const currentBuffer: Buffers = { chunks: [], length: 0 };
```
xfournet, line 15: i think you could keep only one buffer and simplify a bit the code, eg:

```
  let partNumber = 1;
  let currentBuffer = Buffer.of();

  for await (const datum of getNextData(data)) {
    currentBuffer = Buffer.concat([currentBuffer, datum]);

    while (currentBuffer.length >= partSize) {
      yield {
        partNumber,
        data: currentBuffer.slice(0, partSize),
      };

      //update the buffer
      currentBuffer = currentBuffer.slice(partSize);
      partNumber += 1;
    }
  }
  yield {
    partNumber,
    data: currentBuffer,
  };
```

monken: See https://github.com/aws/aws-sdk-js-v3/pull/1955 why this is a bad idea. Buffer.concat is surprisingly slow and should be avoided at all cost. Doing it in a loop is even worse as you copy the buffers over and over as they grow larger. Also increasing the memory footprint of the node process as it might not get a chance to GC.

monken:
@alexforsyth it looks good to me and thank you very much for incorporating my changes.

monken:
Is there already a benchmark framework for this repository? I think it would be helpful to have one and also integrated in the build pipeline to ensure that performance doesn't regress. People will try and clean up or simplify the code without noticing that is serious performance implications.

lib/storage/src/chunks/getChunkStream.ts
alexforsyth: You're right, nice catch, lemme wrap this in a branch

lib/storage/src/chunks/getChunkStream.ts
alexforsyth: Yep, building off that, thats why this tries to wait until the last possible moment to run the concat.

alexforsyth:
@monken we do not have a benchmark framework quite yet! We should be benchmarking a bunch of things like bundle size, perf, etc. Opened a feature request here. Please feel free to contribute to that guy. https://github.com/aws/aws-sdk-js-v3/issues/2044

lib/storage/package.json
alexforsyth: The newest version of abort controller is 3.4.1. https://www.npmjs.com/package/@aws-sdk/abort-controller. We only bump packages when needed because otherwise we end up publishing a bunch of packages with zero diffs from previous versions. 

This might be a bit confusing since the versions aren't exactly following one another. But we do also have an @latest tag to alleviate it.


**eg.**  its entirely possible in the future to only have versions `3.6.0` and `3.9.0` of ` "@aws-sdk/abort-controller"` published. Without the minor versions in between them (`3.7.0`, `3.8.0`).

lib/storage/src/Upload.ts
```diff
@@ -0,0 +1,185 @@
...
+    if (this.uploadEvent) {
+      this.emit(this.uploadEvent, progress);
+    }
+  }
+
+  async __abortTimeout(abortSignal: AbortSignal): Promise<ServiceOutputTypes> {
+    return new Promise((resolve) => {
+      abortSignal.onabort = () => resolve({} as ServiceOutputTypes);
+    });
+  }
```
trivikr, line 184: abortTimeout should reject onabort.
```suggestion
  async __abortTimeout(abortSignal: AbortSignal): Promise<unknown> {
    return new Promise((reject) => {
      abortSignal.onabort = () => {
        const abortError = new Error("Upload aborted.");
        abortError.name = "AbortError";
        reject(abortError);
      };
    });
  }
```

lib/storage/src/Upload.ts
```diff
@@ -0,0 +1,185 @@
...
+    // Create and start concurrent uploads
+    for (let index = 0; index < this.queueSize; index++) {
+      const currentUpload = this.__doConcurrentUpload(dataFeeder, index);
+      this.concurrentUploaders.push(currentUpload);
+    }
+
+    // Create and start concurrent uploads
+    await Promise.all(this.concurrentUploaders);
+    if (this.abortSignal.signal.aborted) {
+      return {} as ServiceOutputTypes;
```
trivikr, line 156:
```suggestion
      throw Object.assign(new Error("Upload aborted."), { name: "AbortError" });
```

lib/storage/src/Upload.ts
```diff
@@ -0,0 +1,185 @@
...
+        let partResult = await this.client.send(
+          new UploadPartCommand({
+            ...this.params,
+            UploadId: this.uploadId,
+            Body: dataPart.data,
+            PartNumber: dataPart.partNumber,
+          })
+        );
+        if (this.abortSignal.signal.aborted) {
+          return;
```
trivikr, line 114: This line should throw AbortError
```suggestion
          throw Object.assign(new Error("Upload aborted."), { name: "AbortError" });
```

trivikr (changes requested):

lib/storage/src/Upload.ts
```diff
@@ -0,0 +1,185 @@
...
+  data: BodyDataTypes;
+}
+
+const MIN_PART_SIZE = 1024 * 1024 * 5;
+
+export class Upload extends EventEmitter {
+  /**
+   * S3 multipart upload does not allow more than 1000 parts.
+   */
+  private MAX_PARTS = 1000;
```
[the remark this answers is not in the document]
alexforsyth: Lol, nice! Yep thats the correct value

lib/storage/src/Upload.ts
alexforsyth: I dont think each of the concurrent uploaders should throw an abort error. This return is here as part of clean up. Otherwise we would propagate 5 errors to the user by default.

4 from the 4 default concurrent uploaders and 1 from the main multipart upload manager.

lib/storage/src/Upload.ts
alexforsyth: Yep! We totally should. Adding

lib/storage/src/Upload.ts
```diff
@@ -0,0 +1,185 @@
...
+  async done(): Promise<ServiceOutputTypes> {
+    return await Promise.race([this.__doMultipartUpload(), this.__abortTimeout(this.abortSignal.signal)]);
+  }
+
+  on(event: "httpUploadProgress", listener: (progress: Progress) => void): any {
+    this.uploadEvent = event;
+    super.on(event, listener);
+  }
+
+  async __doConcurrentUpload(dataFeeder: AsyncGenerator<RawDataPart>, id: number): Promise<void> {
```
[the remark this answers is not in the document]
alexforsyth: nice. fixing.

lib/storage/src/Upload.ts
```diff
@@ -0,0 +1,185 @@
...
+
+  private client: ServiceClients;
+  private params: PutObjectCommandInput;
+
+  // used for reporting progress
+  private totalBytes?: number;
+  private bytesUploadedSoFar: number;
+
+  // used in the upload
+  private abortSignal: AbortController;
```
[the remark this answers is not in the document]
alexforsyth: good call. This is a much better name

lib/storage/src/Upload.ts
```diff
@@ -0,0 +1,185 @@
...
+
+const MIN_PART_SIZE = 1024 * 1024 * 5;
+
+export class Upload extends EventEmitter {
+  /**
+   * S3 multipart upload does not allow more than 1000 parts.
+   */
+  private MAX_PARTS = 1000;
+
+  // Defaults
```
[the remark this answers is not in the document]
alexforsyth: done for all the comments in the PR

lib/storage/src/Upload.ts
```diff
@@ -0,0 +1,185 @@
...
+  async __doConcurrentUpload(dataFeeder: AsyncGenerator<RawDataPart>, id: number): Promise<void> {
+    for await (const dataPart of dataFeeder) {
+      if (this.uploadedParts.length > this.MAX_PARTS) {
+        throw new Error(
+          `Exceeded ${this.MAX_PARTS} as part of the upload to ${this.params.Key} and ${this.params.Bucket}.`
+        );
+      }
+
+      try {
+        let partResult = await this.client.send(
```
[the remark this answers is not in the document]
alexforsyth: done

lib/storage/src/Upload.ts
```diff
@@ -0,0 +1,185 @@
...
+        // otherwise swallow the error
+        if (this.leavePartsOnError) {
+          throw e;
+        }
+      }
+    }
+  }
+
+  async __doMultipartUpload(): Promise<ServiceOutputTypes> {
+    let createMultipartUploadResult = await this.client.send(new CreateMultipartUploadCommand(this.params));
```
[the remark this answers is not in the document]
alexforsyth: done

lib/storage/src/chunks/getChunkStream.ts
alexforsyth: done

lib/storage/package.json
```diff
@@ -49,6 +50,14 @@
       ]
     }
   },
+  "browser": {
+    "stream": "stream-browserify",
+    "fs": "./src/runtimeConfig.browser",
```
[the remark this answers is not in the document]
alexforsyth: fs doesn't exist in the browser, this should probably point to a file that exports an empty object by default. This is just here so browsers don't get the fs is not defined issue detailed here: https://github.com/aws/aws-sdk-js-v3/issues/1688

lib/storage/src/Upload.ts
alexforsyth: actually.... doing some testing .... If we reject, the abort Timeout throws an error to the parent of Upload. I dont think we should throw again here. 
1. This branch is a shortcut to clean up once the the uploader has entered an abort state
2. Its not possible for the top level `Upload.done()` to throw multiple times (i'm 99% sure - and if it threw more than once for a single abort that would be scary too).  Since we throw from abortTimeout and thats triggered immediately in the top level promise race the exception will always come from that abortTimeout function. Errors thrown additional times from inside the concurrent uploader or the main uploader wont ever been seen. They may as well just return in my opinion. No reason I see to throw the abort error from multiple places. Especially if those secondary places have no way of making their error visible.

What do you think? @trivikr

lib/storage/src/chunks/getChunkStream.ts
xfournet: you're right, i miss that point!

markb-trustifi:
Please check that it supports server-side encryption with `SSECustomerKey`.

alexforsyth:
@markb-trustifi it should, all the params you normally pass to multipart upload should work just the same as they are simply passed through to the underlying client.

lib/storage/src/Upload.ts
alexforsyth: Spoke offline with trivikr about this. We'll only be throwing from the abortTimeout and using checks here to clean up promises in flight.

lib/storage/package.json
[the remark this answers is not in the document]
alexforsyth: Actually I think this is fine since the browser runtimeconfig doesn't actually include anything as a default export.

lib/storage/src/Upload.ts
alexforsyth: addressed

lib/storage/src/Upload.ts
alexforsyth: Addressed

lib/storage/src/chunks/getChunkStream.ts
```diff
@@ -0,0 +1,43 @@
...
+
+  for await (const datum of getNextData(data)) {
+    currentBuffer.chunks.push(datum);
+    currentBuffer.length += datum.length;
+
+    while (currentBuffer.length >= partSize) {
+      // Concat all the buffers together once.
+      let dataChunk: Buffer = currentBuffer.chunks[0];
+      if(currentBuffer.chunks.length) {
```
trivikr, line 24: looks like prettier is not run on this file. Commenting to verify prettier is run on all files.
Previous discussion on this issue: https://github.com/aws/aws-sdk-js-v3/issues/1888

```suggestion
      if (currentBuffer.chunks.length) {
```

lib/storage/src/chunks/getChunkStream.ts
```diff
@@ -0,0 +1,43 @@
...
+        dataChunk = Buffer.concat(currentBuffer.chunks);
+      }
```
trivikr, line 26: The chunks array is never going to be empty in this LoC as:
* The chunks and length is populated on lines 18 and 19. The while loop will not run if both `currentBuffer.chunks` and `datum` are emoty.
* The while loop enters only when `currentBuffer.length >= partSize`. Line 35 ensures that while loop exits in case currentBuffer.length was multiple of partSize (only case when `currentBuffer.chunks.length` can be empty)

If `currentBuffer.chunks` is empty, line 27 is going to throw `TypeError: Cannot read property 'slice' of undefined` any way.

```suggestion
      const dataChunk = Buffer.concat(currentBuffer.chunks);
```

lib/storage/src/chunks/getChunkStream.ts
```diff
@@ -0,0 +1,43 @@
...
+interface Buffers {
+  chunks: Buffer[];
+  length: number;
+}
+
+export async function* getChunk<T>(
+  data: T,
+  partSize: number,
+  getNextData: (data: T) => AsyncGenerator<Buffer>
+): AsyncGenerator<RawDataPart> {
```
trivikr, line 13: Updates:
* value `void` for `TReturn`: The generator function falls off the end without a return and the caller gets back `{ value: undefined, done: true }`.
* value `undefined` for `TNext`: next function does not accept any parameters.
```suggestion
): AsyncGenerator<RawDataPart, void, undefined> {
```

lib/storage/src/chunks/getChunkBuffer.ts
```diff
@@ -0,0 +1,22 @@
+import { RawDataPart } from "../Upload";
+
+export async function* getChunk(data: Buffer, partSize: number): AsyncGenerator<RawDataPart> {
```
trivikr, line 3: Updates:
* value `void` for `TReturn`: The generator function falls off the end without a return and the caller gets back `{ value: undefined, done: true }`.
* value `undefined` for `TNext`: next function does not accept any parameters.

```suggestion
export async function* getChunk(data: Buffer, partSize: number): AsyncGenerator<RawDataPart, void, undefined> {
```

lib/storage/src/chunks/getChunkBuffer.ts
```diff
@@ -0,0 +1,22 @@
...
+  let partNumber = 1;
+  let startByte = 0;
+  let endByte = startByte + partSize;
```
trivikr, line 6: The value in `startByte` is 0 and is not required to be added to `endByte`.
```suggestion
  let endByte = partSize;
```

trivikr (changes requested):

lib/storage/src/chunker.ts
```diff
@@ -0,0 +1,29 @@
+import stream from "stream";
+import { Buffer } from "buffer";
+
+import { BodyDataTypes } from "./types";
+import { getChunk as getChunkBuffer } from "./chunks/getChunkBuffer";
+import { getChunk as getChunkStream } from "./chunks/getChunkStream";
+import { getNextData as getNextDataReadableStream } from "./chunks/getDataReadableStream";
+import { getNextData as getNextDataReadable } from "./chunks/getDataReadable";
```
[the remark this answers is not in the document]
alexforsyth: Yeah not sure. This is not run for me.

lib/storage/src/chunks/getChunkStream.ts
alexforsyth: You're right that this should be `currentBuffer.chunks.length > 1`. But this branch is still needed. For example:

```
currentBuffer = {
    chunks: [.... length 55 ...]
    length: 55
} 
```

And we return chunks of size 10. We should only call concat if there are multiple chunks to concat. Otherwise we call concat 5 more times than needed on an array of size 1

lib/storage/src/chunks/getChunkBuffer.ts
alexforsyth: I'll add this for verboseness, but I dont think this is strictly needed right? These types are implied. Are you getting type errors?

lib/storage/src/chunker.ts
```diff
@@ -0,0 +1,29 @@
...
+
+export const getChunk = (data: BodyDataTypes, partSize: number) => {
+  if (data instanceof Buffer) {
+    return getChunkBuffer(data, partSize);
+  } else if (data instanceof stream.Readable) {
+    return getChunkStream<stream.Readable>(data, partSize, getNextDataReadable);
+  } else if (data instanceof String || typeof data == "string" || data instanceof Uint8Array) {
```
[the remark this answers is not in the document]
alexforsyth: nice catch!

lib/storage/src/Upload.ts
```diff
@@ -0,0 +1,202 @@
...
+        `EntityTooSmall: Your proposed upload partsize [${this.partSize}] is smaller than the minimum allowed size [${MIN_PART_SIZE}] (5MB)`
+      );
+    }
+
+    if (this.queueSize < 1) {
+      throw new Error(`Queue size: Must have atleast one uploading queue.`);
+    }
+
+    this.client = options.client;
+    this.params = options.params;
```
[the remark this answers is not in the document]
alexforsyth: Good call. I'll add these

lib/storage/src/chunks/getChunkBuffer.ts
trivikr: They're implied to `any` and `unknown` which may cause issues in future https://github.com/microsoft/TypeScript/blob/2428ade1a91248e847f3e1561e31a9426650efee/src/lib/es2018.asyncgenerator.d.ts#L3

I'm not getting any type errors right now.

lib/storage/src/chunks/getDataReadableStream.ts
```diff
@@ -0,0 +1,20 @@
...
+  try {
+    while (true) {
+      // Read from the stream.
+      const { done, value } = await reader.read();
+      // Exit if we're done.
+      if (done) return;
+      // Else yield the chunk.
+      yield Buffer.from(value);
+    }
+  } finally {
```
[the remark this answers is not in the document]
alexforsyth: I dont think the catch block is needed since this would be an invalid readableStream and caught a level higher (in the chunker). But i'll add it anyways. I agree it's probably best for this function do that.

lib/storage/src/chunks/getDataReadable.ts
```diff
@@ -0,0 +1,8 @@
+import Stream from "stream";
+import { Buffer } from "buffer";
+
+export async function* getNextData(data: Stream.Readable): AsyncGenerator<Buffer> {
```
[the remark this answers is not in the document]
alexforsyth: agree

lib/storage/src/chunks/getChunkStream.ts
trivikr: Sounds good!

How about using const instead of let like this with a comment about Buffer.concat?
```suggestion
      // Buffer.concat is slow operation and should be avoided for performance improvement.
      const dataChunk = currentBuffer.chunks.length > 1 ? Buffer.concat(currentBuffer.chunks) : currentBuffer.chunks[0];
```

lib/storage/src/chunker.ts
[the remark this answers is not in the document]
alexforsyth: manually ran linter, although it did not reorder imports. I think this should be done in another PR. Seems like its not running on commit for some reason for me.

lib/storage/src/chunks/getChunkStream.ts
alexforsyth: Added comment. Sure i'll update this to a one-line branch

lib/storage/src/chunks/getChunkBuffer.ts
alexforsyth: okay gotcha. These are updated

lib/storage/src/chunks/getChunkBuffer.ts
alexforsyth: done

lib/storage/src/chunks/getDataReadableStream.ts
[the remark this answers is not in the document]
alexforsyth: done

lib/storage/src/chunker.ts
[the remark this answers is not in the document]
trivikr: Created issue: https://github.com/aws/aws-sdk-js-v3/issues/2050

samsullivan (commented):
Don't have time to review the code changes, but I :100: agree that this needed to be rewritten and I appreciate your work. I built you're branch in my test build and threw this into our stage AWS environment just now and it worked fine!

Super excited to have this released. v2 of the JS SDK has an ECS Fargate bug where it doesn't use the IAM role on the container, so I'll get to remove some access keys once upgraded :slightly_smiling_face:

lib/storage/src/bytelength.ts
```diff
@@ -5,15 +5,18 @@ const toBuffer = (data: any) => {
 export const byteLength = (input: any) => {
   if (input === null || input === undefined) return 0;
   if (typeof input === "string") input = toBuffer(input);
```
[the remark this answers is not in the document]
alexforsyth: done

lib/storage/src/Upload.ts
```diff
@@ -0,0 +1,216 @@
...
+    if (this.abortController.signal.aborted) {
+      throw Object.assign(new Error("Upload aborted."), { name: "AbortError" });
+    }
+
+    this.uploadedParts.sort((a, b) => a.PartNumber! - b.PartNumber!);
+    const completeMultipartUpload = await this.client.send(
+      new CompleteMultipartUploadCommand({
+        ...this.params,
+        Bucket: this.params.Bucket,
+        Key: this.params.Key,
```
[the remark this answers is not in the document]
alexforsyth: done

trivikr (approved):

alexforsyth:
@monken @samsullivan @xfournet Thanks so much for the engagement. It's awesome to see the involvement. We're aiming at releasing these changes soon

alexforsyth:
FYI: This should be out with `3.6.0`

skozer:
Confirmed working in 3.6.0, thanks!

samsullivan:
:point_up: same here, really appreciate the effort by everybody on this one. Happy to be off of v2 SDK!

PR #2043, opened 2021-02-17
fix(middleware-sdk-ec2): add undeclared dependency @aws-sdk/protocol-http

blake-transcend:
### Issue #
<none>

### Description
There is an undeclared dependency on `@aws-sdk/protocol-http` in https://github.com/blake-transcend/aws-sdk-js-v3/blob/patch-2/packages/middleware-sdk-ec2/src/index.ts#L1. This adds the dependency to the package.json file

### Testing
tested locally

### Additional context
<none>

---


trivikr (approved):
Thanks for the catch!

Issue #1032, opened 2020-03-23
Use ts-jest so that test files can be run without transpiling


trivikr:
**Is your feature request related to a problem? Please describe.**
The test files `*.spec.ts` are currently transpiled to JavaScript before being run

**Describe the solution you'd like**
Use ts-jest so that `*.spec.ts` files can be run without transpiling

**Describe alternatives you've considered**
Continue transpiling files before running tests

trivikr:
compilerOptions.paths needs to be set up in order to use ts-jest:
* Documentation: https://kulshekhar.github.io/ts-jest/user/config/#paths-mapping
* Related: https://github.com/aws/aws-sdk-js-v3/issues/1026

Syy0n:
https://jestjs.io/blog/2019/01/25/jest-24-refreshing-polished-typescript-friendly
jest v24+ already has native support for typescript.

trivikr:
> One caveat to the default TypeScript support, similar to Flow, is that Babel will only strip out the type annotations to make your code valid JavaScript. It will not typecheck your code. 
> If you want to run typechecks while you test, you should use ts-jest

The native support for TypeScript in Jest 24+ won't typecheck the code.

PR #2088, opened 2021-02-25
feat: use ts-jest for running jest tests

trivikr:
### Issue
Fixes: https://github.com/aws/aws-sdk-js-v3/issues/1032

### Description
Runs all jest tests using ts-jest. This removes the need of transpiling `*.spec.ts` files allowing developers to build source just once, and run tests without transpling.

Caveat: The build command is removed from pretest for individual packages. Developers will have to explicitly run `build` instead of `pretest` before running tests.

### Testing
Verified that jest tests are successfully run after the build is successful.
```console
$ yarn test:all

$ yarn test:integration

# in packages/abort-controller
$ yarn test

# in lib/lib-storage
$ yarn test

```

---


alexforsyth (approved):
Some discussion offline. LGTM!

PR #2141, opened 2021-03-16
fix(credential-provider-cognito-identity): add identityId to return type of fromCognitoIdentity

inokawa:
### Description
In #1635, identityId was added to return value of `fromCognitoIdentity` and `fromCognitoIdentityPool` but not to the return type.

### Testing
Check tsc does not report error.

### Additional context

---


AllanZhengYP (approved):
⛴⛴

PR #2200, opened 2021-03-31
Ssdk rest json protocol tests

JordonPhillips:
This updates the generated protocol tests for rest-json so that the ssdk tests are being generated. The first commit has the actual change, the second is the updated generated code.

Note that to run these you still need to install the `smithy-server-common` manually since it's not published yet.

---


adamthom-amzn (approved):
I assume the build failure is due to the missing aformentioned server-common?

PR #2214, opened 2021-04-06
fix(util-user-agent-browser): `window` can be undefined

pkit:
### Description
In sandboxed environments (pure v8) `window` can be undefined. Should not fail.
Actually should always check for `globalThis` instead, but IE 11 is still alive....

---


AllanZhengYP (approved):
Looks good to me! :shipit:

PR #2220, opened 2021-04-08
Differentiate server serde

JordonPhillips:
This adds differentation for empty body behavior in restJson1. For
responses this protocol requires always sending an empty json object
if there is some shape defined as output but nothing bound to the
body or payload.

### Additional context

depends on: https://github.com/awslabs/smithy-typescript/pull/305

---

### DIFF
--- a/codegen/smithy-aws-typescript-codegen/src/main/java/software/amazon/smithy/aws/typescript/codegen/RestJsonProtocolGenerator.java
+++ b/codegen/smithy-aws-typescript-codegen/src/main/java/software/amazon/smithy/aws/typescript/codegen/RestJsonProtocolGenerator.java
@@ -89,8 +89,11 @@ abstract class RestJsonProtocolGenerator extends HttpBindingProtocolGenerator {
     @Override
     protected void writeDefaultHeaders(GenerationContext context, Shape operationOrError, boolean isInput) {
         super.writeDefaultHeaders(context, operationOrError, isInput);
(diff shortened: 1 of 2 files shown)


codegen/smithy-aws-typescript-codegen/src/main/java/software/amazon/smithy/aws/typescript/codegen/RestJsonProtocolGenerator.java
```diff
@@ -85,27 +85,33 @@ public void generateSharedComponents(GenerationContext context) {
...
-        super.writeDefaultErrorHeaders(context, error);
-        context.getWriter().write("'x-amzn-errortype': $S,", error.getId().getName());
+    protected void writeDefaultHeaders(GenerationContext context, Shape operationOrError, boolean isInput) {
+        super.writeDefaultHeaders(context, operationOrError, isInput);
+        if (operationOrError.isOperationShape()) {
+            OperationShape operation = operationOrError.asOperationShape().get();
+            if (isInput) {
+                AwsProtocolUtils.generateUnsignedPayloadSigV4Header(context, operation);
+            }
+        } else if (operationOrError.isStructureShape()) {
```
adamthom-amzn, line 95: Should we throw an exception in `else` as an assertion?

adamthom-amzn (approved):
Don't know how I overlooked this PR.

adamthom-amzn (approved):

PR #2222, opened 2021-04-08
Support generating non-AWS client

gosar:
### Description
These changes are to support generating a non-AWS typescript client. There are changes in the code generation logic to skip pieces that apply only to AWS services. Some typescript dependencies (aws-sdk/config-resolver) also needed updates.

This is initial prototype work for smithy-typescript-ssdk-demo.

### Testing
Using these changes I successfully generated a working non-AWS typescript client for the bootleg-service in smithy-typescript-ssdk-demo.

---

### DIFF
--- a/codegen/smithy-aws-typescript-codegen/src/main/java/software/amazon/smithy/aws/typescript/codegen/AddAwsAuthPlugin.java
+++ b/codegen/smithy-aws-typescript-codegen/src/main/java/software/amazon/smithy/aws/typescript/codegen/AddAwsAuthPlugin.java
@@ -44,6 +44,8 @@ import software.amazon.smithy.utils.SetUtils;
 /**
  * Configure clients with AWS auth configurations and plugin.
  */
+// TODO: Should this check for the presence of SigV4 trait instead of assuming the AWS service supports it?
+// TODO: Think about AWS Auth supported for only some operations and not all, when not AWS service
 public final class AddAwsAuthPlugin implements TypeScriptIntegration {
 
     @Override
@@ -54,7 +56,7 @@ public final class AddAwsAuthPlugin implements TypeScriptIntegration {
         TypeScriptWriter writer
     ) {
         ServiceShape service = settings.getService(model);
-        if (!areAllOptionalAuthOperations(model, service)) {
+        if (isAwsService(service) && !areAllOptionalAuthOperations(model, service)) {
             writer.addImport("Credentials", "__Credentials", TypeScriptDependency.AWS_SDK_TYPES.packageName);
             writer.writeDocs("Default credentials provider; Not available in browser runtime.")
                 .write("credentialDefaultProvider?: (input: any) => __Provider<__Credentials>;\n");
@@ -66,16 +68,17 @@ public final class AddAwsAuthPlugin implements TypeScriptIntegration {
         return ListUtils.of(
             RuntimeClientPlugin.builder()
                     .withConventions(AwsDependency.MIDDLEWARE_SIGNING.dependency, "AwsAuth", HAS_CONFIG)
-                    .servicePredicate((m, s) -> !areAllOptionalAuthOperations(m, s))
+                    .servicePredicate((m, s) -> isAwsService(s) && !areAllOptionalAuthOperations(m, s))
                     .build(),
             RuntimeClientPlugin.builder()
                     .withConventions(AwsDependency.MIDDLEWARE_SIGNING.dependency, "AwsAuth", HAS_MIDDLEWARE)
                     // See operationUsesAwsAuth() below for AwsAuth Middleware customizations.
                     .servicePredicate(
-                        (m, s) -> !testServiceId(s, "STS") && !hasOptionalAuthOperation(m, s)
+                        (m, s) -> isAwsService(s) && !testServiceId(s, "STS") && !hasOptionalAuthOperation(m, s)
                     ).build(),
             RuntimeClientPlugin.builder()
                     .withConventions(AwsDependency.MIDDLEWARE_SIGNING.dependency, "AwsAuth", HAS_MIDDLEWARE)
+                    // TODO: Does isAwsService need to be checked here again?
                     .operationPredicate(AddAwsAuthPlugin::operationUsesAwsAuth)
                     .build()
         );
@@ -89,7 +92,7 @@ public final class AddAwsAuthPlugin implements TypeScriptIntegration {
         LanguageTarget target
     ) {
         ServiceShape service = settings.getService(model);
-        if (areAllOptionalAuthOperations(model, service)) {
+        if (!isAwsService(service) || areAllOptionalAuthOperations(model, service)) {
             return Collections.emptyMap();
         }
         switch (target) {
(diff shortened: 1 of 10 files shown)


codegen/smithy-aws-typescript-codegen/src/main/java/software/amazon/smithy/aws/typescript/codegen/AddAwsAuthPlugin.java
```diff
@@ -44,6 +45,7 @@
 /**
  * Configure clients with AWS auth configurations and plugin.
  */
+// TODO: Think about AWS Auth supported for only some operations and not all, when not AWS service, with say @auth([])
```
JordonPhillips, line 48: If the service has an auth trait you need to add the mechanics needed to support that regardless of what the default is (where `@auth([])` is no auth by default). Since the `auth` trait can only reference auth traits on the service, you don't have to worry about a single operation introducing its own auth. I'm pretty sure the current setup already handles operations with disabled auth.

codegen/smithy-aws-typescript-codegen/src/main/java/software/amazon/smithy/aws/typescript/codegen/AwsTraitsUtils.java
```diff
@@ -0,0 +1,37 @@
+/*
```
JordonPhillips, line 1: There's already an `AwsProtocolUtils` where this would fit

codegen/smithy-aws-typescript-codegen/src/main/java/software/amazon/smithy/aws/typescript/codegen/AddAwsRuntimeConfig.java
```diff
@@ -82,10 +84,12 @@ public void addConfigInterfaceFields(
...
 
-        writer.writeDocs("Unique service identifier.\n@internal")
-                .write("serviceId?: string;\n");
-        writer.writeDocs("The AWS region to which this client will send requests")
-                .write("region?: string | __Provider<string>;\n");
+        if (isAwsService(settings, model)) {
+            writer.writeDocs("Unique service identifier.\n@internal")
+                    .write("serviceId?: string;\n");
+            writer.writeDocs("The AWS region to which this client will send requests")
+                    .write("region?: string | __Provider<string>;\n");
```
JordonPhillips, line 91: This is needed for sigv4

codegen/smithy-aws-typescript-codegen/src/main/java/software/amazon/smithy/aws/typescript/codegen/AddAwsRuntimeConfig.java
```diff
@@ -114,11 +118,14 @@ public void addConfigInterfaceFields(
...
             }
         }
-        runtimeConfigs.putAll(getDefaultConfig(target));
+        runtimeConfigs.putAll(getDefaultConfig(target, settings, model));
         return runtimeConfigs;
     }
 
-    private Map<String, Consumer<TypeScriptWriter>> getDefaultConfig(LanguageTarget target) {
+    private Map<String, Consumer<TypeScriptWriter>> getDefaultConfig(LanguageTarget target, TypeScriptSettings settings,
+                                                                     Model model) {
```
JordonPhillips, line 126: nit: when it overflows, one arg per line ideally like

```java
foo(
    boolean arg1,
    String arg2
)
```

codegen/smithy-aws-typescript-codegen/src/main/java/software/amazon/smithy/aws/typescript/codegen/AddAwsRuntimeConfig.java
```diff
@@ -128,40 +135,45 @@ public void addConfigInterfaceFields(
...
-                        }
-                );
+                if (isAwsService) {
+                    defaultConfigs.put("region", writer -> {
+                        writer.addDependency(TypeScriptDependency.INVALID_DEPENDENCY);
+                        writer.addImport("invalidProvider", "invalidProvider",
+                                TypeScriptDependency.INVALID_DEPENDENCY.packageName);
+                        writer.write("region: invalidProvider(\"Region is missing\"),");
+                    });
+                }
```
JordonPhillips, line 145: region is needed for sigv4

codegen/smithy-aws-typescript-codegen/src/main/java/software/amazon/smithy/aws/typescript/codegen/AddAwsRuntimeConfig.java
```diff
@@ -128,40 +135,45 @@ public void addConfigInterfaceFields(
...
-                            writer.addImport("loadConfig", "loadNodeConfig",
-                                    AwsDependency.NODE_CONFIG_PROVIDER.packageName);
-                            writer.addDependency(TypeScriptDependency.CONFIG_RESOLVER);
-                            writer.addImport("NODE_REGION_CONFIG_OPTIONS", "NODE_REGION_CONFIG_OPTIONS",
-                                    TypeScriptDependency.CONFIG_RESOLVER.packageName);
-                            writer.addImport("NODE_REGION_CONFIG_FILE_OPTIONS", "NODE_REGION_CONFIG_FILE_OPTIONS",
-                                    TypeScriptDependency.CONFIG_RESOLVER.packageName);
-                            writer.write(
+                if (isAwsService) {
+                    defaultConfigs.put("region", writer -> {
```
JordonPhillips, line 155: sigv4, region, all that

codegen/smithy-aws-typescript-codegen/src/main/java/software/amazon/smithy/aws/typescript/codegen/AddUserAgentDependency.java
```diff
@@ -33,12 +35,15 @@
 /**
  * Add client plubins and configs to support injecting user agent.
  */
+// TODO: Looks to add this back for non-AWS service clients, by fixing the dependency on ClientSharedValues.serviceId
```
JordonPhillips, line 38: should be fairly easy to just write out the service shape name / id. This is what Go does

JordonPhillips (changes requested):
We'll still need to support configuring the region even if it isn't being pulled from the aws config file if the service is using sigv4. Their region may be static or may not map to an AWS region, but something needs to be there for the request to be signed.

codegen/smithy-aws-typescript-codegen/src/main/java/software/amazon/smithy/aws/typescript/codegen/AwsTraitsUtils.java
gosar: I think `AwsProtocolUtils` is for protocol specific utilities and this isn't about protocols, so I kept it separate.

codegen/smithy-aws-typescript-codegen/src/main/java/software/amazon/smithy/aws/typescript/codegen/AddAwsAuthPlugin.java
gosar: I put this TODO to think more deeply about different cases when I handle supporting SigV4 for non-AWS clients.

gosar:
@JordonPhillips, thanks for the review. I addressed the code formatting comment. Re: sigv4, region, etc. yes I'm aware and those are next tasks for me. This version was supposed to be without sigV4 support. I also have a task to make UserAgent work again. Baby steps...

JordonPhillips (approved):

PR #2239, opened 2021-04-13
fix(client-lambda): added nodejs 14 runtime

rehrumesh:
### Description
Added nodejs14.x to Runtime enum.

### Testing
NA

### Additional context
NA

---


trivikr:
The nodejs14 runtime would be added to client-lambda once the models are updated at https://github.com/aws/aws-sdk-js-v3/blob/d6a67d7cd471e3440fa7b23ba00b11c00b10a566/codegen/sdk-codegen/aws-models/lambda.2015-03-31.json#L6409-L6412

We're working with internal teams to fix issues with smithy models.

Issue #2231, opened 2021-04-10
docs(clients) async/await example does not use async await


pratikpc:
## Describe the issue with documentation
async/await example in client-ses does not sue async await

## Steps to reproduce
Present in [README.md at example here](https://github.com/aws/aws-sdk-js-v3/blob/main/clients/client-ses/README.md#v2-compatible-style)

trivikr:
Thank you @pratikpc for finding this issue, and adding a PR to fix it.

This fix is required in codegen for all clients. I cherry-picked commit from your PR, and posted fix for all clients in https://github.com/aws/aws-sdk-js-v3/pull/2243

pratikpc:
@trivikr thanks

I renamed the issue to now make it more clear that multiple projects are affected

PR #2243, opened 2021-04-13
docs(clients): add missing await for client.operationName call

trivikr:
### Issue
Fixes: https://github.com/aws/aws-sdk-js-v3/issues/2231

### Description
Adds missing await for client.operationName call in README

### Testing
Verified that missing await is added in README

---

### DIFF
--- a/clients/client-accessanalyzer/README.md
+++ b/clients/client-accessanalyzer/README.md
@@ -137,7 +137,7 @@ const client = new AWS.AccessAnalyzer({ region: "REGION" });
 
 // async/await.
 try {
-  const data = client.createAnalyzer(params);
+  const data = await client.createAnalyzer(params);
   // process data.
 } catch (error) {
   // error handling.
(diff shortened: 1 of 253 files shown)


pratikpc:
@trivikr it would be nice if we could have better examples too

For example, for AWS SES, I had to go to https://docs.aws.amazon.com/sdk-for-javascript/v3/developer-guide/ses-examples-sending-email.html to find out how to send EMails which I think should be the focus of AWS SES docs

trivikr:
> it would be nice if we could have better examples too
> 
> For example, for AWS SES, I had to go to [docs.aws.amazon.com/sdk-for-javascript/v3/developer-guide/ses-examples-sending-email.html](https://docs.aws.amazon.com/sdk-for-javascript/v3/developer-guide/ses-examples-sending-email.html) to find out how to send EMails which I think should be the focus of AWS SES docs

The current README.md has auto generated examples.
We're working on improved API Reference story, where specific examples related to service would be added.

AllanZhengYP (approved):

PR #2287, opened 2021-04-21
chore(codegen,config-resolver): refactor how endpoint is resolved for non-AWS client

gosar:
### Description
This is a refactoring of my earlier attempt at getting a non-AWS client working wrt endpoint resolution. This addresses the concern in https://github.com/aws/aws-sdk-js-v3/pull/2279#discussion_r617402875 by separating out the logic used by AWS client v/s non-AWS client. Also makes endpoint required for non-AWS client, addressing https://github.com/adamthom-amzn/aws-sdk-js-v3/pull/1#issuecomment-804665638.

This approach adds some duplication. But seems to fit within the current plugin model in the codegen. Trying to use the same EndpointsConfig.ts but different methods/types would not fit easily into the ServiceGenerator logic. I thought it is better to have some minimal duplication in config-resolver than bigger restructing in codegen.

### Testing
I generated the client used in our smithy-typescript-ssdk-demo package and it worked with these changes.

### Questions:
* Let me know if `chore(codegen,config-resolver): refactor how endpoint is resolved for non-AWS client` matches the expectation of git commit titles for when this is squashed and merged. Or is `chore: refactor how endpoint is resolved for non-AWS client` better?
* EndpointsConfig and AwsEndpointsConfig may be better than NonAwsEndpointsConfig and EndpointsConfig. Is it fine to rename these? Or other suggestions on names. (edit): Using EndpointsConfig (for AWS) and CustomEndpointsConfig (for non-AWS) now. Let me know if other suggestions.

---

### DIFF
--- a/codegen/smithy-aws-typescript-codegen/src/main/java/software/amazon/smithy/aws/typescript/codegen/AddBuiltinPlugins.java
+++ b/codegen/smithy-aws-typescript-codegen/src/main/java/software/amazon/smithy/aws/typescript/codegen/AddBuiltinPlugins.java
@@ -47,8 +47,14 @@ public class AddBuiltinPlugins implements TypeScriptIntegration {
                         .withConventions(TypeScriptDependency.CONFIG_RESOLVER.dependency, "Region", HAS_CONFIG)
                         .servicePredicate((m, s) -> isAwsService(s))
                         .build(),
+                // Only one of Endpoints or CustomEndpoints should be used
                 RuntimeClientPlugin.builder()
                         .withConventions(TypeScriptDependency.CONFIG_RESOLVER.dependency, "Endpoints", HAS_CONFIG)
+                        .servicePredicate((m, s) -> isAwsService(s))
+                        .build(),
+                RuntimeClientPlugin.builder()
+                        .withConventions(TypeScriptDependency.CONFIG_RESOLVER.dependency, "CustomEndpoints", HAS_CONFIG)
+                        .servicePredicate((m, s) -> !isAwsService(s))
                         .build(),
                 RuntimeClientPlugin.builder()
                         .withConventions(TypeScriptDependency.MIDDLEWARE_RETRY.dependency, "Retry")
--- a/packages/config-resolver/src/CustomEndpointsConfig.ts
+++ b/packages/config-resolver/src/CustomEndpointsConfig.ts
@@ -1,0 +1,45 @@
+// TODO: Create a .spec.ts for this
+import { Endpoint, Provider, UrlParser } from "@aws-sdk/types";
+
+export interface CustomEndpointsInputConfig {
+  /**
+   * The fully qualified endpoint of the webservice.
+   */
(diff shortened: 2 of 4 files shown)


codegen/smithy-aws-typescript-codegen/src/main/java/software/amazon/smithy/aws/typescript/codegen/AddBuiltinPlugins.java
```diff
@@ -47,8 +47,14 @@
                         .withConventions(TypeScriptDependency.CONFIG_RESOLVER.dependency, "Region", HAS_CONFIG)
                         .servicePredicate((m, s) -> isAwsService(s))
                         .build(),
+                // Only one of Endpoints of NonAwsEndpoints should be used
```
adamthom-amzn, line 50: nit: second "of" should be "or"

packages/config-resolver/src/NonAwsEndpointsConfig.ts
```diff
@@ -0,0 +1,45 @@
...
+  tls?: boolean;
+}
+
+interface PreviouslyResolved {
+  urlParser: UrlParser;
+}
+
+export interface NonAwsEndpointsResolvedConfig extends Required<NonAwsEndpointsInputConfig> {
+  endpoint: Provider<Endpoint>;
+  isCustomEndpoint: boolean; // TODO: Can this be removed or some other logic depends on this?
```
adamthom-amzn, line 22: is this always true? you could type it as `true` if so.

packages/config-resolver/src/NonAwsEndpointsConfig.ts
```diff
@@ -0,0 +1,45 @@
+// TODO: Create a .spec.ts for this
+import { Endpoint, Provider, UrlParser } from "@aws-sdk/types";
+
+export interface NonAwsEndpointsInputConfig {
```
JordonPhillips, line 4: Can we call this `CustomEndpoint` or `StaticEndpoint` rather than `NonAwsEndpoint`?

JordonPhillips (changes requested):

packages/config-resolver/src/NonAwsEndpointsConfig.ts
gosar: Yeah I asked this in my description:
> EndpointsConfig and AwsEndpointsConfig may be better than NonAwsEndpointsConfig and EndpointsConfig. Is it fine to rename these? Or other suggestions on names.

Can we do EndpointsConfig and AwsEndpointsConfig? If not, a name different than NonAwsEndpointsConfig definitely needed. I like CustomEndpoint; it has a `isCustomEndpoint:true` too.

JordonPhillips: It would probably  be a breaking change to rename the existing one is the problem

JordonPhillips (approved):

Issue #1525, opened 2020-09-14
`NodeHttp2Handler` does not handle session failures


isker:
**Describe the bug**
`NodeHttp2Handler` pools its `Http2Session`s per authority.  An `Http2Session` can be closed for a variety of reasons, including those outside of the control of the client.  (See [the docs for the `goaway` event](https://nodejs.org/api/http2.html#http2_event_goaway) which can be triggered by the server at any time.)

Whenever a session is closed, it is invalid for further use, but `NodeHttp2Handler` only invalidates its cache for sessions it has closed itself.  This is insufficient - if the server triggers node to close the session for any reason, the `NodeHttp2Handler` is invalid for that authority forever.

**SDK version number**
`@aws-sdk/node-http-handler@1.0.0-gamma.6`

**Is the issue in the browser/Node.js/ReactNative?**
node

**Details of the browser/Node.js/ReactNative version**
```
λ node -v
v13.9.0
```

**To Reproduce (observed behavior)**
The repro steps and context for this issue are the same as in https://github.com/aws/aws-sdk-js-v3/issues/1524.

**Expected behavior**
`NodeHttp2Handler` should invalidate its on the `close` event for any `Http2Session`.  Compare to another popular client:
https://github.com/hisco/http2-client/blob/12a9e6fa6701a46e92e9b7adf395ce646b2a26b4/lib/request.js#L299-L303

TysonAndre:
I believe that a patch to fix this would look something along these lines? Note that I haven't tested or even run this - anything involving http/2 would probably require multiple unit tests - node's http/2 implementation has some edge cases in older node versions that can lead to crashes (possibly related to closing a session while it's in the process of closing)


```patch
--- a/packages/node-http-handler/src/node-http2-handler.ts
+++ b/packages/node-http-handler/src/node-http2-handler.ts
@@ -107,14 +107,31 @@ export class NodeHttp2Handler implements HttpHandler {
 
     const newSession = connect(authority);
     connectionPool.set(authority, newSession);
+    const destroySessionCb = () => {
+      this.destroySession(authority, newSession);
+    };
+    newSession.on('close' , destroySessionCb); // probably redundant?
+    newSession.on('goaway' , destroySessionCb);
+    newSession.on('error' , destroySessionCb);
+    newSession.on('frameError' , destroySessionCb);
 
     const sessionTimeout = this.sessionTimeout;
     if (sessionTimeout) {
[...]

PR #2289, opened 2021-04-22
fix(node-http-handler): handle NodeHttp2Handler session failure

TysonAndre:
### Issue
#1525 

### Description
detect errors on the NodeHttp2Handler, immediately destroy
connections on unexpected error mode, and reconnect.

Prior to this PR, if the server sent the client a GOAWAY frame,
the session would not be removed from the connection pool and requests
would fail indefinitely. This fixes that.

This tries to avoid keeping streams(requests) on the Http2Session
(tcp connection) from being stuck in an open state waiting for a gentle
close even if there were unexpected protocol or connection errors
during the request, assuming http2 errors are rare.
(if a server or load balancer or network is misbehaving,
close() might get stuck waiting for requests to finish,
especially if requests and sessions don't have timeouts?)

I'm only slightly familiar with http/2 client implementations from
working on clients for Apple Push Notification Service.

- In those, the client could rely on a session and request timeout
  existing, so close() would finish. In aws-sdk-js-v3, timeouts are optional.
- I've seen some strange race conditions prior to node 12 in different
  client implementations for (certain combinations of) close and/or destroying

### Testing
Unit tests were added to confirm that disconnections initiated by the server
resulted in the connection being removed from the connection pool.

More thorough manual tests would be useful to test all possible networking edge cases,
e.g. completely losing a connection to aws load balancers, a connection closing without a goaway frame, etc,
testing this with a high volume of traffic to see if it recovers properly, etc.,
but I'm not familiar with how to simulate that accurately in node

### Additional context

---

### DIFF
--- a/packages/node-http-handler/src/node-http2-handler.spec.ts
+++ b/packages/node-http-handler/src/node-http2-handler.spec.ts
@@ -121,55 +121,51 @@ describe("NodeHttp2Handler", () => {
       mockH2Server3.on("connection", () => {
         establishedConnections += 1;
       });
-      try {
-        // should throw
-        const req = new HttpRequest({ ...getMockReqOptions(), port: port3 });
-        expect(establishedConnections).toBe(0);
-        expect(numRequests).toBe(0);
-        await assert.rejects(
-          nodeH2Handler.handle(req, {}),
-          UNEXPECTEDLY_CLOSED_REGEX,
-          "should be rejected promptly due to goaway frame"
-        );
-        expect(establishedConnections).toBe(1);
-        expect(numRequests).toBe(1);
-        await assert.rejects(
-          nodeH2Handler.handle(req, {}),
-          UNEXPECTEDLY_CLOSED_REGEX,
-          "should be rejected promptly due to goaway frame"
-        );
-        expect(establishedConnections).toBe(2);
-        expect(numRequests).toBe(2);
-        await assert.rejects(
-          nodeH2Handler.handle(req, {}),
-          UNEXPECTEDLY_CLOSED_REGEX,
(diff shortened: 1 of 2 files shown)


trivikr (approved):
LGTM

packages/node-http-handler/src/node-http2-handler.spec.ts
```diff
@@ -121,55 +121,51 @@ describe("NodeHttp2Handler", () => {
...
+        const buffers = [];
+        resultReader.on("data", (chunk) => buffers.push(chunk));
+        resultReader.on("end", () => {
+          resolve(Buffer.concat(buffers).toString("utf8"));
         });
-        expect(responseBody).toBe("test");
-        expect(establishedConnections).toBe(4);
-        expect(numRequests).toBe(5);
-      } finally {
-        mockH2Server3.close();
```
TysonAndre: The intent of this check was to avoid making other tests fail if this test failed (if they ever reused the same port in the future) - it could waste time debugging if `"handles connections destroyed by servers"` below failed because it's also using port3 (or something even less straightforward).

Now that I think of it, checking for the mockH2Server in an after() handler instead and calling mockH2Server.close() may have been cleaner than this try{}finally{} approach. Thoughts?

trivikr: `mockH2Server.close()` is called in `afterAll` as mockH2Server is reused between tests.

If any additional servers are created by tests, they need to be closed within the test.
For example, `mockH2Server2` while testing new authority and `mockH2Server3` for goaway frames or connections destroyed by servers.

PR #2364, opened 2021-05-08
fix(clients): properly decode SQS messages

### LINKED ISSUE #2362: ZA̡͊͠͝LGΌ causes "Invalid MD5 checksum on messages"
The unholy child weeps the blood of virgins, the tainted souls are summoned into the realm of the living, dooming humanity to an eternity of dread torture, transporting the programmer's consciousness into a world of ceaseless screaming. You have summoned Zalgo.

```ts
const decodeEscapedXML = (str: string) =>
  str
    .replace(/&amp;/g, "&")
    .replace(/&apos;/g, "'")
    .replace(/&quot;/g, '"')
    .replace(/&gt;/g, ">")
    .replace(/&lt;/g, "<");
```

https://github.com/aws/aws-sdk-js-v3/blob/v3.14.0/clients/client-sqs/protocols/Aws_query.ts#L3133-L3139

Maybe try using an XML parser instead?

---

There are multiple obvious problems with this:

1. `&#xD;` decodes to `&#xD;` (wrong) not carriage return (correct)
2. `&amp;lt;` decodes to `<` (wrong) not `&lt;` (correct)

Thanks to (1), all my HL7 messages in my SQS FIFO queue won't process because carriage returns broke the client. I haven't run into (2), but I'm sure I will eventually.

Such is the life of the foolish souls who try parsing XML with 5 lines of regex.

(Silver lining is thanks to the MD5 checksum, this bug only causes unavailability instead of corruption.)

pauldraper:
Fixes #2362.

---

### DIFF
--- a/clients/client-sqs/package.json
+++ b/clients/client-sqs/package.json
@@ -58,10 +58,12 @@
     "@aws-sdk/util-utf8-browser": "3.13.1",
     "@aws-sdk/util-utf8-node": "3.13.1",
     "fast-xml-parser": "3.19.0",
+    "he": "^1.0.0",
     "tslib": "^2.0.0"
   },
   "devDependencies": {
     "@aws-sdk/client-documentation-generator": "3.14.0",
+    "@types/he": "^1.0.0",
     "@types/node": "^12.7.5",
     "downlevel-dts": "0.7.0",
     "jest": "^26.1.0",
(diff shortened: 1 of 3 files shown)


trivikr:
Superceded by https://github.com/aws/aws-sdk-js-v3/pull/2381

pauldraper:
Thanks for the real fix :)

PR #2010, opened 2021-02-09
feat(util-dynamodb): marshall JavaScript Maps

trivikr:
### Issue
Fixes: https://github.com/aws/aws-sdk-js-v3/issues/2008

### Description
adds support to marshall [JavaScript Map](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map) into [DynamoDB Attribute Map](https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_AttributeValue.html#DDB-Type-AttributeValue-M) 

### Testing
Unit tests

---
By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #2489, opened 2021-06-14
fix(util-dynamodb): unmarshall JavaScript Maps

rensvanleeuwen:
### Issue
No issue created for it

### Description
In the DynamoDB document client, by means of https://github.com/aws/aws-sdk-js-v3/pull/2010, the marshalling of JavaScript Maps was implemented.

However, the opposite direction does not return a `Map` and therefore, from the API perspective, it is cumbersome to work with, as it is no symmetrical. 

DISCLAIMER: I am relatively new to TypeScript, so some sloppiness ahead of you.

Example (semi-pseudo) code:

```
const dynamoDBClient = new DynamoDBClient({ region: appConfig.awsRegion });
const documentClient = DynamoDBDocumentClient.from(dynamoDBClient, {
  marshallOptions: {
    removeUndefinedValues: true,
  },
});

interface Person {
  pk: string;
  properties: Map<string, any>;
}

const person: Person = {
  pk: '1',
  properties: new Map<string, any>(
    [
      ['age': 42],
      ['name': 'Arthur Dent'],
    ]),
};

console.log(person.properties); // Map(2) { 'age' => 41, 'name' => 'Arthur Dent' }

const command = new PutCommand({
  TableName: 'persons',
  Item: person,
});

// This leads to a nested property that has the type 'M'. So far, so good.
await documentClient.send(command);

// This seems to be the point where the type is different after we have read
// it from DynamoDB again. 
const retrievalCommand = {} // Some command that retrieves the person
const retrievedPerson = await documentClient.send(command);

console.log(retrievedPerson.properties); // { age: 41, name: 'Arthur Dent' }
```

### Testing
I have amended the unit tests as they were to ensure that they return a `Map` when needed. I also added a test case where a `Map` is nested inside another `Map`, which appeared to be missing.

### Additional context

---

### DIFF
--- a/packages/util-dynamodb/src/convertToNative.spec.ts
+++ b/packages/util-dynamodb/src/convertToNative.spec.ts
@@ -128,11 +128,24 @@ describe("convertToNative", () => {
         {
           input: [
             { M: { nullKey: { NULL: true }, boolKey: { BOOL: false } } },
-            { M: { stringKey: { S: "one" }, numberKey: { N: "1.01" }, bigintKey: { N: "9007199254740996" } } },
+            {
+              M: {
+                stringKey: { S: "one" },
+                numberKey: { N: "1.01" },
+                bigintKey: { N: "9007199254740996" },
+              },
+            },
           ],
           output: [
-            { nullKey: null, boolKey: false },
-            { stringKey: "one", numberKey: 1.01, bigintKey: BigInt(9007199254740996) },
+            new Map<string, any>([
+              ["nullKey", null],
+              ["boolKey", false],
+            ]),
+            new Map<string, any>([
+              ["stringKey", "one"],
+              ["numberKey", 1.01],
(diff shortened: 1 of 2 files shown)


rensvanleeuwen:
I will close this one.

I was too focused on 'just' getting the `Map` back, and I completely missed the fact that there are other branches in the marshaller that write their output to map-like output (with a `M` data type descriptor) on the side of DynamoDB. Most prominently the `convertToMapAttrFromEnumerableProps` function that is invoked for regular `Object`s.

Multiple sources of input ending up in the same DynamoDB output means that it can not be converted back to its original input, unless the (un)marshalling would write additional serialisation hints to the data it writes and reads. That doesn't sound feasible, as users will have different clients, etc ..

I guess that it is up the user of the API to convert the `Object` that is returned from the unmarshalling and turn that back to the shape that the client expects it to be, which is a bit inconvenient.

PR #2499, opened 2021-06-17
chore(codegen): pin smithy to 1.8.x

trivikr:
### Issue
Follow-up to https://github.com/aws/aws-sdk-js-v3/pull/2153

### Description
Pins smithy to minor version 1.8.x

### Testing
CI

---
By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #2500, opened 2021-06-18
fix(smithy-client): use uppercase for percent encodings

trivikr:
### Issue
Refs: https://github.com/aws/aws-sdk-js-v3/pull/2499#issuecomment-864117029

### Description
Use uppercase for percent encodings

As per [rfc3986 section 2.1](https://datatracker.ietf.org/doc/html/rfc3986#section-2.1), it should be `"%2A"`

> The uppercase hexadecimal digits 'A' through 'F' are equivalent to
   the lowercase digits 'a' through 'f', respectively.  If two URIs
   differ only in the case of hexadecimal digits used in percent-encoded
   octets, they are equivalent.  For consistency, URI producers and
   normalizers should use uppercase hexadecimal digits for all percent-
   encodings.

### Testing
CI

---


alexforsyth (approved):
Nice catch

PR #2514, opened 2021-06-23
chore(clients): strictly parse booleans

JordonPhillips:
### Issue

n/a

### Description

This updates boolean parsing to be more strict. Previously there was a simple `=== "true"` check, which allowed malformed requests to be accepted.

### Testing

Tests were added, including tests covering common alternatives for boolean string representations.

### Additional context

This is part of a pass through to make sure the types we're receiving are actually the types we expect. There will likely be more PRs on this theme.

See https://github.com/awslabs/smithy-typescript/pull/364 for the other half of this.

---


adamthom-amzn (approved):

packages/smithy-client/src/parse-utils.ts
```diff
@@ -0,0 +1,16 @@
+/**
+ * Give an input string, stritctly parses a boolean value.
```
trivikr, line 2: nit: typo
```suggestion
 * Give an input string, strictly parses a boolean value.
```

packages/smithy-client/src/parse-utils.ts
```diff
@@ -0,0 +1,16 @@
...
+ *
+ * @param value The boolean string to parse.
+ * @returns true for "true", false for "false", otherwise an error is thrown.
+ */
+export function parseBoolean(value: "true" | "false"): boolean {
```
trivikr, line 7: Prefer arrow function
```suggestion
export const parseBoolean = (value: "true" | "false"): boolean => {
```

trivikr:
CI Error:

```console
lerna ERR! yarn run build exited 2 in '@aws-sdk/aws-restjson'
--
429 | lerna ERR! yarn run build stdout:
430 | $ yarn build:cjs && yarn build:es
431 | $ tsc -p tsconfig.json
432 | protocols/Aws_restJson1.ts(2561,46): error TS2345: Argument of type 'string' is not assignable to parameter of type '"true" \| "false"'.
433 | protocols/Aws_restJson1.ts(2564,47): error TS2345: Argument of type 'string' is not assignable to parameter of type '"true" \| "false"'.
434 | protocols/Aws_restJson1.ts(2580,39): error TS2345: Argument of type 'string' is not assignable to parameter of type '"true" \| "false"'.
435 | info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
436 | info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.


```

packages/smithy-client/src/parse-utils.ts
```diff
@@ -0,0 +1,16 @@
+/**
+ * Give an input string, strictly parses a boolean value.
+ *
+ * @param value The boolean string to parse.
+ * @returns true for "true", false for "false", otherwise an error is thrown.
+ */
+export const parseBoolean = (value: "true" | "false"): boolean => {
```
trivikr, line 7: Allow passing string in parseBoolean, as it throws error when it's value is neither true/false.
```suggestion
export const parseBoolean = (value: string): boolean => {
```

trivikr (approved):

packages/smithy-client/src/parse-utils.ts
```diff
@@ -0,0 +1,16 @@
...
+ * @returns true for "true", false for "false", otherwise an error is thrown.
+ */
+export const parseBoolean = (value: string): boolean => {
+  switch (value) {
+    case "true":
+      return true;
+    case "false":
+      return false;
+    default:
+      throw new Error(`Unable to parse boolean value "${value}"`);
```
adamthom-amzn, line 14: should this be TypeError like the `expect` methods, or no?

adamthom-amzn (approved):

packages/smithy-client/src/parse-utils.ts
JordonPhillips: Technically it's not a type error, it's a parse error. We can always change it if necessary

adamthom-amzn: As long as at the end of the day we can generate a meaningful SerializationException for callers (when SerializationException has messages... whoops), that's all I care about. Reading messages from bare `Error` types is dangerous, could leak something unintended.

PR #2515, opened 2021-06-23
chore: assert deserialized types are correct

JordonPhillips:
### Issue

n/a

### Description

This adds expectations for the default member deserialization so that an error will be thrown if the type being deserialized isn't correct.

### Testing

Tests were added.

### Additional context

See https://github.com/awslabs/smithy-typescript/pull/365 for the other half of this

---

### DIFF
--- a/clients/client-accessanalyzer/protocols/Aws_restJson1.ts
+++ b/clients/client-accessanalyzer/protocols/Aws_restJson1.ts
@@ -112,6 +112,9 @@ import {
 import { HttpRequest as __HttpRequest, HttpResponse as __HttpResponse } from "@aws-sdk/protocol-http";
 import {
   SmithyException as __SmithyException,
+  expectBoolean as __expectBoolean,
+  expectNumber as __expectNumber,
+  expectString as __expectString,
   extendedEncodeURIComponent as __extendedEncodeURIComponent,
 } from "@aws-sdk/smithy-client";
 import {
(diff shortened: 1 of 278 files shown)


packages/smithy-client/src/parse-utils.ts
```diff
@@ -0,0 +1,38 @@
+/**
+ * Asserts a value is a boolean and returns it.
+ *
+ * @param value A value that is expected to be a boolean.
+ * @returns The value if it is a boolean, otherwise an error is thrown.
+ */
+export function expectBoolean(value: any): boolean {
```
adamthom-amzn, line 7: it seems like a lot of the deserialization code would be cleaner if these expectation functions returned `undefined` if `value` was `undefined` or `null` -- this could reduce the amount of code overall instead of adding to it.

JordonPhillips: You are correct. Went from adding about 20k lines to removing about 20k lines

clients/client-accessanalyzer/protocols/Aws_restJson1.ts
```diff
@@ -1165,7 +1168,7 @@ export const deserializeAws_restJson1CreateAccessPreviewCommand = async (
   };
   const data: any = await parseBody(output.body, context);
   if (data.id !== undefined && data.id !== null) {
```
adamthom-amzn, line 1170: Don't know how the generator is structured, but we should be able to remove this wrapping `if` function that appears a lot as well. (Though I wouldn't block on it)

adamthom-amzn (approved):

clients/client-accessanalyzer/protocols/Aws_restJson1.ts
JordonPhillips: This is probably due to differences with handling of top-level members.

adamthom-amzn (approved):

trivikr (approved):

PR #2551, opened 2021-07-07
SSDK Protocol tests

adamthom-amzn:
### Description
This changes `generate-clients` to also generate and copy over SSDK protocol tests when they are enabled. Only REST-JSON is enabled for now.

### Testing
I generated these tests based on Smithy 1.9.1 with [this PR](https://github.com/awslabs/smithy/pull/857) as well as [this PR to smithy-typescript](https://github.com/awslabs/smithy-typescript/pull/377). With those two changes, `yarn generate-clients` followed by `yarn test:protocols` ran all of the protocol tests, and they all passed, including the new SSDK protocol tests.

The aforementioned PRs also result in slight changes to the client protocol tests, but I did not include them in this, as they are not required for the protocol tests to pass and are not merged into their respective repositories.

---


adamthom-amzn:
I'm going to close this while I debug an alternate CI method.

Issue #2550, opened 2021-07-07
Add an option to disable session cache in NodeHttp2Handler


trivikr:
### Is your feature request related to a problem? Please describe.

The services which support Http2 streaming might limit max concurrent connections on the server side. This leads to an error being thrown if concurrent requests more than `maxConcurrentConnections` are created on the client side which share an HTTP2 session.

The issue with AWS Transcribe Streaming is explained in repo https://github.com/jjtindale/aws-parallel-transcribe-repro
The error thrown in Node.js with concurrent requests more than `maxConcurrentConnections` is explained in https://github.com/trivikr/node-http2-parallel-requests-on-session

### Describe the solution you'd like
Add an option to disable session cache in NodeHttp2Handler. This would ensure that a new session is created for every request.

For example:
```js
const client = new TranscribeStreamingClient({
  requestHandler: new NodeHttp2Handler({
    disableSessionCache: true
  }),
});
```

### Describe alternatives you've considered
Write my own Http2Handler which does not share ClientHttp2Sessions between requests.

PR #2553, opened 2021-07-08
feat(node-http-handler): configure disableConcurrentStreams in NodeHttp2Handler

trivikr:
### Issue
Fixes: https://github.com/aws/aws-sdk-js-v3/issues/2550

### Description
Adds an option to disableConcurrentStreams in NodeHttp2Handler

Example:
```js
const client = new H2HandlerClient({
  requestHandler: new NodeHttp2Handler({
    disableConcurrentStreams: true
  }),
});
```

### Testing
CI is successful
Verified that TranscribeStreaming works in parallel in demo code https://github.com/trivikr/aws-parallel-transcribe-repro/pull/1

---

### DIFF
--- a/packages/node-http-handler/src/node-http2-handler.spec.ts
+++ b/packages/node-http-handler/src/node-http2-handler.spec.ts
@@ -305,25 +305,41 @@ describe(NodeHttp2Handler.name, () => {
   });
 
   describe("sessionTimeout", () => {
-    const sessionTimeout = 500;
+    const sessionTimeout = 200;
 
-    beforeEach(() => {
-      nodeH2Handler = new NodeHttp2Handler({ sessionTimeout });
-    });
+    describe("destroys session on sessionTimeout", () => {
+      it("disableSessionCache: false (default)", async (done) => {
+        nodeH2Handler = new NodeHttp2Handler({ sessionTimeout });
+        await nodeH2Handler.handle(new HttpRequest(getMockReqOptions()), {});
 
-    it("closes and removes session on sessionTimeout", async (done) => {
-      await nodeH2Handler.handle(new HttpRequest(getMockReqOptions()), {});
-
-      const authority = `${protocol}//${hostname}:${port}`;
-      // @ts-ignore: access private property
-      const session: ClientHttp2Session = nodeH2Handler.connectionPool.get(authority);
-      expect(session.closed).toBe(false);
-      setTimeout(() => {
-        expect(session.closed).toBe(true);
+        const authority = `${protocol}//${hostname}:${port}`;
         // @ts-ignore: access private property
-        expect(nodeH2Handler.connectionPool.get(authority)).not.toBeDefined();
-        done();
-      }, sessionTimeout + 100);
+        const session: ClientHttp2Session = nodeH2Handler.connections[0];
+        expect(session.closed).toBe(false);
+        // @ts-ignore: access private property
+        expect(nodeH2Handler.connectionPool.get(authority)).toBeDefined();
+        setTimeout(() => {
+          expect(session.closed).toBe(true);
+          expect(session.destroyed).toBe(false);
+          // @ts-ignore: access private property
+          expect(nodeH2Handler.connectionPool.get(authority)).not.toBeDefined();
+          done();
+        }, sessionTimeout + 100);
+      });
+
+      it("disableSessionCache: true", async (done) => {
+        nodeH2Handler = new NodeHttp2Handler({ sessionTimeout, disableSessionCache: true });
+        await nodeH2Handler.handle(new HttpRequest(getMockReqOptions()), {});
+
+        // @ts-ignore: access private property
+        const session: ClientHttp2Session = nodeH2Handler.connections[0];
+        expect(session.closed).toBe(false);
+        setTimeout(() => {
+          expect(session.closed).toBe(true);
+          expect(session.destroyed).toBe(false);
+          done();
+        }, sessionTimeout + 100);
+      });
     });
   });
 
--- a/packages/node-http-handler/src/node-http2-handler.spec.ts
+++ b/packages/node-http-handler/src/node-http2-handler.spec.ts
@@ -290,7 +290,7 @@ describe(NodeHttp2Handler.name, () => {
     const requestTimeout = 200;
 
     describe("does not throw error when request not timed out", () => {
-      it("disableSessionCache: false (default)", async () => {
+      it("disableConcurrentStreams: false (default)", async () => {
         mockH2Server.removeAllListeners("request");
         mockH2Server.on("request", createResponseFunctionWithDelay(mockResponse, requestTimeout - 100));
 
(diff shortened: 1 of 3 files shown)


packages/node-http-handler/src/node-http2-handler.ts
```diff
@@ -118,15 +141,45 @@ export class NodeHttp2Handler implements HttpHandler {
...
+      this.destroySession(newSession);
+    };
+    newSession.on("goaway", destroySessionCb);
+    newSession.on("error", destroySessionCb);
+    newSession.on("frameError", destroySessionCb);
+
+    const sessionTimeout = this.sessionTimeout;
+    if (sessionTimeout) {
+      newSession.setTimeout(sessionTimeout, () => {
+        newSession.close();
```
AllanZhengYP, line 161: In case of connection timeout, we should also remove it from the session cache.

packages/node-http-handler/src/node-http2-handler.ts
```diff
@@ -22,50 +22,68 @@ export interface NodeHttp2HandlerOptions {
...
 }
 
 export class NodeHttp2Handler implements HttpHandler {
   private readonly requestTimeout?: number;
   private readonly sessionTimeout?: number;
-  private readonly connectionPool: Map<string, ClientHttp2Session>;
+  private readonly disableSessionCache?: boolean;
+
   public readonly metadata = { handlerProtocol: "h2" };
+  private sessionList: ClientHttp2Session[];
```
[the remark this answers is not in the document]
trivikr: The `sessionList` stores all sessions, including the ones not present in `sessionCache`
When `disableSessionCache` is set to true, the sessions should not be stored in `sessionCache`.

packages/node-http-handler/src/node-http2-handler.ts
trivikr: It's removed on Line 182 in `getSessionFromCache`

trivikr: We should remove it from `sessionList` here.

packages/node-http-handler/src/node-http2-handler.ts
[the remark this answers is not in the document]
trivikr: When `disableSessionCache: true`
* we're closing the session on successful response.
* we're destroying the session on reject.

So, `sessionList` is actually not required. I'll remove it.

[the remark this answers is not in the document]
trivikr: > So, `sessionList` is actually not required. I'll remove it.

The `sessionList` is required to store sessions, in case `destroy()` is called on handler while sessions are active.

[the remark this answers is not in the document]
trivikr: Decisions from off-GitHub discussions:
* rename config `disableSessionCache` to `disableConcurrentStreams`.
* convert sessionCache from `Map<string, ClientHttp2Session>()` to `Map<string, ClientHttp2Session[]>()`:
  * when disableConcurrentStreams: false, the array will have only one value.
  * when disableConcurrentStreams: true, the array will have new session per stream.

packages/node-http-handler/src/node-http2-handler.spec.ts
```diff
@@ -28,288 +29,420 @@ describe("NodeHttp2Handler", () => {
...
+
+        // Make second request on URL with port2.
+        await nodeH2Handler.handle(new HttpRequest({ ...getMockReqOptions(), port: port2 }), {});
+
+        const authorityPrefix = `${protocol}//${hostname}`;
+        expect(connectSpy).toHaveBeenCalledTimes(2);
+        expect(connectSpy).toHaveBeenNthCalledWith(1, `${authorityPrefix}:${port}`);
+        expect(connectSpy).toHaveBeenNthCalledWith(2, `${authorityPrefix}:${port2}`);
+        mockH2Server2.close();
+      });
```
AllanZhengYP, line 104: Another unit test for `"is many for same URL but concurrent stream is disabled"`

packages/node-http-handler/src/node-http2-handler.spec.ts
```diff
@@ -28,288 +29,420 @@ describe("NodeHttp2Handler", () => {
...
+            },
+          })
+        ).rejects.toHaveProperty("name", "AbortError");
+        // @ts-ignore: access private property
+        expect(nodeH2Handler.sessionCache.size).toBe(0);
       });
-      mockH2Server3.on("connection", () => {
-        establishedConnections += 1;
+
+      it("will not create request on session if request already aborted", async () => {
```
AllanZhengYP, line 250: Does this make sense. why do we need to create a new session if a request is aborted?

packages/node-http-handler/src/node-http2-handler.spec.ts
```diff
@@ -28,288 +29,420 @@ describe("NodeHttp2Handler", () => {
...
-      await expect(
-        nodeH2Handler.handle(new HttpRequest(getMockReqOptions()), {
-          abortSignal: {
-            aborted: true,
-            onabort: null,
-          },
-        })
-      ).rejects.toHaveProperty("name", "AbortError");
-      expect(requestSpy.mock.calls.length).toBe(0);
+    describe("number calls to http2.connect", () => {
```
AllanZhengYP, line 385: duplicated to line 57?

packages/node-http-handler/src/node-http2-handler.spec.ts
```diff
@@ -28,288 +29,420 @@ describe("NodeHttp2Handler", () => {
...
-          })
-      );
-
-      await expect(
-        nodeH2Handler.handle(new HttpRequest(getMockReqOptions()), {
-          abortSignal: abortController.signal
-        })
-      ).rejects.toHaveProperty("name", "AbortError");
-      expect(requestSpy.mock.calls.length).toBe(1);
+    describe("destroy", () => {
```
AllanZhengYP, line 436: duplicated to line 216?

packages/node-http-handler/src/node-http2-handler.spec.ts
trivikr: The concurrent stream disabled tests are in their own section on line 378.
The section starts with `describe("disableConcurrentStreams", () => {`

The specific test mentioned `is many for same URL but concurrent stream is disabled` is called `is many if multiple requests are made on same URL`, and is on line 402.

packages/node-http-handler/src/node-http2-handler.spec.ts
trivikr: This test verifies that new request is not created on existing session, if request is already aborted. It does not create a new session.
* A session is created when first request is made on line 252.
* Second request is not made when handle is called on line 259.

packages/node-http-handler/src/node-http2-handler.spec.ts
trivikr: No. This section is for `disableConcurrentStreams: true`
The section on line 57 is for `disableConcurrentStreams: false`

packages/node-http-handler/src/node-http2-handler.spec.ts
trivikr: No. This section is for `disableConcurrentStreams: true`
The section on line 216 is for `disableConcurrentStreams: false`

alexforsyth (commented):
LGTM, but I lack the context. i'll wait for @AllanZhengYP to approve

AllanZhengYP (approved):
Looks good to me!

PR #2572, opened 2021-07-12
Create waitForDNSFirewallRuleGroupNotShared.ts

gsingh04:
waiter to wait for rule group shared status change to NOT_SHARED

### Description
Waiter to wait for DNS Firewall Rule Group share status transition to NOT_SHARED

### Testing
```
// waiting for rule group share status to change to NOT_SHARED
await waitUntilDNSFirewallRuleGroupNotShared(
  { client: route53Client, maxWaitTime: 300 }, // aggressive timeout, as there can be delay in deleting RAM resource share
  { FirewallRuleGroupId: ruleGrp.Id }
);

await route53Client.send(
  new DeleteFirewallRuleGroupCommand({
    FirewallRuleGroupId: ruleGrp.Id,
  })
);

```
waited for rule group share status to change before deletion, rule group cannot be deleted if it is shared

---

### DIFF
--- a/clients/client-route53resolver/waiters/waitForDNSFirewallRuleGroupNotShared.ts
+++ b/clients/client-route53resolver/waiters/waitForDNSFirewallRuleGroupNotShared.ts
@@ -1,0 +1,51 @@
+import {
+  GetFirewallRuleGroupCommand,
+  GetFirewallRuleGroupCommandInput,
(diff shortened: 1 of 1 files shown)


kuhe:
Waiters are code-generated and thus we cannot accept contributions to them. Requests for new waiters should go to the AWS service team through your AWS account manager.

PR #2592, opened 2021-07-16
chore(clients): call postinstall-node-version-check

trivikr:
### Issue
Internal JS-2727

### Description
Calls postinstall-node-version-check in clients

### Testing
Tested in package `@aws-sdk/client-s3`:

```console
$ fnm use 10
Using Node v10.24.1

$ yarn postinstall
yarn run v1.22.10
$ postinstall-node-version-check
(node:21294) NodeDeprecationWarning: The AWS SDK for JavaScript (v3) will
no longer support Node.js v10.24.1 as of January 1, 2022.
To continue receiving updates to AWS services, bug fixes, and security
updates please upgrade to Node.js 12.x or later.

More information can be found at: https://a.co/1l6FLnu
Done in 0.11s.

$ fnm use 12
Using Node v12.22.1

$ yarn postinstall
yarn run v1.22.10
$ postinstall-node-version-check
Done in 0.10s.
```

---
By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #2594, opened 2021-07-16
chore(clients): call postinstall-node-version-check

trivikr:
### Issue
Follow-up to https://github.com/aws/aws-sdk-js-v3/pull/2592

### Description
Calls postinstall-node-version-check in clients

### Testing
Tested in package `@aws-sdk/client-s3`:

```console
$ fnm use 8
Using Node v8.17.0

$ yarn postinstall
yarn run v1.22.10
error @aws-sdk/client-s3@3.21.0: The engine "node" is incompatible with this module. Expected version ">=10.0.0". Got "8.17.0"
error Commands cannot run with an incompatible environment.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.

$ fnm use 10
Using Node v10.24.1

$ yarn postinstall
yarn run v1.22.10
$ postinstall-node-version-check
Done in 0.11s.
```

---

### DIFF
--- a/clients/client-accessanalyzer/package.json
+++ b/clients/client-accessanalyzer/package.json
@@ -44,7 +44,7 @@
     "@aws-sdk/middleware-user-agent": "3.20.0",
     "@aws-sdk/node-config-provider": "3.20.0",
     "@aws-sdk/node-http-handler": "3.21.0",
-    "@aws-sdk/postinstall-node-version-check": "3.0.0",
+    "postinstall-node-version-check": "0.2.4",
     "@aws-sdk/protocol-http": "3.20.0",
     "@aws-sdk/smithy-client": "3.20.0",
     "@aws-sdk/types": "3.20.0",
--- a/clients/client-accessanalyzer/package.json
+++ b/clients/client-accessanalyzer/package.json
@@ -12,7 +12,8 @@
     "build:cjs": "tsc -p tsconfig.json",
     "build:es": "tsc -p tsconfig.es.json",
     "build": "yarn build:cjs && yarn build:es",
-    "downlevel-dts": "downlevel-dts dist/types dist/types/ts3.4"
+    "downlevel-dts": "downlevel-dts dist/types dist/types/ts3.4",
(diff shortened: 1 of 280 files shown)


G-Rath:
Just an FYI package managers will typically not cache packages that have `postinstall` (as well as other differences in behaviour), as it's meant to be for running additional build steps (i.e, compiling native C libs) that could be OS and environment specific, so this change will impact every downstream consumer (and their CIs) more than you might expect.

A more idiomatic way of enforcing the node version would be to specify it in the `engines` field, which package managers check when installing.

trivikr:
> A more idiomatic way of enforcing the node version would be to specify it in the engines field, which package managers check when installing.

The aim of this change is to notify about Node.js 10.x deprecation for a future date, i.e. `2022-01-01` and not to deprecated Node.js 10.x right away.

> this change will impact every downstream consumer (and their CIs) more than you might expect

Is there a better way to notify users of the pending deprecation you can provide?
We are going to publish a blog post, and depending on Node.js 10.x usage drop we'll also update API reference or Developer Guide. But we wanted to notify developers using the SDK directly.

G-Rath:
@trivikr the best idiomatic way I think would be to publish the new `> 10` major alongside the current `>= 10` major, but have those versions be marked as [`deprecated`](https://docs.npmjs.com/cli/v7/commands/npm-deprecate) with the message being "Support for Node 10 will be dropped at some point - ideally switch to v<whatever the version is>".

Since you're using transpiling tools, this shouldn't be a big increase in maintenance aside from a one-off bit of work to rework your CI to handle publishing both major versions & deprecating one of them.

If you were confident in your setup, you could even have the packages actually be compiled for the different minimum node versions, so the `> 10` major wouldn't have unnecessarily transpiled code - but that wouldn't be required for this setup to work.

trivikr:
Closing as we would be calling `emitWarningIfUnsupportedVersion` in runtimeConfig after https://github.com/aws/aws-sdk-js-v3/pull/2574 is merged.

PR #2637, opened 2021-08-03
fix(util-body-length-node): calculateBodyLength for Readable stream with range

pingiun:
### Description
A ReadStream created with createReadStream(file, {start: 0, end: 0}) would not be detected by this function, or would falsly just use the size of the original file at body.path. This uses the end and start numbers if present, both are inclusive so 1 is added. 

Inclusive according to: https://nodejs.org/api/fs.html#fs_fs_createreadstream_path_options

### Additional context

I encountered this bug while uploading a multipart file to s3

---

### DIFF
--- a/packages/util-body-length-node/src/index.spec.ts
+++ b/packages/util-body-length-node/src/index.spec.ts
@@ -1,3 +1,8 @@
+import { promises } from 'fs';
+import { createReadStream, writeFile } from "fs-extra";
+import * as os from "os";
+import * as path from "path";
+
 import { calculateBodyLength } from "./";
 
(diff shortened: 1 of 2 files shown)


AllanZhengYP (commented):
This change looks good, can you add a unit test?

 https://github.com/aws/aws-sdk-js-v3/blob/74b3f15f1a47cf8077908525b5a767c4aa606317/packages/util-body-length-node/src/index.spec.ts

pingiun:
Note that this still doesn't handle filestreams with just an fd and no path, also doesn't handle a start but no end

kuhe:
this package has moved, opened new PR at the new site

PR #2650, opened 2021-08-08
docs(client-cognito-identity-provider): Update `AllowedOAuthFlows` with allowed options

danielsharvey:
Cognito does not allow combining `client_credentials` with either of `code` 
or `implicit`; updating documentation to note this.

### Issue
N/A

### Description
An error is produced if you attempt to do this; I have updated the
documentation to note this.

### Testing
N/A

### Additional context
N/A

---


kuhe:
The client model is generated from upstream models and cannot be modified. 

For issues related to AWS documentation, use the provide feedback option on the original AWS documentation page, such as https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-cognito-userpoolclient.html.

danielsharvey:
@kuhe Is the suggested correction valid? Can you submit it upstream?

PR #2662, opened 2021-08-11
chore(clients): parse numbers more strictly

JordonPhillips:
### Issue

n/a

### Description

This provides methods to more strictly parse floats and ints. The strict float parser only allows explicit NaN and does not attempt to parse non-string, non-number values. The strict int parser disallows NaN entirely, and will opt to error if there are any decimals present. It also will not parse non-string, non-number values.

### Testing

Unit tests were added for new methods. This also fixes a number of generated tests not released yet in Smithy.

### Additional context

JS is very loose about what it will convert to a number, and we don't want to be nearly so loose, especially in the SSDK.

Related: https://github.com/awslabs/smithy-typescript/pull/397

---


adamthom-amzn (approved):

packages/smithy-client/src/parse-utils.ts
```diff
@@ -67,14 +84,41 @@ export const expectString = (value: any): string | undefined => {
 };
 
 /**
- * Asserts a value is a number and returns it, and also converts string
- * representations of non-numeric floats into Numbers.
+ * Parses a value into a float. If the value is null or undefined, undefined
+ * will be returned. If the value is a string, it will be parsed by the standard
+ * parseFloat with one exception: NaN may only be explicitly set as the string
+ * "NaN", any implicit nans will result in an error being thrown. If any other
```
trivikr, line 90:
```suggestion
 * "NaN", any implicit NaN values will result in an error being thrown. If any other
```

packages/smithy-client/src/parse-utils.spec.ts
```diff
@@ -93,6 +92,28 @@ describe("expectNumber", () => {
...
 
+describe("expectInt", () => {
+  it("accepts integers", () => {
+    expect(expectInt(1)).toEqual(1);
+  });
+
+  it("accepts null/undefined", () => {
+    expect(expectInt(null)).toEqual(undefined);
+    expect(expectInt(undefined)).toEqual(undefined);
+  });
```
trivikr, line 103: Use [jest-each](https://jestjs.io/docs/api#testeachtablename-fn-timeout) for more descriptive tests here and in other tests.
```suggestion
  it.each([null, undefined])("accepts %s", (value) => {
    expect(expectInt(value)).toEqual(undefined);
  });
```

packages/smithy-client/src/parse-utils.spec.ts
```diff
@@ -93,6 +92,28 @@ describe("expectNumber", () => {
...
+    expect(() => expectInt(1.1)).toThrowError();
+    expect(() => expectInt("1")).toThrowError();
+    expect(() => expectInt("1.1")).toThrowError();
+    expect(() => expectInt(NaN)).toThrowError();
+    expect(() => expectInt(true)).toThrowError();
+    expect(() => expectInt(false)).toThrowError();
+    expect(() => expectInt([])).toThrowError();
+    expect(() => expectInt({})).toThrowError();
+  });
+});
```
trivikr, line 115:
```suggestion
  describe("rejects non-integers", () => {
    it.each([1.1, "1", "1.1", NaN, true, [], {}])("rejects %s", (value) => {
      expect(() => expectInt(value)).toThrowError();
    });
  });
});
```

trivikr (approved):

PR #2695, opened 2021-08-20
feat(codegen): support malformed request tests

JordonPhillips:
### Issue

n/a

### Description

This adds suppressions for a number of malformed request tests which are known to fail currently. 

### Testing

Built `main` on the smithy repo and ran the malformed request tests.

### Additional context

Malformed request tests exist for testing the SSDK against requests that should be rejected by virtue of being malformed in some way. For instance, there are a number of tests asserting strict timestamp handling behavior that isn't implemented yet.

The tests in question are still awaiting release in the next version of Smithy, but since the actual traits and supporting infrastructure were released in 1.11.0 we can do this now so that upgrading the code generator to the next version of Smithy is painless.

---


JordonPhillips:
Is main failing? iirc that codebuild test only runs the already-generated TS code, which this doesn't touch

gosar (approved):

AllanZhengYP (approved):

PR #2729, opened 2021-08-30
fix(parse): reject numeric strings with non-numbers

JordonPhillips:
### Issue

N/A

### Description

This updates the number parsing utilities to reject strings like "1A", which `parseFloat` would ordinarily happily accept as the value `1`. It also updates it to reject the wide swath of alternative number formats that JS numbers can have.

### Testing

Additional unit tests were added.

### Additional context

---

### DIFF
--- a/packages/smithy-client/src/parse-utils.ts
+++ b/packages/smithy-client/src/parse-utils.ts
@@ -233,15 +233,8 @@ export const expectString = (value: any): string | undefined => {
  * @returns The value as a number, or undefined if it's null/undefined.
  */
 export const strictParseDouble = (value: string | number): number | undefined => {
-  if (value === "NaN") {
-    return NaN;
-  }
   if (typeof value == "string") {
-    const parsed: number = parseFloat(value);
-    if (Number.isNaN(parsed)) {
-      throw new TypeError(`Expected real number, got implicit NaN`);
-    }
-    return expectNumber(parsed);
+    return expectNumber(parseNumber(value));
   }
   return expectNumber(value);
 };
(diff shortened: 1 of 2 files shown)


adamthom-amzn (approved):

packages/smithy-client/src/parse-utils.ts
```diff
@@ -262,19 +255,28 @@ export const strictParseFloat = strictParseDouble;
...
   return expectFloat32(value);
 };
 
+// This regex matches JSON-style numbers. In short:
+// * The integral may start with a negative sign, but not a positive one
+// * No leading 0 on the integral unless it's immediately followed by a '.'
+// * Exponent indicated by a case-insensitive 'E' optionally followed by a
+//   positive/negative sign and some number of digits.
+// It also matches both positive and negative infinity as well and explicit NaN.
+const NUMBER_REGEX = /(-?((0(\.\d+)?)|(([1-9]\d*(\.\d)?)\d*)((e|E)(\+|-)?\d+)?))|(-?Infinity)|(NaN)/g;
```
adamthom-amzn, line 270: Saw `/-?(?:0|[1-9]\d*)(?:\.\d+)?(?:[eE][+-]?\d+)?/` on stack overflow, which seems like a pretty compact way to write it, though I think this is equivalent.

packages/smithy-client/src/parse-utils.ts
```diff
@@ -262,19 +255,28 @@ export const strictParseFloat = strictParseDouble;
...
+
+const parseNumber = (value: string): number => {
+  const matches = value.match(NUMBER_REGEX);
+  if (matches === null || matches[0].length !== value.length) {
```
adamthom-amzn, line 274: Can you anchor the regex so you can just use `RegExp.test` and not have to check the length of the match?

adamthom-amzn (approved):

packages/smithy-client/src/parse-utils.ts
JordonPhillips: hahaha your faith in js is heartwarming, no it'll match partials anyway

packages/smithy-client/src/parse-utils.ts
JordonPhillips: Yeah I didn't really spend any time regex-golfing, I probably should have just googled it in the first place

packages/smithy-client/src/parse-utils.ts
```diff
@@ -262,19 +255,28 @@ export const strictParseFloat = strictParseDouble;
...
   return expectFloat32(value);
 };
 
+// This regex matches JSON-style numbers. In short:
+// * The integral may start with a negative sign, but not a positive one
+// * No leading 0 on the integral unless it's immediately followed by a '.'
+// * Exponent indicated by a case-insensitive 'E' optionally followed by a
+//   positive/negative sign and some number of digits.
+// It also matches both positive and negative infinity as well and explicit NaN.
+const NUMBER_REGEX = /(-?(?:0|[1-9]\d*)(?:\.\d+)?(?:[eE][+-]?\d+)?)|(-?Infinity)|(NaN)/g;
```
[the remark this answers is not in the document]
adamthom-amzn: It might be overkill at this point (I think the concern was [stuff like this](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseInt#octal_interpretations_with_no_radix)) but we can always loosen this later. It's much more difficult to tighten it.

PR #2737, opened 2021-09-01
feat(smithy-client): support strict timestamp parsing

adamthom-amzn:
### Description
From context, we know what the format of a timestamp should be, and we can
restrict the input to match that format rather than relying on the unspecified
behavior of new Date(string).

### Testing
Unit tests, plus regeneration of clients, client protocol tests, server protocol tests. (those PRs to follow)

### Additional context
- We generally try to avoid adding dependencies, which is why I wrote these functions here, but there's potential for even more verification (like day of week) that I'm avoiding altogether that we might not have to with a real date/time library.
- We could write a more performant parser without using regex, but getting the behavior and abstraction right first is more important to me.
- I was split between putting these in date-utils or parse-utils, I'm open to suggestions.

---

### DIFF
--- a/packages/smithy-client/src/date-utils.spec.ts
+++ b/packages/smithy-client/src/date-utils.spec.ts
@@ -1,0 +1,214 @@
+import { parseEpochTimestamp, parseRfc3339DateTime, parseRfc7231DateTime } from "./date-utils";
+
+describe("parseRfc3339DateTime", () => {
+  it.each([null, undefined])("returns undefined for %s", (value) => {
+    expect(parseRfc3339DateTime(value)).toBeUndefined();
+  });
+
+  describe("parses properly formatted dates", () => {
+    it("with fractional seconds", () => {
+      expect(parseRfc3339DateTime("1985-04-12T23:20:50.52Z")).toEqual(new Date(Date.UTC(1985, 3, 12, 23, 20, 50, 520)));
+    });
+    it("without fractional seconds", () => {
+      expect(parseRfc3339DateTime("1985-04-12T23:20:50Z")).toEqual(new Date(Date.UTC(1985, 3, 12, 23, 20, 50, 0)));
+    });
+    it("with leap seconds", () => {
+      expect(parseRfc3339DateTime("1990-12-31T15:59:60Z")).toEqual(new Date(Date.UTC(1990, 11, 31, 15, 59, 60, 0)));
+    });
+    it("with leap days", () => {
+      expect(parseRfc3339DateTime("2004-02-29T15:59:59Z")).toEqual(new Date(Date.UTC(2004, 1, 29, 15, 59, 59, 0)));
+    });
+    it("with leading zeroes", () => {
+      expect(parseRfc3339DateTime("0004-02-09T05:09:09.09Z")).toEqual(new Date(Date.UTC(4, 1, 9, 5, 9, 9, 90)));
+      expect(parseRfc3339DateTime("0004-02-09T00:00:00.00Z")).toEqual(new Date(Date.UTC(4, 1, 9, 0, 0, 0, 0)));
+    });
+  });
+
+  it.each([
+    "85-04-12T23:20:50.52Z",
+    "985-04-12T23:20:50.52Z",
+    "1985-13-12T23:20:50.52Z",
+    "1985-00-12T23:20:50.52Z",
+    "1985-4-12T23:20:50.52Z",
+    "1985-04-32T23:20:50.52Z",
+    "1985-04-00T23:20:50.52Z",
+    "1985-04-05T24:20:50.52Z",
+    "1985-04-05T23:61:50.52Z",
+    "1985-04-05T23:20:61.52Z",
+    "1985-04-31T23:20:50.52Z",
+    "2005-02-29T15:59:59Z",
+    "Mon, 31 Dec 1990 15:59:60 GMT",
+    "Monday, 31-Dec-90 15:59:60 GMT",
+    "Mon Dec 31 15:59:60 1990",
+    "1985-04-12T23:20:50.52Z1985-04-12T23:20:50.52Z",
+    "1985-04-12T23:20:50.52ZA",
+    "A1985-04-12T23:20:50.52Z",
+  ])("rejects %s", (value) => {
+    expect(() => parseRfc3339DateTime(value)).toThrowError();
+  });
+});
+
+describe("parseRfc7231DateTime", () => {
+  it.each([null, undefined])("returns undefined for %s", (value) => {
+    expect(parseRfc7231DateTime(value)).toBeUndefined();
+  });
+
+  describe("parses properly formatted dates", () => {
+    describe("with fractional seconds", () => {
+      it.each([
+        ["imf-fixdate", "Sun, 06 Nov 1994 08:49:37.52 GMT"],
+        ["rfc-850", "Sunday, 06-Nov-94 08:49:37.52 GMT"],
+        ["asctime", "Sun Nov  6 08:49:37.52 1994"],
+      ])("in format %s", (_, value) => {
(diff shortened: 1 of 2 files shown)


AllanZhengYP (commented):
I didn't look very deep into all the date RFCs, only a few minor points to the syntax. You can leave the function in `date-util`. Otherwise, most other function in the `smithy-client` are `parser-util`.

packages/smithy-client/src/date-utils.ts
```diff
@@ -33,3 +36,232 @@ export function dateToUtcString(date: Date): string {
 
   return `${days[dayOfWeek]}, ${dayOfMonthString} ${months[month]} ${year} ${hoursString}:${minutesString}:${secondsString} GMT`;
 }
+
+const daysInMonth = [31, 28, 31, 30, 31, 30, 31, 31, 30, 31, 30, 31];
```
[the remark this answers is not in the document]
adamthom-amzn: I was matching the style of the file previously (like days and months, above). I'll go ahead and change all constants to be consistent.

packages/smithy-client/src/date-utils.ts
```diff
@@ -33,3 +36,232 @@ export function dateToUtcString(date: Date): string {
...
+  return new Date(
+    Date.UTC(
+      year,
+      adjustedMonth,
+      day,
+      parseDateValue(match.groups!["H"]!, "hour", 0, 23),
+      parseDateValue(match.groups!["m"]!, "minute", 0, 59),
+      // seconds can go up to 60 for leap seconds
+      parseDateValue(match.groups!["s"]!, "seconds", 0, 60),
+      parseMilliseconds(match.groups!["frac"])
```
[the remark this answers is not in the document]
adamthom-amzn: There's no official support for fractional seconds in any of these RFCs other than 3339 as far as I can tell. It's an extension we have [documented](https://awslabs.github.io/smithy/1.0/spec/core/protocol-traits.html#timestampformat-trait) in Smithy, and I just went ahead and applied it consistently to all of the formats.

packages/smithy-client/src/date-utils.ts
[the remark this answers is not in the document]
gosar: nit: might move line just below `const months` since the order is related. Or just before `const validateDayOfMonth` if you want to keep it closer to where it is used.

packages/smithy-client/src/date-utils.spec.ts
```diff
@@ -0,0 +1,180 @@
...
+    });
+    it("with leap seconds", () => {
+      expect(parseRfc3339DateTime("1990-12-31T15:59:60Z")).toEqual(new Date(Date.UTC(1990, 11, 31, 15, 59, 60, 0)));
+    });
+    it("with leap days", () => {
+      expect(parseRfc3339DateTime("2004-02-29T15:59:59Z")).toEqual(new Date(Date.UTC(2004, 1, 29, 15, 59, 59, 0)));
+    });
+  });
+
+  it.each([
```
JordonPhillips, line 23: You should have tests in all of these where there's some extra leading or trailing garbage to make sure the whole string is matched.

packages/smithy-client/src/date-utils.ts
```diff
@@ -31,5 +34,267 @@ export function dateToUtcString(date: Date): string {
...
+    return undefined;
+  }
+  if (typeof value !== "string") {
+    throw new TypeError("RFC-3339 date-times must be expressed as strings");
+  }
+  const match = RFC3339.exec(value);
+  if (!match || !match.groups) {
+    throw new TypeError("Invalid RFC-3339 date-time value");
+  }
+  const year = strictParseShort(match.groups["Y"])!;
```
JordonPhillips, line 68: nit: You can just use `parseNumber` here to avoid unnecessary validation since you already have asserted that it's exactly 4 digits in your regex. Actually.... now that I think about it you'll probably have to do your own thing that just discards leading zeroes if you want to be strictly correct - so that dates prior to Jan 1st 1000 can be represented.

packages/smithy-client/src/date-utils.ts
```diff
@@ -31,5 +34,267 @@ export function dateToUtcString(date: Date): string {
...
+export const parseEpochTimestamp = (value: unknown): Date | undefined => {
+  if (value === null || value === undefined) {
+    return undefined;
+  }
+
+  let valueAsDouble: number;
+  if (typeof value === "number") {
+    valueAsDouble = value;
+  } else if (typeof value === "string") {
+    // TODO: remove when https://github.com/aws/aws-sdk-js-v3/pull/2729 is merged, maybe
```
JordonPhillips, line 158: this is merged

packages/smithy-client/src/date-utils.ts
```diff
@@ -31,5 +34,267 @@ export function dateToUtcString(date: Date): string {
...
+ * adjustRfc850Year will adjust '11' to 2011, (as 2111 is more than 50 years in the future),
+ * but keep '22' as 2022. in 2099, '11' will represent '2111', but '98' should be '2098'.
+ * There's no description of an RFC 850 date being considered too far in the past in RFC-7231,
+ * so it's entirely possible that 2011 is a valid interpretation of '11' in 2099.
+ * @param value the 2 digit year to parse
+ * @return number a year that is equal to or greater than the current UTC year
+ */
+const parseTwoDigitYear = (value: string): number => {
+  const thisYear = new Date().getUTCFullYear();
+  const valueInThisCentury = Math.floor(thisYear / 100) * 100 + strictParseShort(value)!;
```
JordonPhillips, line 214: You definitely can't use `strictParseShort` here - `02` is valid for 2002 for example but would be rejected for the leading 0.

packages/smithy-client/src/date-utils.ts
```diff
@@ -31,5 +34,267 @@ export function dateToUtcString(date: Date): string {
...
+    // TODO: remove when https://github.com/aws/aws-sdk-js-v3/pull/2729 is merged, maybe
+    if (!VALID_NUMBER.exec(value)) {
+      throw new TypeError("Epoch timestamps must be valid floating point numbers.");
+    }
+    valueAsDouble = strictParseDouble(value)!;
+  } else {
+    throw new TypeError("Epoch timestamps must be expressed as floating point numbers or their string representation");
+  }
+
+  if (isNaN(valueAsDouble) || valueAsDouble === Infinity || valueAsDouble === -Infinity) {
```
JordonPhillips, line 167: `Number.isNaN` will reject anything that isn't a number type.

```suggestion
  if (Number.isNaN(valueAsDouble) || valueAsDouble === Infinity || valueAsDouble === -Infinity) {
```

JordonPhillips (commented):
I put this in a few spots, but you're gonna have to account for leading zeroes being valid all over the place, where the `strictParseX` methods explicitly reject them. It'd be great if you could just use the built in `parseInt` but that'll probably take a single leading zero as an octal.

JordonPhillips (changes requested):
Whoops, meant to do request changes

packages/smithy-client/src/date-utils.spec.ts
adamthom-amzn: Sure. It is; I don't know why anchors didn't work for you previously but they do for me, I verified after our dicussion yesterday.

JordonPhillips: Teach me your ways magic man

packages/smithy-client/src/date-utils.ts
adamthom-amzn: Stripped leading 0s from every component before parsing, and added tests to verify components with leading 0s as well as components that are all 0.

AllanZhengYP (approved):

JordonPhillips (approved):

PR #2860, opened 2021-10-01
feat(codegen): do not pull client versions from packageInfo

adamthom-amzn:
### Description
To avoid having to @ts-ignore importing packageInfo from package.json,
write the version contained in the typescript settings into the
UserAgent provider's arguments. This removes the need to postprocess
clients after they are generated.

### Testing
I generated clients locally, as well as a white-label client that failed without manual postprocessing.

### Additional context
This is just a demonstration of how we could drive versioning through updates to the central package.json, instead of lerna, since we generate a smithy-build that we can inject a version into before generation. 

---


trivikr:
Superceded by https://github.com/aws/aws-sdk-js-v3/pull/2875

PR #2906, opened 2021-10-16
feat(lib-dynamodb): add pagination for query and scan

Sordie:
### Issue
N/A

### Description
Analogous to the implementation in @aws-sdk/client-dynamodb, pagination would also be immensely helpful in @aws-sdk/lib-dynamodb. The two relevant commands are query and scan. The implementation is the same as in @aws-sdk/client-dynamodb except that the clients and commands point their @aws-sdk/lib-dynamodb counterparts.

### Testing
Build locally and executed both new functions.

---


samchungy:
I did something similar before but their code is codegened friendo, you'd need to contribute to their smithy repo

trivikr:
This was added in https://github.com/aws/aws-sdk-js-v3/pull/3069, and released with `v3.43.0`

PR #2928, opened 2021-10-25
fix(dependencies): move production deps to dependencies

voidus:
Having them only in devDependencies doesn't work because they are
imported in production code

Right now, we are overriding these in our yarnrc, but it would be nice to fix it for everyone

---

### DIFF
--- a/packages/config-resolver/package.json
+++ b/packages/config-resolver/package.json
@@ -18,12 +18,12 @@
   },
   "license": "Apache-2.0",
   "devDependencies": {
-    "@aws-sdk/node-config-provider": "3.38.0",
     "@types/jest": "^26.0.4",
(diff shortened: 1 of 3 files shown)


trivikr:
The equivalent PR in `@smithy/` packages was posted at https://github.com/awslabs/smithy-typescript/pull/905

trivikr:
> The equivalent PR in `@smithy/` packages was posted at [awslabs/smithy-typescript#905](https://github.com/awslabs/smithy-typescript/pull/905)

The fix in `@smithy/` packages was merged.

trivikr:
The fix for middleware-bucket-endpoint is posted at https://github.com/aws/aws-sdk-js-v3/pull/5196

PR #3011, opened 2021-11-11
fix(codegen): use specific Smithy version

adamthom-amzn:
### Description
This centralizes the Smithy version in the root Gradle file for ease of use,
and locks the CLI to the same Smithy version as the Smithy dependencies.

### Testing
Built locally in combination with https://github.com/awslabs/smithy-typescript/pull/465

---


AllanZhengYP (approved):
🐑🇮🇹

trivikr (approved):

PR #3086, opened 2021-12-09
chore(hash-stream-node): move fileStreamHasher to own file

trivikr:
### Issue
Preparing for adding hasher for non-file streams

### Description
Moves fileStreamHasher to own file.

### Testing
CI

---


AllanZhengYP (approved):
Approved. One thing to notice here is that the current file stream hasher returns the hash before the request payload is even start to be sent. However, for a real stream hasher, the hash can only be returned util all the payload stream has been sent. So you cannot use it in the request header. We need to find ways to notify this difference.

Issue #2263, opened 2021-04-15
Uncaught Exception complicates debugging


netbymatt:
SDK v3.12.0
Node v14.15.0
Visual Studio Code v1.55.2 for debugging
Credentials loaded from ~/.aws/credentials
Actual package loaded @aws-sdk/client-s3 v3.12.0

I typically debug with "uncaught exceptions" turned on. The code in line 16 below pre-generates a default rejected promise that is then typically overwritten in the loop on the following lines. 

https://github.com/aws/aws-sdk-js-v3/blob/4f9f4a79b019c1c35530e05c4138fde88a6fb547/packages/property-provider/src/chain.ts#L16-L25

Promise.reject is considered an uncaught exception by the debugger thus it pauses here on each debug run. Even though the credentials are ultimately successfully loaded from the default location ~/.aws/credentials I need to step through this "exception" before actually getting to my code that I'm targeting for debug.

I believe a better solution would be to restructure the loop so the final return is where the rejected promise is generated.

kilotone:
This has been driving me absolutely mad the last 3 days

vargaendre:
Still happening with version 3.85.0
I also spent like two hours trying to figure out what I did wrong. Turns out that nothing.
I am glad this issue exists at least!

ppold:
Has anyone found a workaround to silence that exception in the meantime?

belenkymr:
Still happening with [v3.127.0](https://github.com/aws/aws-sdk-js-v3/releases/tag/v3.127.0)

SabeehY:
+1

berkeleymalagon:
+1

chriskovacs:
+1 ...this is driving my team nuts.

nconnector:
+1 this is still happening

RanVaknin:
Hi all,

I have marked this for prioritization, if someone wants to take a stab at a PR that would be very appreciated. 

Thanks,
Ran~

trivikr:
Verified that VSCode pauses on Promise.reject when "Uncaught Exceptions" is turned on.

<details>
<summary>Screen recording</summary>

https://github.com/aws/aws-sdk-js-v3/assets/16024985/54a4fff2-b6c9-4a8c-9d78-18abded150eb

</details>

PR #3180, opened 2022-01-12
fix(property-provider): only reject if providers are not given to avoid debugger pause

omairvaiyani:
In VSCode, when you enable the breakpoint "Uncaught exceptions", the debugger falsely pauses in `chain.js`. This is because the method calls `Promise.reject()` without immediately catching it, therefore confusing the debugger.

### Issue
#2263

### Description
The promise now only creates and returns a rejection when truly required. The chain behaviour is kept using an initial `Promise.resolve()` instance.

### Testing
Against my local environment. The debugger no longer pauses and the providers are still behaving as expected.

### DIFF
--- a/packages/property-provider/src/chain.ts
+++ b/packages/property-provider/src/chain.ts
@@ -13,15 +13,23 @@ import { ProviderError } from "./ProviderError";
  */
 export function chain<T>(...providers: Array<Provider<T>>): Provider<T> {
   return () => {
-    let promise: Promise<T> = Promise.reject(new ProviderError("No providers in chain"));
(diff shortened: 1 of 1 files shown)


wberrytbt:
This is happening on my local development instance as well.  @omairvaiyani [omairvaiyani](https://github.com/omairvaiyani) did u have a workaround?

AllanZhengYP (approved):

kuhe:
https://github.com/awslabs/smithy-typescript/blob/main/packages/property-provider/src/chain.ts

should no longer be an issue since there is no longer a promise rejection happening in the chain function.

PR #3288, opened 2022-02-07
Smithy 1.17.0 update and server protocol test workaround

adamthom-amzn:
### Description
This updates to Smithy 1.17.0 and temporarily disables type checking for ts-jest in the server protocol tests only. We have been unable to find any other way to get server protocol tests to type check correctly.

### Testing
`yarn generate-clients && yarn test:protocols && yarn generate-clients -s && yarn test:server-protocols`

---

### DIFF
--- a/codegen/build.gradle.kts
+++ b/codegen/build.gradle.kts
@@ -31,7 +31,7 @@ allprojects {
     version = "0.8.0"
 }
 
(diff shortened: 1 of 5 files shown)


AllanZhengYP (approved):
Approve this since the tsc passes for the changed code, only jest-ts fails. We will follow up on the root cause of the jest-ts issue.

Issue #3279, opened 2022-02-04
Leaking file descriptors when generating too many http requests


adrai:
*This issue is extracted from this original issue: https://github.com/aws/aws-sdk-js-v3/issues/3019, like requested: https://github.com/aws/aws-sdk-js-v3/issues/3019#issuecomment-1030267030*

#3019 is concentrated to the readFile (leaking file descriptor) "issue" and this issue is concentrated to the network (leaking file descriptor) "issue"

---

It seems lambda is not "waiting" for the file descriptors to be closed.

This can be observed especially, when having warm lambda executions with a lot of sdk calls, like for DynamoDb or S3, etc...
This each http request opens a network socket which results in an open file descriptor.
Since by default in Node.js the socket timeout is set to 120000ms (2 minutes) it may be the lambda is already finished, but the sockets are still open. When "restarting" the lambda for the next invocations, those file descriptors may still be open.
This leads to this type of EMFILE errors:
```
Error: connect EMFILE 52.94.5.100:443 - Local (undefined:undefined)
Error: getaddrinfo EMFILE dynamodb.eu-west-1.amazonaws.com
Error: A system error occurred: uv_os_homedir returned EMFILE (too many open files)
```

These basic tests shows the count (and leaks) of the emfile count:

Tests originally done in this issue here: https://github.com/aws/aws-sdk-js-v3/issues/3019#issuecomment-1028840006

<details>
<summary>Details</summary>

![image](https://user-images.githubusercontent.com/1086194/152322285-081469dd-6327-4559-819e-bcd9741a8bf2.png)
![image](https://user-images.githubusercontent.com/1086194/152324356-dab148b1-1660-4e7e-9301-0f89a14b50e1.png)

</details>

compared to this tests: https://github.com/aws/aws-sdk-js-v3/issues/3019#issuecomment-1029287130

Defining a custom requestHandler, with a very low socketTimeout reduces drastically the emfiles count:
```js
  requestHandler: new NodeHttpHandler({
      socketTimeout: 10000 // <- this decreases the emfiles count, the Node.js default is 120000
  })
```

<details>
<summary>Details</summary>

![image](https://user-images.githubusercontent.com/1086194/152410670-2a46bea7-83b3-41ff-b582-70d6b7e0ca2f.png)
![image](https://user-images.githubusercontent.com/1086194/152410697-8d1376a6-2945-41b9-bcb3-33051b543ff0.png)

</details>

[...]

PR #3291, opened 2022-02-08
fix(node-http-handler): by default use 10 seconds as socketTimeout option, to prevent (#3279)

adrai:
to prevent leaking file descriptors, like described in #3279

### Issue
Fixes: #3279

### Description
Sets a default socketTimeout for the node-http-handler to decrease the amount of open file descriptors (sockets)

### Testing
Was tested in user land described [here](https://github.com/aws/aws-sdk-js-v3/issues/3279) and [here](https://github.com/aws/aws-sdk-js-v3/issues/3019#issuecomment-1029287130)

---

### DIFF
--- a/packages/node-http-handler/src/node-http-handler.ts
+++ b/packages/node-http-handler/src/node-http-handler.ts
@@ -59,7 +59,7 @@ export class NodeHttpHandler implements HttpHandler {
     const maxSockets = 50;
     return {
       connectionTimeout,
-      socketTimeout,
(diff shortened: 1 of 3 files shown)


kuhe:
After your test of https://github.com/aws/aws-sdk-js-v3/pull/3545 this may not be necessary? 

I'm not inclined to change the default value (requiring opt-out and applying unexpected changes to default users) when it is already available as opt-in.

PR #3322, opened 2022-02-16
fix(codegen): fix set deserialization in SSDKs

adamthom-amzn:
### Description
Using JavaScript's Set to detect duplicate items does not work for
non-primitive types. There's a function in server-common to detect duplicates
of any type, and detecting them in server deserialization is much more
important than client deserialization, so the check has been moved to
server-only.

### Testing
New protocol tests in https://github.com/awslabs/smithy/pull/1094

---


adamthom-amzn:
Note that this won't pass CI until https://github.com/awslabs/smithy-typescript/pull/511 is merged

gosar (approved):
Good call on removing it from client code.

PR #3317, opened 2022-02-15
chore(codegen): generate heap dump on OutOfMemory error

trivikr:
### Issue
The OutOfMemory error appears even after we increased maximum size of memory allocation pool to `4g` in https://github.com/aws/aws-sdk-js-v3/pull/3253 

### Description
Temporarily disable manual increase of maximum memory size of memory allocation pool in Java, and generate heap dump on OutOfMemory.

### Testing
The command `yarn generate-clients` is successful in the workspace hosted on `c5.9xlarge`

### Additional context
Java docs: https://docs.oracle.com/javase/7/docs/webnotes/tsg/TSG-VM/html/clopts.html

---
By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #3343, opened 2022-02-21
fix(codegen): increase memory allocation pool maximum size

trivikr:
### Issue
Internal JS-3123

### Description
Sets memory allocation pool maximum size to 4 gigabytes.
* This was removed while testing in https://github.com/aws/aws-sdk-js-v3/pull/3317
* Although we're generating clients in batches from https://github.com/aws/aws-sdk-js-v3/pull/3319, the OutOfMemory gets thrown in the workspace intermittently.

### Testing
Verified that `yarn generate-clients` is successful.

---


trivikr:
Not needed as the issue was with workspace which was using Java `1.8.0_322`.
It was fixed by upgrading to `11.0.13` from `Corretto`

Issue #3377, opened 2022-02-28
S3 throws `NotImplemented` error if readStream is created from a file descriptor


trivikr:
## Describe the bug
S3 throws `NotImplemented` error if readStream is created from a file descriptor

#### SDK version number
`@aws-sdk/client-s3@3.53.1`

#### Is the issue in the browser/Node.js/ReactNative?
Node.js

#### Details of the browser/Node.js/ReactNative version
```console
$ node -v
v16.13.2
```

## Steps to reproduce

Code to run:
```js
import { S3 } from "@aws-sdk/client-s3"; // v3.53.1
import { open } from "fs/promises";

const region = "us-west-2";
const client = new S3({ region });

// File created with command "(for i in {1..1000}; do echo $i "hello world"; done) > hello-world.txt"
const fileName = "hello-world.txt";
const fd = await open(fileName);
const readableStream = fd.createReadStream();

const Bucket = "aws-sdk-js-trivikr";
const Key = fileName;
const Body = readableStream;

try {
  await client.putObject({ Bucket, Key, Body });
} catch (error) {
  console.log({ error });
}
```

#### Observed behavior
Error thrown:
```console
{
  error: NotImplemented: A header you provided implies functionality that is not implemented
      at deserializeAws_restXmlPutObjectCommandError (/local/home/trivikr/workspace/temp/node_modules/@aws-sdk/client-s3/dist-cjs/protocols/Aws_restXml.js:8037:24)
      at processTicksAndRejections (node:internal/process/task_queues:96:5)
      at async /local/home/trivikr/workspace/temp/node_modules/@aws-sdk/client-s3/node_modules/@aws-sdk/middleware-serde/dist-cjs/deserializerMiddleware.js:7:24
      at async /local/home/trivikr/workspace/temp/node_modules/@aws-sdk/client-s3/node_modules/@aws-sdk/middleware-signing/dist-cjs/middleware.js:11:20
      at async StandardRetryStrategy.retry (/local/home/trivikr/workspace/temp/node_modules/@aws-sdk/client-s3/node_modules/@aws-sdk/middleware-retry/dist-cjs/StandardRetryStrategy.js:51:46)
      at async /local/home/trivikr/workspace/temp/node_modules/@aws-sdk/middleware-flexible-checksums/dist-cjs/flexibleChecksumsMiddleware.js:56:20
[...]

PR #2612, opened 2021-07-22
feat(lib-storage): improve performance by reducing buffer copies

Linkgoron:
Reduce memory usage by reducing buffer-copying and returning the original buffer instead.

### Description
This PR reduces the number of buffers copied and created from existing buffers. The most significant change was made to streams that return buffers and Uint8Arrays, but some minor improvements were made for `string` and `Uint8Array` bodies as well. The following changes were made:

- Yield without copying the given Buffer/Uint8Array in `getDataReadable.ts` and `getDataReadableStream.ts` (probably the most significant change).
- In `byteLength.ts`, use `Buffer.byteLength(string)` instead of `Buffer.from(string).byteLength` for computing the string byteLength, so that the string is not copied into a buffer just for its byteLength.
- Treat `Uint8Array` the same as a Buffer in `chunker.ts`
- Use `subarray` instead of `slice` in `getChunkBuffer.ts` so that `Uint8Array` and `Buffer` inputs behave the same. In Node `.slice` on Uint8Array copies the data, while `.subarray` just takes the subarray without copying. Slice and subarray for Buffers in Node are the same (both don't copy the data).
- `getChunkStream.ts` was changed in a similar way to `getChunkBuffer.ts`, and also added a change so that the last  buffer is not copied if there's only one buffer left (the last change is probably relatively minor).

### Testing
I ran the following file for different values:

```ts
import { getChunk } from './chunker';
import { Readable } from 'stream';
const buf = Buffer.from('#'.repeat(1024 * 1024 * 10));
const getIter = () => getChunk(Readable.from((async function* () {
    const chunkSize = 1024 * 1024;
    for (let i = 0; i < buf.byteLength; i += chunkSize) {
        yield buf.subarray(i, i + chunkSize);
    }
})()), 1024 * 1024 * 5)

async function iterateTimes(times: number) {
    console.time('timing')
    for (let i = 0; i < times; i++) {
        for await (const z of getIter()) { }
    }
    console.log(`times ${times}`);
    const used = process.memoryUsage() as any;
    for (let key in used) {
        console.log(`${key} ${Math.round(used[key] / 1024 / 1024 * 100) / 100} MB`);
    }
    console.timeEnd('timing')
}

iterateTimes(10);
```
<details>
<summary>results</summary>
Original code

>>>>>
times 1
rss 59.67 MB
heapTotal 15.68 MB
heapUsed 12.78 MB
external 30.91 MB
arrayBuffers 30.02 MB
timing: 17.289ms

times 2
[...]

PR #3385, opened 2022-02-28
fix(util-body-length-node): support fd.createReadStream

trivikr:
### Issue
Fixes: https://github.com/aws/aws-sdk-js-v3/issues/3377

### Description
Compute length when fs.ReadStream is created using fd.createReadStream

### Testing
Verified that object is successfully put in the following cases:

<details>
<summary>PutObject without Checksum</summary>

```js
import { S3 } from "../aws-sdk-js-v3/clients/client-s3/dist-cjs/index.js";
import { open } from "fs/promises";

const region = "us-west-2";
const client = new S3({ region });

// File created with command "(for i in {1..1000}; do echo $i "hello world"; done) > hello-world.txt"
const fileName = "hello-world.txt";
const fd = await open(fileName);
const readableStream = fd.createReadStream();

const Bucket = "aws-sdk-js-trivikr";
const Key = fileName;
const Body = readableStream;

try {
  await client.putObject({ Bucket, Key, Body });
} catch (error) {
  console.log({ error });
}
```

</details>

<details>
<summary>PutObject with Checksum</summary>

```js
import { S3 } from "../aws-sdk-js-v3/clients/client-s3/dist-cjs/index.js";
import { open } from "fs/promises";

const region = "us-west-2";
const client = new S3({ region });

// File created with command "(for i in {1..1000}; do echo $i "hello world"; done) > hello-world.txt"
const fileName = "hello-world.txt";
const fd = await open(fileName);
const readableStream = fd.createReadStream();

const Bucket = "aws-sdk-js-trivikr";
const Key = fileName;
const Body = readableStream;
const ChecksumAlgorithm = "CRC32";

try {
  await client.putObject({ Bucket, Key, Body, ChecksumAlgorithm });
} catch (error) {
  console.log({ error });
}
```

</details>

### Additional context
~~This PR will be rebased after https://github.com/aws/aws-sdk-js-v3/pull/3384 is merged.~~ Ready!

---

### DIFF
--- a/packages/util-body-length-node/src/calculateBodyLength.ts
+++ b/packages/util-body-length-node/src/calculateBodyLength.ts
@@ -1,4 +1,4 @@
-import { lstatSync } from "fs";
+import { fstatSync, lstatSync } from "fs";
 
 export const calculateBodyLength = (body: any): number | undefined => {
   if (!body) {
@@ -14,5 +14,8 @@ export const calculateBodyLength = (body: any): number | undefined => {
   } else if (typeof body.path === "string" || Buffer.isBuffer(body.path)) {
     // handles fs readable streams
     return lstatSync(body.path).size;
+  } else if (typeof body.fd === "number") {
+    // handles fd readable streams
+    return fstatSync(body.fd).size;
   }
 };
--- a/packages/util-body-length-node/src/calculateBodyLength.spec.ts
+++ b/packages/util-body-length-node/src/calculateBodyLength.spec.ts
@@ -1,12 +1,23 @@
+import { lstatSync } from "fs";
+
 import { calculateBodyLength } from "./calculateBodyLength";
 
-const arrayBuffer = new ArrayBuffer(1);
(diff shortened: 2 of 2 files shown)


packages/util-body-length-node/src/calculateBodyLength.ts
```diff
@@ -0,0 +1,21 @@
+import { fstatSync, lstatSync } from "fs";
+
+export const calculateBodyLength = (body: any): number | undefined => {
+  if (!body) {
+    return 0;
+  }
+  if (typeof body === "string") {
+    return Buffer.from(body).length;
```
Linkgoron, line 8: If you're already changing this method, can I recommend changing this to `Buffer.byteLength(body)` instead? I think that you would have the same behaviour, just without creating a new buffer.

Also, I would be thrilled if you or another member of your team could take a look at my PR which has other performance improving changes: https://github.com/aws/aws-sdk-js-v3/pull/2612

AllanZhengYP (approved):

PR #3442, opened 2022-03-17
test(shared-ini-file-loader): use getter with import for slurpFile

trivikr:
### Issue
Noticed while trying to write tests for files which use global variables in JS-3150

### Description
Uses getter with import for slurpFile

### Testing
CI

---


trivikr:
The official recommendation from jest is to use `jest.resetModules()` and require/import https://jestjs.io/docs/jest-object#jestresetmodules

PR #3475, opened 2022-03-24
chore: remove profile specific data from providers

trivikr:
### Issue
Refs: https://github.com/aws/aws-sdk-js-v3/issues/3469

### Description
Removes profile specific data from providers.
* Since we support setting profile only in AWS_PROFILE, this information is only read from getProfileName().
* Caching of configuration happens at slurpFile level in https://github.com/aws/aws-sdk-js-v3/pull/3285, so the function can be called multiple times and doesn't have to be called at the top.

### Testing
* ToDo: Unit testing
* ToDo: Integration testing
* ToDo: Manual testing to ensure there are only two readFile calls.

---


trivikr:
Superceded by https://github.com/aws/aws-sdk-js-v3/pull/3478

PR #3477, opened 2022-03-24
fix(node-config-provider): remove profile specific data from fromSharedConfigFiles

trivikr:
### Issue
Refs: https://github.com/aws/aws-sdk-js-v3/issues/3469

### Description
Removes profile specific data from fromSharedConfigFiles.
* Since we support setting profile only in AWS_PROFILE, this information is not required from callee.
* Caching of loadedConfig happens at slurpFile level as done in https://github.com/aws/aws-sdk-js-v3/pull/3285, so it's not required to be done by callee.

### Testing
Unit testing

---


trivikr:
To be revisited after `loadedConfig` is removed.

Issue #2807, opened 2021-09-21
@aws-sdk/lib-dynamodb incorrectly marshalls non-object types


abuzhynsky:
## Describe the bug
`QueryCommandInput AttributeValueList` is not marshaled correctly and the `query` command gives me this error:

```json
{
  "error": {
    "name": "ValidationException",
    "message": "One or more parameter values were invalid: Invalid number of argument(s) for the EQ ComparisonOperator",
    "callstack": "ValidationException: One or more parameter values were invalid: Invalid number of argument(s) for the EQ ComparisonOperator\n' +
        '    at deserializeAws_json1_0QueryCommandError (/Users/abuzhinsky/Projects/work/yaft/packages/feature-sync/node_modules/@aws-sdk/client-dynamodb/dist/cjs/protocols/Aws_json1_0.js: 2984: 41)\n' +
        '    at processTicksAndRejections (internal/process/task_queues.js: 97: 5)\n' +
        '    at async /Users/abuzhinsky/Projects/work/yaft/packages/feature-sync/node_modules/@aws-sdk/middleware-serde/dist/cjs/deserializerMiddleware.js: 6: 20\n' +
        '    at async /Users/abuzhinsky/Projects/work/yaft/packages/feature-sync/node_modules/@aws-sdk/middleware-signing/dist/cjs/middleware.js: 11: 20\n' +
        '    at async StandardRetryStrategy.retry (/Users/abuzhinsky/Projects/work/yaft/packages/feature-sync/node_modules/@aws-sdk/middleware-retry/dist/cjs/StandardRetryStrategy.js: 51: 46)\n' +
        '    at async /Users/abuzhinsky/Projects/work/yaft/packages/feature-sync/node_modules/@aws-sdk/middleware-logger/dist/cjs/loggerMiddleware.js: 6: 22\n' +
        '    at async /Users/abuzhinsky/Projects/work/yaft/packages/feature-sync/node_modules/@aws-sdk/lib-dynamodb/dist/cjs/commands/QueryCommand.js: 44: 26\n' +
        '    at async /Users/abuzhinsky/Projects/work/yaft/packages/feature-sync/bin/index.js: 26: 21"
  }
}
```

[...]

PR #3539, opened 2022-04-13
fix(util-dynamodb): allow marshall function to handle more input types

kuhe:
### Issue
Fixes https://github.com/aws/aws-sdk-js-v3/issues/2807 
> @aws-sdk/lib-dynamodb incorrectly marshalls non-object types

### Description
- modify the `marshall` method in util-dynamodb and used by lib-dynamodb
  - currently this only returns the `M` property assuming a map is the input
  - I modified this function to return the M or L properties for maps and lists, but otherwise return the converted attributeValue. 
- separately, I believe the function `processObj` in lib-dynamodb needs to accept falsy values other than undef.

### Testing
added unit/integration specs to cover the cases described in the issue and my other change to lib-dynamodb

### Additional context
`util-dynamodb` is an internal package providing the marshaling functionality.
`lib-dynamodb` is a wrapper for `client-dynamodb` that aims to simplify the object structure to one that is more JavaScript-like.

### DIFF
--- a/packages/util-dynamodb/src/marshall.spec.ts
+++ b/packages/util-dynamodb/src/marshall.spec.ts
@@ -4,64 +4,62 @@ import { marshall } from "./marshall";
 jest.mock("./convertToAttr");
 
 describe("marshall", () => {
-  describe("mocked convertToAttr", () => {
-    const mockOutput = { S: "mockOutput" };
-    (convertToAttr as jest.Mock).mockReturnValue({ M: mockOutput });
+  const mockOutput = { S: "mockOutput" };
+  (convertToAttr as jest.Mock).mockReturnValue({ M: mockOutput });
 
-    afterEach(() => {
-      jest.clearAllMocks();
-    });
+  afterEach(() => {
+    jest.clearAllMocks();
+  });
 
-    it("with object as an input", () => {
-      const input = { a: "A", b: "B" };
-      expect(marshall(input)).toEqual(mockOutput);
-      expect(convertToAttr).toHaveBeenCalledTimes(1);
-      expect(convertToAttr).toHaveBeenCalledWith(input, undefined);
-    });
+  it("with object as an input", () => {
+    const input = { a: "A", b: "B" };
+    expect(marshall(input)).toEqual(mockOutput);
+    expect(convertToAttr).toHaveBeenCalledTimes(1);
+    expect(convertToAttr).toHaveBeenCalledWith(input, undefined);
+  });
 
-    ["convertEmptyValues", "removeUndefinedValues"].forEach((option) => {
-      describe(`options.${option}`, () => {
-        [false, true].forEach((value) => {
-          it(`passes ${value} to convertToAttr`, () => {
-            const input = { a: "A", b: "B" };
-            expect(marshall(input, { [option]: value })).toEqual(mockOutput);
-            expect(convertToAttr).toHaveBeenCalledTimes(1);
-            expect(convertToAttr).toHaveBeenCalledWith(input, { [option]: value });
-          });
+  ["convertEmptyValues", "removeUndefinedValues"].forEach((option) => {
+    describe(`options.${option}`, () => {
+      [false, true].forEach((value) => {
+        it(`passes ${value} to convertToAttr`, () => {
+          const input = { a: "A", b: "B" };
+          expect(marshall(input, { [option]: value })).toEqual(mockOutput);
+          expect(convertToAttr).toHaveBeenCalledTimes(1);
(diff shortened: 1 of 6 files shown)


packages/util-dynamodb/src/marshall-integ.spec.ts
```diff
@@ -0,0 +1,143 @@
+import { marshallInput } from "../../../lib/lib-dynamodb/src/commands/utils";
+import { marshall } from "./marshall";
+import { convertToAttr } from "./convertToAttr";
```
trivikr, line 3: The marshal unit tests need go to `marshall.spec.ts`

packages/util-dynamodb/src/marshall-integ.spec.ts
```diff
@@ -0,0 +1,143 @@
+import { marshallInput } from "../../../lib/lib-dynamodb/src/commands/utils";
```
trivikr, line 1: This can be converted to unit test and moved to lib-dynamodb.

lib/lib-dynamodb/src/commands/utils-integ.spec.ts
```diff
@@ -0,0 +1,11 @@
+import { marshallInput } from "./utils";
```
trivikr, line 1: Rename this file to `.integ.spec.ts`

trivikr (approved):
Code LGTM, just some comments about move some tests to exists unit tests and using `.integ.spec.ts

packages/util-dynamodb/src/marshall-integ.spec.ts
trivikr: If you plan to write integration tests, please using `.integ.spec.ts`

kuhe: I've opted not to use integration tests, or name anything `.integ.spec.ts`. It seems there are no tests of that pattern in our libs and packages and the jest config is incomplete for supporting that. I created a `marshallInput.spec.ts` instead. 

As for the additional `marshall` unit tests, I tried to put them in the existing `marshall.spec.ts` but couldn't find a working jest setup to both mock and unmock `convertToAttr` in the same file. It was expedient to put it in a second file, if that's ok? I tried the jest docs for mocking and unmocking in the same file, but in the short amount of time I spent I couldn't get it working. The setup is also verbose compared to having a second file.

trivikr: > It seems there are no tests of that pattern in our libs

We do. But we use it only for actual integration tests https://github.com/aws/aws-sdk-js-v3/blob/main/clients/client-transcribe-streaming/test/index.integ.spec.ts

> jest config is incomplete for supporting that.


Our default jest config will run based on configurations defined in `jest.config.js` in each related packages.
 https://github.com/aws/aws-sdk-js-v3/blob/99235b45be62706d77ce93f22f8cac736cd9ce1b/jest.config.js#L1-L7
 
The default jest config in util-dynamodb extends from base which skips `*.integ.ts`
 https://github.com/aws/aws-sdk-js-v3/blob/99235b45be62706d77ce93f22f8cac736cd9ce1b/jest.config.base.js#L5

trivikr: > I created a marshallInput.spec.ts instead.

This should work.

packages/util-dynamodb/src/marshall.spec.ts
```diff
@@ -4,62 +4,64 @@ import { marshall } from "./marshall";
 jest.mock("./convertToAttr");
 
 describe("marshall", () => {
-  const mockOutput = { S: "mockOutput" };
-  (convertToAttr as jest.Mock).mockReturnValue({ M: mockOutput });
+  describe("mocked convertToAttr", () => {
```
trivikr, line 7: The change is `marshall.spec.ts` is not needed as the unmocked tests are in marshallInput.spec.ts

packages/util-dynamodb/src/marshallTypes.spec.ts
```diff
@@ -0,0 +1,59 @@
+import { convertToAttr } from "./convertToAttr";
+import { marshall } from "./marshall";
+
+describe("marshall type discernment", () => {
+  describe("behaves as convertToAttr for non-collection values or Sets", () => {
+    it("marshals string", () => {
+      const value = "hello";
+      expect(marshall(value)).toEqual(convertToAttr(value));
+    });
+    it("marshals number", () => {
```
trivikr, line 10: nit: newline between tests for clarity
```suggestion

    it("marshals number", () => {
```

trivikr (approved):

PR #3545, opened 2022-04-14
fix(node-http-handler): resolve config provider only once per NodeHttpHandler instance

kuhe:
### Issue
https://github.com/aws/aws-sdk-js-v3/issues/3279

### Description
- node-http-handler config can be an async provider
- this PR modifies the handle method such that the async provider is only invoked once per instance

### Testing
- added unit tests for this scenario


packages/node-http-handler/src/node-http-handler.ts
```diff
@@ -39,18 +39,19 @@ interface ResolvedNodeHttpHandlerConfig {
...
-    if (typeof options === "function") {
-      this.configProvider = async () => {
-        return this.resolveDefaultConfig(await options());
-      };
-    } else {
-      this.config = this.resolveDefaultConfig(options);
-    }
+    this.configProvider = new Promise(async (resolve) => {
+      if (typeof options === "function") {
+        resolve(this.resolveDefaultConfig(await options()));
```
trivikr, line 50: Using async inside Promise constructor is an [anti-pattern](https://stackoverflow.com/a/43050114)

```suggestion
    this.configProvider = new Promise((resolve) => {
      if (typeof options === "function") {
        options().then((options) => resolve(this.resolveDefaultConfig(options)));
```

packages/node-http-handler/src/node-http-handler.spec.ts
```diff
@@ -38,55 +36,88 @@ describe("NodeHttpHandler", () => {
...
+        for (let i = 0; i < 100; ++i) {
+          promises.push(nodeHttpHandler.handle({} as unknown as HttpRequest));
+        }
+
+        expect(providerInvokedCount).toBe(1);
+        expect(providerResolvedCount).toBe(0);
+        await new Promise((r) => setTimeout(r, 50));
+        expect(providerInvokedCount).toBe(1);
+        expect(providerResolvedCount).toBe(1);
+        await Promise.all(promises);
```
trivikr, line 116: await Promise.all() to call handle in parallel while slowConfigProvide is getting resolved.
```suggestion
        await Promise.all(promises);
        expect(providerInvokedCount).toBe(1);
        expect(providerResolvedCount).toBe(1);
```

trivikr (approved):

AllanZhengYP (approved):

Issue #1920, opened 2021-01-15
Can not pass PassThrough stream to send object


cojack:
**Describe the bug**

There is no possibility to pass stream as type `PassThrough` to the Body. 

**3.3.0**

**Is the issue in the browser/Node.js/ReactNative?**
Node.js

**Details of the browser/Node.js/ReactNative version**
Paste output of `npx envinfo --browsers` or `node -v` or `react-native -v`

 node -v = v15.5.1

**To Reproduce (observed behavior)**
Steps to reproduce the behavior (please share code or minimal repo)

This is not working:

```typescript
        const stream1 = createReadStream(__dirname+'/test.json');
        const stream2 = new PassThrough();
        stream1.pipe(stream2);

        type PutObjectParams = ConstructorParameters<typeof PutObjectCommand>[0];
        const params: PutObjectParams = { Bucket: this.configAws.s3.mediaBucket, Key: filePath, Body: stream2, Expires: expiresAt };
        await this.s3.send(new PutObjectCommand(params));
```

But this is working:

```typescript
        const stream1 = createReadStream(__dirname+'/test.json');

        type PutObjectParams = ConstructorParameters<typeof PutObjectCommand>[0];
        const params: PutObjectParams = { Bucket: this.configAws.s3.mediaBucket, Key: filePath, Body: stream1, Expires: expiresAt };
        await this.s3.send(new PutObjectCommand(params));
```

**Expected behavior**
I should be able to pass the stream object fe Duplex, Transform or PassThrough 

**Screenshots**
None

**Additional context**

```
/home/cojack/Projects/foo/bar/node_modules/@aws-sdk/client-s3/protocols/Aws_restXml.ts:9275
  return Promise.reject(Object.assign(new Error(message), response));
                                      ^
NotImplemented: A header you provided implies functionality that is not implemented
    at deserializeAws_restXmlPutObjectCommandError (/home/cojack/Projects/foo/bar/node_modules/@aws-sdk/client-s3/protocols/Aws_restXml.ts:9275:39)
    at processTicksAndRejections (node:internal/process/task_queues:93:5)
    at /home/cojack/Projects/foo/bar/node_modules/@aws-sdk/middleware-serde/src/deserializerMiddleware.ts:20:18
    at /home/cojack/Projects/foo/bar/node_modules/@aws-sdk/middleware-signing/src/middleware.ts:26:22
[...]

PR #3571, opened 2022-04-27
fix(middleware-sdk-s3): improve error message on stream upload

kuhe:
### Issue
https://github.com/aws/aws-sdk-js-v3/issues/1920
https://github.com/aws/aws-sdk-js-v3/issues/2348

### Description
- add a warning if the Body of a PutObject request might be a Stream of unknown length.

### Testing
Unit tests added. Manual client testing in a linked workspace.


AllanZhengYP (commented):
Can you provide the rationale why we don't fail the request but log a warning message for future reference?

AllanZhengYP (approved):

kuhe:
from discussion: check for content length _only_ at this middleware stage. Throw error instead of only warning. Should be safe to assume the request will fail if it reaches this step without having determined the content length.

AllanZhengYP (changes requested):
The original server-side error is actually complaining of missing `Content-Length` header. So instead of validating the input type, we should validate the header.

SDK will calculate the content length injected the `Content-Length` header in [content length middleware](https://github.com/aws/aws-sdk-js-v3/blob/main/packages/middleware-content-length/src/index.ts). In this case, the SDK can still populate the content length if the input is stream and from the file system, hense the API request can succeed.

Regarding whether we should log warning or fail-fast, I think we can keep **warning instead of throw**. Because in browser, browser may populate the content-length header on behalf of users in some cases: [Content-Length is a forbidden header by browser](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Length). So fail-fast for customer may block some cases where SDK cannot calculate the content length but brower can.

AllanZhengYP (approved):
Approve again!

PR #3576, opened 2022-04-29
fix(types): make Logger methods optional

everett1992:
`{} as __Logger` is in this repository more than 300 times including the
smithy code gen package that will copy the pattern into every client it
generates.

Code that calls `logger.debug(...)` is wrong because `{}.debug(...)` throws an
error. The 'correct' call should be `logger.debug?.(...)`.

IMO the ideal fix would be to replace all `{} as __Logger` with
`NO_OP_LOGGER` but that can't address any of the generated clients.

Maybe another option would be to wrap the `logger` that goes thru
configuration with a proxy that responds to the types methods.

This commit simply marks the methods in the type as optional so people
calling `Logger` realize the method may be undefined.

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,23 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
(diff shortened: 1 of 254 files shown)


ajredniwja (approved):

kuhe:
I have this task in my queue. I would've accepted the PR too, but by the time I saw it it had no commits in it.

PR #3682, opened 2022-06-13
fix(shared-ini-file-loader): update ini parsing

kuhe:
### Description
Improve the ini file parsing behavior. This does not change the functionality, and is covered by existing unit tests.

### Testing
Existing unit tests.

related: https://github.com/aws/aws-sdk-js/pull/4125

### DIFF
--- a/packages/shared-ini-file-loader/src/parseIni.ts
+++ b/packages/shared-ini-file-loader/src/parseIni.ts
@@ -16,7 +16,10 @@ export const parseIni = (iniData: string): ParsedIniData => {
       }
     } else if (currentSection) {
       const indexOfEqualsSign = line.indexOf("=");
-      const isAssignment: boolean = indexOfEqualsSign !== -1 && indexOfEqualsSign > 0;
+      const start = 0;
+      const end: number = line.length - 1;
+      const isAssignment: boolean =
+        indexOfEqualsSign !== -1 && indexOfEqualsSign !== start && indexOfEqualsSign !== end;
       if (isAssignment) {
         const [name, value]: [string, string] = [
           line.substring(0, indexOfEqualsSign).trim(),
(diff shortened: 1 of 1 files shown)


packages/shared-ini-file-loader/src/parseIni.ts
```diff
@@ -15,10 +15,15 @@ export const parseIni = (iniData: string): ParsedIniData => {
...
     } else if (currentSection) {
-      const item = line.match(/^\s*(.+?)\s*=\s*(.+?)\s*$/);
-      if (item) {
+      const expressionComponents: string[] = line.split("=");
+      const isAssignment: boolean = expressionComponents.length >= 2 && !!expressionComponents[0];
+      if (isAssignment) {
+        const [name, value]: [string, string] = [
+          expressionComponents[0].trim(),
+          expressionComponents.slice(1).join("=").trim(),
+        ];
```
trivikr, line 24: Use [indexOf](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/indexOf) and [substring](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substring) instead of split?
```suggestion
      const indexOfEqualsSign = line.indexOf("=");
      if (indexOfEqualsSign !== -1) {
        const [name, value]: [string, string] = [
          line.substring(0, indexOfEqualsSign).trim(),
          line.substring(indexOfEqualsSign + 1).trim(),
        ];
```

trivikr (changes requested):

trivikr:
Remove regular expression for section too?

From:
```ts
    line = line.split(/(^|\s)[;#]/)[0]; // remove comments
    const section = line.match(/^\s*\[([^\[\]]+)]\s*$/);
    if (section) {
      currentSection = section[1];
```

To:
```ts
    line = line.split(/(^|\s)[;#]/)[0].trim(); // remove comments and trim
    if (line[0] === "[" && line[line.length - 1] === "]") {
      currentSection = line.substring(1, line.length - 1);
```

GitHub is not allowing me to post suggestion, as it's not part of the diff.

trivikr (approved):

PR #3719, opened 2022-06-20
fix(lib-dynamodb): fix marshalling of list attributes in UpdateCommand

brendanhonea:
### Description
I recently found a bug in the  `@aws-sdk/lib-dynamodb` package (document client) . The UpdateCommand was failing when providing a list attribute to AttributeUpdates. UpdateCommand allows you to supply native attribute values to update a dynamo record like so: 

```
new UpdateCommand({
    TableName: "SomeTable",
    Key: {
      pk: "somePk",
      sk: "someSk",
    },
    AttributeUpdates: {
        stringAttr: { Action: "PUT", Value: "Some string value" },
        numberAttr: { Action: "PUT", Value: 12345 },
        listAttr: { Action: "PUT", Value: ["list item 1", "list item 2"] },
    },
    ReturnValues: "ALL_NEW",
})
```

The issue I was running in to was that list attributes were not being marshaled properly and causing a JSON parsing error downstream in the library. The marshaler would convert the above AttributeUpdates map into:

```
AttributeUpdates: {
        stringAttr: { Action: "PUT", Value: {S: "some string value"} },
        numberAttr: { Action: "PUT", Value: { N: "12345"} },
        listAttr: { Action: "PUT", Value: [ { S: "list item 1"} , { S: "list item 2" } ] }, // Here is the problem
    },
```

The dynamodb client expects lists to also be prefaced with an attribute type like so 

```
AttributeUpdates: {
        stringAttr: { Action: "PUT", Value: {S: "some string value"} },
        numberAttr: { Action: "PUT", Value: { N: "12345"} },
        listAttr: { Action: "PUT", Value: { L: [ { S: "list item 1"} , { S: "list item 2" } ] } }, // This is what dynamodb needs to see
    },
```

This PR addresses the issue in the lib-dynamo marshaling code so that lists are passed properly to the dynamodb client. I added an optional parameter to the KeyNode type so that this only affects commands that need to process nodes in this fashion.

### Testing
I wrote unit tests to ensure that the marshaling works properly and ran the entire test suite to test for breaking changes.

### Additional Context

I could not regenerate clients due to `Projection ec2.2016-11-15 failed: java.lang.OutOfMemoryError: Java heap space`. I don't believe this was caused by my changes as I did not touch any EC2 code.
---

### DIFF
--- a/packages/util-dynamodb/src/marshall.ts
+++ b/packages/util-dynamodb/src/marshall.ts
@@ -34,8 +34,8 @@ export function marshall(data: Set<NativeAttributeBinary>, options?: marshallOpt
 export function marshall<M extends { [K in keyof M]: NativeAttributeValue }>(
   data: M,
   options?: marshallOptions
-): Record<string, AttributeValue>;
-export function marshall<L extends NativeAttributeValue[]>(data: L, options?: marshallOptions): AttributeValue[];
+): AttributeValue.MMember;
+export function marshall<L extends NativeAttributeValue[]>(data: L, options?: marshallOptions): AttributeValue.LMember;
 export function marshall(data: string, options?: marshallOptions): AttributeValue.SMember;
 export function marshall(data: number, options?: marshallOptions): AttributeValue.NMember;
 export function marshall(data: NativeAttributeBinary, options?: marshallOptions): AttributeValue.BMember;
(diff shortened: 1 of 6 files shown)


packages/util-dynamodb/src/marshall.ts
```diff
@@ -34,31 +34,14 @@ export function marshall(data: Set<NativeAttributeBinary>, options?: marshallOpt
...
-    case "S":
-    case "N":
-    case "B":
-    case "NULL":
-    case "BOOL":
-    case "$unknown":
-    default:
-      return attributeValue;
-  }
+  return convertToAttr(data, options);
```
kuhe, line 46: discussed it with Brendan -- the main issue is that the interface of `marshall`, this function, should not be changed since I have seen that it is used independently.

The option could be implemented as a new marshallOptions entry and applied only to UpdateCommand.

kuhe:
superceded by https://github.com/aws/aws-sdk-js-v3/pull/5306

Issue #3095, opened 2021-12-14
MiddlewareStacks added to commands in lib-dynamodb never get applied.


c0pp3rt0p:
## Describe the bug
Commands in lib-dynamodb do not add their `middelwareStack` to the underlying command.

Adding any middleware stack to a command in the lib-dynamodb package has no effect.

## Your environment

NodeJs/TS on mac

#### SDK version number
`@aws-sdk/package-name@3.43.0`

#### Is the issue in the browser/Node.js/ReactNative?
no

#### Details of the browser/Node.js/ReactNative version
$ node -v
v14.17.0

## Steps to reproduce
Please share code or minimal repo, and steps to reproduce the behavior.
```js
import { QueryCommand } from '@aws-sdk/lib-dynamodb';
import { dynamoDocClient } from './dynamodb-client';

        const cmd = new QueryCommand({
          TableName: "MyTable",
          KeyConditionExpression: "id = :id",
          ExpressionAttributeValues: {
            ':id': id,
          },
        });
        cmd.middlewareStack.add(
          (next) => async (args) => {
            console.log('You will never see this');
            const result = await next(args);
            // result.response contains data returned from next middleware.
            return result;
          },
          {
            step: 'build',
            name: 'addFooMetadataMiddleware',
            tags: ['METADATA', 'FOO'],
          }
        );
      const dynamoDocClient = await client.send(cmd);
      // 
```
#### Observed behavior
The expected console output "You will never see this", is never shown.

#### Expected behavior
when the command is sent "You will never see this" should be output to the console.

## Additional context
It appears the commands in lib-dynamodb are missing the step to concat the current middleware with the newly constructed command.  Only the client middlewares get resolved.  See comments in the code below.

link to relevant code: https://github.com/aws/aws-sdk-js-v3/blob/e35f78c97fa6710ff9c444351893f0f06755e771/lib/lib-dynamodb/src/commands/QueryCommand.ts#L83

```
resolveMiddleware(clientStack, configuration, options) {
        const { marshallOptions, unmarshallOptions } = configuration.translateConfig || {};
        const command = new client_dynamodb_1.QueryCommand(utils_1.marshallInput(this.input, this.inputKeyNodes, marshallOptions));
       // need to concat the current middlewareStack with the constructed client.
        const stack = clientStack.concat(this.middlewareStack);
       // resolve the newly combined stack
        const handler = command.resolveMiddleware(stack, configuration, options);
        return async () => {
[...]

PR #3756, opened 2022-06-28
fix(lib-dynamodb): pass lib Command middleware to client Commands

kuhe:
### Issue
https://github.com/aws/aws-sdk-js-v3/issues/3095

### Description
- adds `.clientMiddlewareStack` to lib-dynamodb commands. This stack will be applied to the underlying client middleware as it is initialized.
- `.middlewareStack` from extending smithy `Command` is not usable because it implies IO types that will not actually appear in the middleware at runtime (the marshalled types). 

### Testing
- [x] add unit test

### DIFF
--- a/lib/lib-dynamodb/src/commands/QueryCommand.ts
+++ b/lib/lib-dynamodb/src/commands/QueryCommand.ts
@@ -5,11 +5,10 @@ import {
   QueryCommandInput as __QueryCommandInput,
   QueryCommandOutput as __QueryCommandOutput,
 } from "@aws-sdk/client-dynamodb";
-import { Command as $Command } from "@aws-sdk/smithy-client";
 import { Handler, HttpHandlerOptions as __HttpHandlerOptions, MiddlewareStack } from "@aws-sdk/types";
 import { NativeAttributeValue } from "@aws-sdk/util-dynamodb";
 
-import { marshallInput, unmarshallOutput } from "../commands/utils";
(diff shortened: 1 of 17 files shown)


lib/lib-dynamodb/src/commands/QueryCommand.ts
```diff
@@ -79,16 +78,11 @@ export class QueryCommand extends $Command<
     configuration: DynamoDBDocumentClientResolvedConfig,
     options?: __HttpHandlerOptions
   ): Handler<QueryCommandInput, QueryCommandOutput> {
-    const { marshallOptions, unmarshallOptions } = configuration.translateConfig || {};
-    const command = new __QueryCommand(marshallInput(this.input, this.inputKeyNodes, marshallOptions));
-    const handler = command.resolveMiddleware(clientStack, configuration, options);
+    const command = new __QueryCommand(this.input);
```
kuhe, line 81: add ts-ignore for future possibility of a command with non-optional fields

kuhe:
discussed with team in a review document, here's what we want to do with the PR:

- ensure no potential for type errors in build. The current polymorphism may only be getting through because all the I/O types have optional fields, or due to some level of indirection obscuring the type shenanigans 
- create new PR to supercede this, leaving it as a record
- initialize inner command in ctor, but not public
- make LibCommand.mwStack simply passthrough ref to the inner command mwStack

PR #3763, opened 2022-06-29
Regenerate clients with aws-models sensitive updates

gosar:
Just a draft PR to show the impact of changing a few models in aws-models - https://github.com/aws/aws-models/pull/791 for upcoming Smithy release. Not intended to be committed.

I just copied the aws-models into aws-sdk-js-v3/codegen/sdk-codegen/aws-models as-is (as I couldn't get the `yarn copy-models` to work). Use 'hide whitespaces' in the diff to see the real diff. There are a few unrelated diffs because of some formatting though. This PR shows the updated models and the output of `yarn generate-clients && yarn test:protocols`.

Analysis of the diffs (please review):
* protocol test diffs are because of our set->unique items change and other new protocol tests added
* kafkaconnect/Aws_restJson1 are changes because a shape name has changed, but for map or string. causing only non-exported serde methods to change
* mwaa/models.ts updated some filterSensitiveLog methods, because some shapes are now sensitive, but weren’t before.

---


gosar:
https://github.com/aws/aws-models/pull/791 is merged. closing this one.

PR #3793, opened 2022-07-11
chore(clients): replace triple-equals undefined+null check with double-equals

trivikr:
### Issue
N/A

### Description
Replaces triple-equals undefined+null check with double-equals for null

From [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Equality_comparisons_and_sameness):
> `undefined` and `null` are loosely equal; that is, `undefined == null` is true, and `null == undefined` is true.

We already use loose equality between null and undefined in https://github.com/aws/aws-sdk-js-v3/blob/ef017be91b1dc02446915806aa71cbcc37d82d78/packages/signature-v4/src/getPayloadHash.ts#L20-L22

### Testing
CI

The client size is reduced by `0.65%`

<details>
<summary>Before</summary>

```console
$ client-s3> du -s dist-* 
1228    dist-cjs
1340    dist-es
1736    dist-types
```

</details>

<details>
<summary>After</summary>

```console
$ client-s3> du -s dist-*
1220    dist-cjs
1332    dist-es
1736    dist-types
```

</details>

---


AllanZhengYP (approved):
Small change brings a lot improment! ⛴

kuhe:
Should be able to use `!= null` instead of `undefined` to save a few letters.

kuhe (approved):

Issue #3872, opened 2022-08-22
Construct Upload object with controlled AbortController


addy:
### Describe the feature

Today the lib-storage `Upload` object constructs an `AbortController` as a private member variable [here](https://github.com/aws/aws-sdk-js-v3/blob/22ede5ab4ca856cbf8ab17a8e48c5f417b6dd291/lib/lib-storage/src/Upload.ts#L80).

This only seems to be accessible by the `Upload.abort()` function [here](https://github.com/aws/aws-sdk-js-v3/blob/22ede5ab4ca856cbf8ab17a8e48c5f417b6dd291/lib/lib-storage/src/Upload.ts#L83-L89).

I'm interested in hearing thoughts on allowing an external `AbortController` to be passed through the `Options` interface.

### Use Case

In my specific case, I've wrapped a subset of S3 functionality in my own service that handles the construction and execution of an `Upload` object. In this way, the caller only has to worry about initiating the upload call and waiting for its success or failure.

In a multipart file upload scenario that is purely streamed (no buffered memory overhead), a good way to use the above wrapper is to push each upload Promise to an array and await all of them once the multipart stream is finished. If, however, there is an issue uploading one of the multiple files in the stream to S3, there's not exactly a clean way of cancelling other files in flight and rolling back. This is because the `abort()` function is tied to a purely internal `AbortController` and `AbortSignal`.

### Proposed Solution

If I had more control over the `Upload` object (i.e. passing my own `AbortController` to the wrapper service) then this approach would feel simpler/cleaner.

### Acknowledgements

- [X] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### SDK version used

3.142.0

### Environment details (OS name and version, etc.)

macOS Monterey 12.5.1

ajredniwja:
Hey @addy thanks for opening the the feature request, the team would access the priority of it, you are welcome to open a PR and implement it if its not a breaking change. Keeping the thread open for more discussion.

addy:
> you are welcome to open a PR and implement it if its not a breaking change.

I'd be more than happy to submit a PR, I don't think it should be a very large change.

PR #3873, opened 2022-08-22
feat(upload-abort-controller): add optional abortController param for Upload

addy:
### Issue
#3872

### Description
This optionally exposes the abortController on the `Upload` object.

### Testing
Wrote new test in Jest

### Additional context
_N/A_

---

### DIFF
--- a/lib/lib-storage/src/types.ts
+++ b/lib/lib-storage/src/types.ts
@@ -1,4 +1,5 @@
 import { PutObjectCommandInput, S3Client, Tag } from "@aws-sdk/client-s3";
+import { AbortController } from "@aws-sdk/abort-controller";
 
 export interface Progress {
   loaded?: number;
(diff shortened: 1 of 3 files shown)


lib/lib-storage/src/types.ts
```diff
@@ -1,4 +1,5 @@
 import { PutObjectCommandInput, S3Client, Tag } from "@aws-sdk/client-s3";
+import { AbortController } from "@aws-sdk/abort-controller";
```
kuhe, line 2: add this as a devDep in package.json?

addy: It's actually already there:

https://github.com/aws/aws-sdk-js-v3/blob/1098e30464355ffcd1a02166f0d748e646953eb8/lib/lib-storage/package.json#L37

kuhe:
this is limited enough in scope to not require a rebase for status checks

kuhe (approved):

Issue #3856, opened 2022-08-14
Failed requests always return 400 as the response code


vlovich:
### Checkboxes for prior research

- [X] I've gone through [Developer Guide](https://docs.aws.amazon.com/sdk-for-javascript/v3/developer-guide) and [API reference](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest)
- [X] I've checked [AWS Forums](https://forums.aws.amazon.com) and [StackOverflow](https://stackoverflow.com/questions/tagged/aws-sdk-js).
- [X] I've searched for [previous similar issues](https://github.com/aws/aws-sdk-js-v3/issues) and didn't find any solution.

### Describe the bug

CopyObject with a non-existent source bucket is defined as returning a 404 NoSuchBucket. However, the S3 V3 SDK returns 400 NoSuchBucket. I've confirmed that both Boto3 v1.24.44 and AWS JS SDK v2 propagate the status code correctly.

I've tracked it down to this:
```
    if (statusCode < 200 && statusCode >= 300)
        return result;
```
in `@aws-sdk/middleware-sdk-s3/dist-cjs/throw-200-exceptions.js`

I'm guessing it's supposed to be `||`. This impacts more than just CopyObject obviously - all requests that return an error always set the status code erroneously to 400.

### SDK version number

@aws-sdk/client-s3@3.145.0

### Which JavaScript Runtime is this issue in?

Node.js

### Details of the browser/Node.js/ReactNative version

v16.11.1

### Reproduction Steps

```
      await s3.copyObject({
        Bucket: 'non-existent-bucket',
        Key: 'testObject',
        CopySource: `/mybucket/nonExistentObject`,
      })
```

### Observed Behavior

Returned `httpStatusCode` in the returned `$metadata` is 400.

### Expected Behavior

Returned `httpStatusCode` in the returned `$metadata` should be 400.

### Possible Solution

Fix erroneous conditional. Status code cannot be both < 200 and >= 300, It should be `< 200 || >= 300`. I've confirmed that fixing the conditional fixes the issue & the returned status code is correct.

yenfryherrerafeliz:
[...]

PR #3874, opened 2022-08-22
fix(middleware-sdk-s3): process errors with 200 status code

siddsriv:
### Issue
Fixes: #3856

### Description
Process errors with 200 status code in throw200ExceptionsMiddleware

### Testing
Unit testing

---

### DIFF
--- a/packages/middleware-sdk-s3/src/throw-200-exceptions.spec.ts
+++ b/packages/middleware-sdk-s3/src/throw-200-exceptions.spec.ts
@@ -15,6 +15,54 @@ describe("throw200ExceptionsMiddlewareOptions", () => {
     jest.clearAllMocks();
   });
 
+  describe("exceptions for code < 200 and >= 300", () => {
+    mockStreamCollector.mockResolvedValue(Buffer.from(""));
+    mockUtf8Encoder.mockReturnValue("");
+    it("throws an exception if code is less than 200", async () => {
+      mockNextHandler.mockReturnValue({
+        response: new HttpResponse({
+          statusCode: 199,
+          headers: {},
+          body: "",
+        }),
+      });
+      const handler = throw200ExceptionsMiddleware(mockConfig)(mockNextHandler, {} as any);
+      try {
+        await handler({
+          input: {},
+          request: new HttpRequest({
+            hostname: "s3.us-east-1.amazonaws.com",
+          }),
+        });
+      } catch (e) {
+        expect(e).toBeDefined();
+        expect(e.name).toEqual("InternalError");
+        expect(e.message).toEqual("S3 aborted request");
+      }
+    });
+    it("throws an exception if code is greater than or equal to 300", async () => {
+      mockNextHandler.mockReturnValue({
+        response: new HttpResponse({
+          statusCode: 300,
+          headers: {},
+          body: "",
+        }),
+      });
+      const handler = throw200ExceptionsMiddleware(mockConfig)(mockNextHandler, {} as any);
+      try {
+        await handler({
+          input: {},
+          request: new HttpRequest({
+            hostname: "s3.us-east-1.amazonaws.com",
+          }),
+        });
+      } catch (e) {
+        expect(e).toBeDefined();
+        expect(e.name).toEqual("InternalError");
+        expect(e.message).toEqual("S3 aborted request");
+      }
+    });
+  });
   it("should throw if response body is empty", async () => {
     expect.assertions(3);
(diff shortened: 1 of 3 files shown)


trivikr:
The failing tests are integration tests in S3. It looks like there's some issue with mocking in those tests.

```console
  1) Throw 200 response
       should throw if UploadPartCopy() return with 200 and empty payload:

      AssertionError: expected promise to be rejected with an error including 'S3 aborted request' but got 'UnknownError'
      + expected - actual

      -UnknownError
      +S3 aborted request
      
  

  2) Throw 200 response
       should throw if CompleteMultipartUpload() return with 200 and empty payload:

      AssertionError: expected promise to be rejected with an error including 'S3 aborted request' but got 'UnknownError'
      + expected - actual

      -UnknownError
      +S3 aborted request
      
```

packages/middleware-sdk-s3/src/throw-200-exceptions.spec.ts
```diff
@@ -15,6 +15,54 @@ describe("throw200ExceptionsMiddlewareOptions", () => {
     jest.clearAllMocks();
   });
 
+  describe("exceptions for code < 200 and >= 300", () => {
```
trivikr, line 18: nit: be explicit
```suggestion
  describe("exceptions for statusCode < 200 and >= 300", () => {
```

packages/middleware-sdk-s3/src/throw-200-exceptions.spec.ts
```diff
@@ -15,6 +15,54 @@ describe("throw200ExceptionsMiddlewareOptions", () => {
...
+    mockStreamCollector.mockResolvedValue(Buffer.from(""));
+    mockUtf8Encoder.mockReturnValue("");
+    it("throws an exception if code is less than 200", async () => {
```
trivikr, line 21: Lot of code is shared between the two tests, where only statusCode is different.
You can use [it.each](https://jestjs.io/docs/api#testeachtablename-fn-timeout) for it as follows:

```js
it.each([199, 300])("throws for status code %d", async (statusCode) => {
```

packages/middleware-sdk-s3/src/throw-200-exceptions.spec.ts
```diff
@@ -15,6 +15,54 @@ describe("throw200ExceptionsMiddlewareOptions", () => {
...
+          request: new HttpRequest({
+            hostname: "s3.us-east-1.amazonaws.com",
+          }),
+        });
+      } catch (e) {
+        expect(e).toBeDefined();
+        expect(e.name).toEqual("InternalError");
+        expect(e.message).toEqual("S3 aborted request");
+      }
+    });
```
trivikr, line 42: The middleware shouldn't throw an error when status code is <200 or >=300.

This code should be:
```suggestion
    const result = await handler({
        input: {},
        request: new HttpRequest({
          hostname: "s3.us-east-1.amazonaws.com",
        }),
      });
    expect(result.response).toBe(mockResponse);
```

Where:

```js
      mockNextHandler.mockReturnValue({
        response: mockResponse,
      });
```

trivikr (changes requested):

siddsriv:
(force) push above -- that refactors the tests. Needs more work before PR.

yarn.lock
```diff
@@ -2080,9 +2080,9 @@
     universal-user-agent "^6.0.0"
 
 "@octokit/endpoint@^6.0.1":
```
trivikr, line 2082: Undo changes in yarn.lock, as this file doesn't need to be edited.

packages/middleware-sdk-s3/src/throw-200-exceptions.spec.ts
```diff
@@ -15,58 +16,81 @@ describe("throw200ExceptionsMiddlewareOptions", () => {
...
-    const handler = throw200ExceptionsMiddleware(mockConfig)(mockNextHandler, {} as any);
-    try {
-      await handler({
+
+    it.each([199, 300])("results for statusCode %i", async (statusCode) => {
+      mockNextHandler.mockReturnValue({
+        response: mockResponse,
+        statusCode,
+        // headers: {},
+        // body: "",
```
trivikr, line 28: Uncomment, and pass body with some content?
```suggestion
        headers: {},
        body: "body",
```

trivikr (changes requested):

AllanZhengYP (approved):

kuhe (approved):

trivikr (approved):

PR #3880, opened 2022-08-24
feat(util-endpoints): add ruleSet standard library

trivikr:
### Issue
Internal JS-3527

### Description
Adds ruleSet standard library for endpoint resolution, except the AWS functions

### Testing
Unit testing

---

### DIFF
--- a/packages/util-endpoints/src/index.ts
+++ b/packages/util-endpoints/src/index.ts
@@ -1,0 +1,1 @@
+export * from "./lib/";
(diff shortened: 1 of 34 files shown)


AllanZhengYP (approved):

packages/util-endpoints/src/index.ts
```diff
@@ -0,0 +1,2 @@
+export * from "./EndpointError";
+export * from "./lib/";
```
trivikr, line 2: lib doesn't have to be exported, as it would be used only by endpoint resolver which would be in the same package.
```suggestion
```

AllanZhengYP (approved):

Issue #3894, opened 2022-09-01
totalRetryDelay observed is less than retryAfterSeconds


sivantcy:
### Checkboxes for prior research

- [X] I've gone through [Developer Guide](https://docs.aws.amazon.com/sdk-for-javascript/v3/developer-guide) and [API reference](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest)
- [X] I've checked [AWS Forums](https://forums.aws.amazon.com) and [StackOverflow](https://stackoverflow.com/questions/tagged/aws-sdk-js).
- [X] I've searched for [previous similar issues](https://github.com/aws/aws-sdk-js-v3/issues) and didn't find any solution.

### Describe the bug

I'm using Smithy and the retryable trait with @httpHeader("Retry-After") and retryAfterSeconds. in the service I'm returning header with Retry-After = 2. I have ts client that generated directly thought Smithy. in the logs the value of totalRetryDelay is around 50. how does it possible that totalRetryDelay value is less than the value in Retry-After after 3 attempts?

### SDK version number

@aws-sdk/package-name@version, ...

### Which JavaScript Runtime is this issue in?

Node.js

### Details of the browser/Node.js/ReactNative version

Node.js 14

### Reproduction Steps

1. create a service that returns error and header with the header  Retry-After = "2"
2. create a smithy model with error that look like that: 
@error("server")
@retryable
@httpError(500)
structure InternalServerException {
    message: String,
    @httpHeader("Retry-After")
    retryAfterSeconds: Integer //To be retried by client after configured seconds.
}
3. generate ts client directly
4. throw this error from the service
5. check the logs that uses the generated ts client.
6. check that attempts are equal to 3 and the value in totalRetryDelay
 

### Observed Behavior

the attempts value was 3, but totalRetryDelay was around 5

### Expected Behavior

attempts=3 and totalRetryDelay around 4000

trivikr:
During deep dive, we noticed that `retryAfterSeconds` value is populated in error, but it's not consumed by retry middleware.

Example model:
https://github.com/aws/aws-sdk-js-v3/blob/cdccc0d68505085288cfa1db8c6f3bea86035b53/codegen/sdk-codegen/aws-models/mobile.json#L702-L720

Example generated code:
https://github.com/aws/aws-sdk-js-v3/blob/51c24224fe8622d011775dd89c6255a7edd29470/clients/client-mobile/src/protocols/Aws_restJson1.ts#L830-L846

[...]

PR #3916, opened 2022-09-09
fix(middleware-retry): use delay from response header if it's higher

trivikr:
### Issue
Fixes: https://github.com/aws/aws-sdk-js-v3/issues/3894

### Description
Checks error response for `Retry-Header`, and uses that value for delay if it's present and higher that delay decided by Standard Retry Strategy.

### Testing
ToDo

### Additional context
MDN: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Retry-After

---

### DIFF
--- a/packages/middleware-retry/src/StandardRetryStrategy.ts
+++ b/packages/middleware-retry/src/StandardRetryStrategy.ts
@@ -127,9 +127,9 @@ export class StandardRetryStrategy implements RetryStrategy {
 const getDelayFromRetryAfterHeader = (response: unknown): number | undefined => {
   if (!HttpResponse.isInstance(response)) return;
 
-  const retryAfter = response.headers["Retry-After"];
+  const retryAfter = response.headers[Object.keys(response.headers).find((key) => key.toLowerCase() === "retry-after")];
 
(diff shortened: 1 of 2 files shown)


packages/middleware-retry/src/StandardRetryStrategy.ts
```diff
@@ -95,10 +95,15 @@ export class StandardRetryStrategy implements RetryStrategy {
...
           retryTokenAmount = this.retryQuota.retrieveRetryTokens(err);
-          const delay = this.delayDecider(
+          const delayFromDecider = this.delayDecider(
             isThrottlingError(err) ? THROTTLING_RETRY_DELAY_BASE : DEFAULT_RETRY_DELAY_BASE,
             attempts
           );
+
+          const delayFromResponse = getDelayFromRetryAfterHeader(err.$response);
+          const delay =
+            delayFromResponse && delayFromResponse > delayFromDecider ? delayFromResponse : delayFromDecider;
```
trivikr, line 105:
```suggestion
          const delay = Math.max(delayFromResponse || 0, delayFromDecider);
```

kuhe (approved):

kuhe (approved):

PR #3917, opened 2022-09-09
fix(e2e): fix "eventually" helper in e2e tests

kuhe:
### Description
Fix the `eventually` helper function.

### Testing
```
npx cucumber-js --fail-fast -t @queues --format json:cucumber-report.json
```

### DIFF
--- a/features/extra/helpers.js
+++ b/features/extra/helpers.js
@@ -40,16 +40,15 @@ module.exports = {
     if (!options.backoff) options.backoff = 500;
     if (!options.maxTime) options.maxTime = 5;
 
-    const delay = options.delay;
-    //TODO: apply clock offset
-    const started = new Date();
+    let delay = options.delay;
+    const started = Date.now();
 
     const self = this;
     const retry = function () {
       callback();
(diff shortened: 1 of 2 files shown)


features/extra/helpers.js
```diff
@@ -40,16 +40,15 @@ module.exports = {
     if (!options.backoff) options.backoff = 500;
     if (!options.maxTime) options.maxTime = 5;
 
-    const delay = options.delay;
```
kuhe, line 43: This function has likely been broken for a very long time. `delay += options.backoff;` could never have worked with `const delay`. It's likely the `AWS.util.date.getDate()` call did not return a reasonable value and the retry did not happen, meaning previously passing e2e tests always succeeded on the first try of the `next.condition`.

trivikr: The `AWS.util.date.getDate()` is a remnant from AWS SDK for JavaScript (v2), which wouldn't have worked in v3.
Thanks for catching it!

trivikr (approved):

PR #3918, opened 2022-09-09
fix(e2e): clean up stale sqs queues before running sqs tests

kuhe:
### Testing
npx cucumber-js --fail-fast -t @sqs


trivikr (changes requested):
A better solution should be to perform cleanup under After or AfterEach call.

I've posted a PR to bump cucumber to latest version in https://github.com/aws/aws-sdk-js-v3/pull/3946, we can take up the cleanup after it's merged.

trivikr:
Replaced by https://github.com/aws/aws-sdk-js-v3/pull/3948

PR #3926, opened 2022-09-12
test(util-endpoints): add integration tests for resolving endpoint

trivikr:
### Issue
Internal JS-3566

### Description
Add integration tests for resolving endpoint

### Testing
Integration testing

```console
$ aws-sdk-js-v3> yarn test:integration
...
Test Suites: 3 passed, 3 total
Tests:       77 passed, 77 total
Snapshots:   0 total
Time:        8.435 s
Ran all test suites in 3 projects.
Done in 9.57s.

$ util-endpoints> yarn test src/resolveEndpoint.integ.spec.ts --config jest.config.integ.js
...
Test Suites: 1 passed, 1 total
Tests:       75 passed, 75 total
Snapshots:   0 total
Time:        3.278 s, estimated 4 s
Ran all test suites matching /src\/resolveEndpoint.integ.spec.ts/i.
Done in 4.19s.
```

---
By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #3909, opened 2022-09-07
chore(util-endpoints): add endpoint ruleset library specific to aws

trivikr:
### Issue
Internal JS-3541

### Description
Adds endpoint ruleset library specific to aws

### Testing
Unit testing

---
By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #3929, opened 2022-09-12
fix(util-endpoints): call multi-level functions from callFunction

trivikr:
### Issue
Noticed during integration tests in https://github.com/aws/aws-sdk-js-v3/pull/3926

### Description
The endpoint ruleset library specific to aws was added in https://github.com/aws/aws-sdk-js-v3/pull/3909

However, the callFunction method was not updated to call multi-level functions, i.e. call lib.aws.partition for function name "aws.partition"

This PR makes that change.

### Testing
Unit testing

---

### DIFF
--- a/packages/util-endpoints/src/utils/callFunction.ts
+++ b/packages/util-endpoints/src/utils/callFunction.ts
@@ -5,5 +5,5 @@ import { evaluateExpression } from "./evaluateExpression";
 export const callFunction = ({ fn, argv }: FunctionObject, options: EvaluateOptions): FunctionReturn => {
   const argvArray = argv.map((arg) => (typeof arg === "boolean" ? arg : evaluateExpression(arg, "arg", options)));
   // @ts-ignore Element implicitly has an 'any' type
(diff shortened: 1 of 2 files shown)


packages/util-endpoints/src/utils/callFunction.ts
```diff
@@ -5,5 +5,5 @@ import { evaluateExpression } from "./evaluateExpression";
 export const callFunction = ({ fn, argv }: FunctionObject, options: EvaluateOptions): FunctionReturn => {
   const argvArray = argv.map((arg) => (typeof arg === "boolean" ? arg : evaluateExpression(arg, "arg", options)));
   // @ts-ignore Element implicitly has an 'any' type
-  return lib[fn](...argvArray);
+  return fn.split(".").reduce((acc, key) => acc[key], lib)(...argvArray);
```
kuhe, line 8: minor naming issue: argvArray is "argument vector array", could instead be something like "evaluatedArguments"

kuhe (approved):

PR #3935, opened 2022-09-12
fix(util-endpoints): return null in parseArn for some empty elements

trivikr:
### Issue
Noticed during integration testing in https://github.com/aws/aws-sdk-js-v3/pull/3926

### Description
Return null in parseArn when partition, service or resourceId is empty

### Testing
Unit testing

---

### DIFF
--- a/packages/util-endpoints/src/lib/aws/parseArn.ts
+++ b/packages/util-endpoints/src/lib/aws/parseArn.ts
@@ -8,24 +8,11 @@ import { EndpointARN } from "@aws-sdk/types";
 export const parseArn = (value: string): EndpointARN | null => {
   const segments = value.split(":");
 
-  if (
-    segments.length < 6 ||
-    segments[0] !== "arn" ||
-    segments[1] === "" || // Partition is required.
-    segments[2] === "" || // Service is required.
-    segments[5] === "" // Resource ID is required.
-  )
-    return null;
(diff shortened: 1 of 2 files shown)


packages/util-endpoints/src/lib/aws/parseArn.ts
```diff
@@ -8,7 +8,14 @@ import { EndpointARN } from "@aws-sdk/types";
...
 
-  if (segments.length < 6 || segments[0] !== "arn") return null;
+  if (
+    segments.length < 6 ||
+    segments[0] !== "arn" ||
+    segments[1] === "" || // Partition is required.
+    segments[2] === "" || // Service is required.
+    segments[5] === "" // Resource ID is required.
+  )
+    return null;
```
kuhe, line 18: this could go below the named segmentation so you can check 

```
if (id !== "arn" || partition === "" || service === "" || resourceId[0] === "") ...
```

kuhe (approved):

packages/util-endpoints/src/lib/aws/parseArn.ts
trivikr: Added in [`e0195d4` (#3935)](https://github.com/aws/aws-sdk-js-v3/pull/3935/commits/e0195d4acc18202773b477df8762b8740280a7f3)

PR #3963, opened 2022-09-16
chore(clients): regenerate clients with DirectedCodegen

gosar:
Based on running `yarn generate-clients` on top of using smithy-typescript changes in https://github.com/awslabs/smithy-typescript/pull/585.

On 8/31 there were no diffs. On 9/15 when I rebase and regenerate, I do not expect any diff, but I'm seeing this diff, just in 1 client - identitystore. It is just moving couple of exceptions to a different location. I don't know why there is this diff and just for this one client - not sure what is special about it. Between 8/31-9/15 the [only changes](https://github.com/awslabs/smithy-typescript/compare/d14731f006c8d73881cab2bc947113e5918bf997..2159d71300e4922c59c468cdb5ccb71ae055fe1f) in smithy-typescript are rebasing and adding the EndpointV2 logic to DirectedTypeScriptCodegen (what used to be in CodegenVistor, which is deleted in favor of DirectedTypeScriptCodegen).

---


trivikr (approved):

trivikr:
I'm also not sure why the rearrangement is happening.
But verified that rearrangement happens in my local workspace too.

This PR looks good to me. If the situation affects more clients, we'll prioritize sorting the models.

PR #3972, opened 2022-09-21
feat(codegen): add query compatible header error code handling for JSON protocols

AndrewFossAWS:
### Description
* This commit adds query compatible header handling for JSON protocols only
* When an AWS service sends a `x-amzn-query-error` header, it contains an error code and type.
* Example header: `x-amzn-query-error:InternalFailure;Receiver`
  * `InternalFailure` is the error code
  * `Receiver` is the Type
* Note: The value is always a `;` delimited string e.g. `InternalFailure;Receiver`, `MalformedInput;Sender`

### Testing
* Generated clients using json protocol
* Verified code generation is correct

---


kuhe (approved):

codegen/smithy-aws-typescript-codegen/src/main/java/software/amazon/smithy/aws/typescript/codegen/AwsProtocolUtils.java
```diff
@@ -117,6 +118,15 @@ static void generateJsonParseBody(GenerationContext context) {
         writer.write("");
     }
 
+    static void generateJsonParseBodyWithQueryHeader(GenerationContext context) {
+        TypeScriptWriter writer = context.getWriter();
+        writer.addImport("HeaderBag", "__HeaderBag", "@aws-sdk/types");
+        writer.addImport("HttpResponse", "__HttpResponse", "@aws-sdk/protocol-http");
```
kstich, line 124: The `HttpResponse` and `SerdeContext` imports don't look used by the loaded stub, can they be removed?

kstich (approved):

PR #4002, opened 2022-09-29
fix(endpoint): misc fixes for endpoints 2.0 based on service unit tests

kuhe:
- misc fixes for endpoints 2.0 based on S3 service unit tests (not shown here).

I have not imported the test cases I used because they are not public yet. Added here are only some revealed fixes and the test case runner code. 

**REQUIRES**: https://github.com/awslabs/smithy-typescript/pull/600

### DIFF
--- a/packages/util-endpoints/src/utils/evaluateCondition.ts
+++ b/packages/util-endpoints/src/utils/evaluateCondition.ts
@@ -6,16 +6,8 @@ export const evaluateCondition = ({ assign, ...fnArgs }: ConditionObject, option
     throw new EndpointError(`'${assign}' is already defined in Reference Record.`);
   }
   const value = callFunction(fnArgs, options);
-  const output = {
+  return {
     result: value === "" ? true : !!value,
     ...(assign != null && { toAssign: { name: assign, value } }),
   };
-
-  // console.log("evaluateCondition", {
-  //   args: JSON.stringify(fnArgs),
-  //   references: options.referenceRecord,
-  //   output,
-  // });
-
-  return output;
 };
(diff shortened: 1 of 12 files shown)


packages/util-endpoints/src/utils/evaluateCondition.ts
```diff
@@ -6,8 +6,16 @@ export const evaluateCondition = ({ assign, ...fnArgs }: ConditionObject, option
...
     ...(assign != null && { toAssign: { name: assign, value } }),
   };
+
+  // console.log("evaluateCondition", {
+  //   args: JSON.stringify(fnArgs),
+  //   references: options.referenceRecord,
+  //   output,
+  // });
+
+  return output;
```
trivikr, line 20: Looks like this diff was added during debugging, and can be deleted.

packages/util-endpoints/src/utils/getEndpointUrl.ts
```diff
@@ -5,7 +5,12 @@ import { evaluateExpression } from "./evaluateExpression";
 export const getEndpointUrl = (endpointUrl: Expression, options: EvaluateOptions): URL => {
   const expression = evaluateExpression(endpointUrl, "Endpoint URL", options);
   if (typeof expression === "string") {
-    return new URL(expression);
+    try {
+      return new URL(expression);
+    } catch (error) {
+      console.error(`Failed to construct URL with ${expression}`, error);
+      throw error;
+    }
```
trivikr, line 13: I believe this diff is also added for debugging, and can be removed?

kuhe: I want to provide more detail here even for production use, because otherwise the error is too generic "Invalid URL".

packages/util-endpoints/src/utils/evaluateCondition.ts
kuhe: I wanted to leave this easily accessible, but I'll remove it. It's easy enough to add as needed.

trivikr (approved):

PR #4011, opened 2022-10-03
fix(endpoint-discovery): delete failed cache entry in blocking operations

kuhe:
In blocking operations where discovered endpoint is required, the `updateDiscoveredEndpointInCache` function must resolve or reject.

Because the function enters a waiting state if it discovers a cached but incomplete endpoint, it could incorrectly wait indefinitely if the incomplete endpoint is the result of a failed request.


packages/middleware-endpoint-discovery/src/updateDiscoveredEndpointInCache.ts
```diff
@@ -54,10 +54,11 @@ export const updateDiscoveredEndpointInCache = async (
...
         .catch((error: any) => {
-          if (error.name === "InvalidEndpointException" || error.$metadata?.httpStatusCode === 421) {
-            // Endpoint is invalid, delete the cache entry.
-            endpointCache.delete(cacheKey);
-          }
+          // The cache entry must be deleted 
+          // because a subsequent blocking request will be stuck
+          // in a waiting state if it sees the cache entry
+          // but we have already flushed the request queue. 
+          endpointCache.delete(cacheKey);
```
kuhe, line 61: This conditional was unnecessary. 

- the cache entry must be deleted in all error cases to unblock the next awaiting request
- the cache entry placeholder is set below for non-blocking requests anyway

siddsriv (approved):

PR #4012, opened 2022-10-03
feat(middleware-api-key): add new auth middleware

eduardomourar:
### Description

We will create a new middleware to handle API keys in the context of Smithy generated models (probably not applicable for AWS specific services). The code is based heavily on [this PR](https://github.com/awslabs/smithy-typescript/pull/473) from the Smithy TypeScript repository.

In order for it to work, we need to ensure the Bearer token middleware succeeds even when the `token` is not set. 

### Testing
Unit tests

### Additional context

---

### DIFF
--- a/packages/middleware-api-key/src/middleware.ts
+++ b/packages/middleware-api-key/src/middleware.ts
@@ -1,0 +1,76 @@
+import { HttpRequest } from "@aws-sdk/protocol-http";
+import {
+  FinalizeHandler,
+  FinalizeHandlerArguments,
+  FinalizeHandlerOutput,
+  FinalizeRequestMiddleware,
+  HandlerExecutionContext,
+  HttpAuthDefinition,
+  Pluggable,
+  RelativeMiddlewareOptions,
+} from "@aws-sdk/types";
+
+import { ApiKeyResolvedConfig } from "./configurations";
+
+/**
+ * Middleware to inject the API key into the HTTP request.
+ *
+ * The middleware will inject the client's configured API key into the
+ * request as defined by the `@httpApiKeyAuth` trait. If the trait says to
+ * put the API key into a named header, that header will be used, optionally
+ * prefixed with a scheme. If the trait says to put the API key into a named
+ * query parameter, that query parameter will be used.
+ *
+ * @param pluginConfig the client configuration. Includes the function that will return the API key value.
+ * @param middlewareConfig the plugin options (location of the parameter, name, and optional scheme)
+ * @returns a function that processes the HTTP request and passes it on to the next handler
+ */
+export const apiKeyMiddleware =
+  <Input extends object, Output extends object>(
+    pluginConfig: ApiKeyResolvedConfig,
+    middlewareConfig: HttpAuthDefinition
+  ): FinalizeRequestMiddleware<Input, Output> =>
+  (next: FinalizeHandler<Input, Output>, context: HandlerExecutionContext): FinalizeHandler<Input, Output> =>
+    async function (args: FinalizeHandlerArguments<Input>): Promise<FinalizeHandlerOutput<Output>> {
+      if (!HttpRequest.isInstance(args.request) || context.currentAuthConfig) return next(args);
+
+      const apiKey = pluginConfig.apiKey && (await pluginConfig.apiKey());
+
+      // This middleware will not be injected if the operation has the @optionalAuth trait.
+      // We don't know if we're the only auth middleware, so let the service deal with the
+      // absence of the API key (or let other middleware do its job).
+      if (!apiKey) {
+        context.currentAuthConfig = undefined;
+        return next(args);
+      }
+      context.currentAuthConfig = middlewareConfig;
+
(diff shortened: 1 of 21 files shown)


packages/middleware-api-key/src/middleware.ts
```diff
@@ -0,0 +1,85 @@
...
+            }),
+          },
+          query: {
+            ...args.request.query,
+            // Set the query parameter, even if it's already been set.
+            ...(middlewareConfig.in === "query" && { [middlewareConfig.name]: apiKey }),
+          },
+        },
+      });
+    };
```
kuhe, line 68: Can you write this traditionally? 

```js
if (middlewareConfig.in === 'header') {
  args.request.headers[middlewareConfig.name.toLowerCase()] = ... ? ... : ...
}
```

packages/middleware-token/src/tokenMiddleware.ts
```diff
@@ -21,12 +23,22 @@ export const tokenMiddleware =
...
+    if (token?.token) {
+      const authConfig: HttpHeaderAuthDefinition = {
+        in: "header",
+        name: "Authorization",
+        scheme: "Bearer",
+      };
+      context.currentAuthConfig = authConfig;
+      args.request.headers[authConfig.name] = `${authConfig.scheme} ${token.token}`;
+    } else {
+      context.currentAuthConfig = null;
```
kuhe, line 40: add this field explicitly to the `HandlerExecutionContext` interface, so that static analysis / IDEs can pick up all the usages of it.

kuhe (changes requested):

packages/middleware-api-key/src/middleware.ts
eduardomourar: yes, i can and it is done

kuhe (approved):

eduardomourar:
I have created another PR with formatting issues: https://github.com/aws/aws-sdk-js-v3/pull/4087

eduardomourar:
@trivikr and @kuhe , could we get this approved and merged, please?

packages/types/src/auth.ts
```diff
@@ -25,3 +25,37 @@ export interface AuthScheme {
   signingScope?: never;
   properties: Record<string, unknown>;
 }
+
+// As described in the Smithy documentation:
+// https://github.com/awslabs/smithy/blob/main/smithy-model/src/main/resources/software/amazon/smithy/model/loader/prelude.smithy
```
trivikr, line 30: Can you use the same interface as in Smithy?

```ts
export interface HttpApiKeyAuth {
  /**
   * Defines the name of the HTTP header or query string parameter
   * that contains the API key.
   */
  name: string;

  /**
   * Defines the location of where the key is serialized.
   */
  in: HttpApiKeyLocation;

  /**
   * Defines the security scheme to use on the `Authorization` header value.
   * This can only be set if the "in" property is set to {@link HttpApiKeyLocation.HEADER}.
   */
  scheme?: string,
}

export enum HttpApiKeyLocation {
  HEADER = "header",
  QUERY = "query"
}
```

trivikr (changes requested):

packages/types/src/auth.ts
eduardomourar: I have created the enum for the location, but for the rest we are using those interfaces for both [API Key](https://github.com/aws/aws-sdk-js-v3/blob/4bc9f53335423221f2f2183a0dab462a0564bf4f/packages/middleware-api-key/src/middleware.ts#L31) and [Bearer token](https://github.com/aws/aws-sdk-js-v3/blob/4bc9f53335423221f2f2183a0dab462a0564bf4f/packages/middleware-token/src/tokenMiddleware.ts#L33-L37) usage.

packages/types/src/auth.ts
```diff
@@ -25,3 +25,42 @@ export interface AuthScheme {
...
+   */
+  name: string;
+}
+
+export enum HttpAuthLocation {
+  HEADER = "header",
+  QUERY = "query",
+}
+
+export type HttpAuthDefinition = HttpHeaderAuthDefinition | HttpQueryAuthDefinition;
```
trivikr, line 66: Redundant interfaces can be removed
```suggestion
export interface HttpAuthDefinition {
  /**
   * Defines the location of where the key is serialized.
   */
  in: HttpApiKeyLocation;

  /**
   * Defines the name of the HTTP header or query string parameter
   * that contains the API key.
   */
  name: string;

  /**
   * Defines the security scheme to use on the `Authorization` header value.
   * This can only be set if the "in" property is set to {@link HttpApiKeyLocation.HEADER}.
   */
  scheme?: string;
}

export enum HttpApiKeyLocation {
  HEADER = "header",
  QUERY = "query",
}
```

trivikr (changes requested):

eduardomourar:
@trivikr , could this be re-approved and merged, please?

kuhe (approved):

trivikr (approved):

PR #4031, opened 2022-10-12
fix(endpoint): misc endpoints 2.0 fixes for s3

kuhe:
- additional fixes for endpoints 2.0 based on S3 tests

### DIFF
--- a/packages/middleware-endpoint/src/service-customizations/s3.ts
+++ b/packages/middleware-endpoint/src/service-customizations/s3.ts
@@ -3,8 +3,27 @@ import { EndpointParameters } from "@aws-sdk/types";
 export const resolveParamsForS3 = async (endpointParams: EndpointParameters) => {
   const bucket = (endpointParams?.Bucket as string) || "";
 
-  if (!isDnsCompatibleBucketName(bucket) || bucket.indexOf(".") !== -1) {
+  if (typeof endpointParams.Bucket === "string") {
+    endpointParams.Bucket = bucket.replace(/#/g, encodeURIComponent("#")).replace(/\?/g, encodeURIComponent("?"));
+  }
+
+  if (isArnBucketName(bucket)) {
+    if (endpointParams.ForcePathStyle === true) {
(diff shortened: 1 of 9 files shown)


packages/middleware-endpoint/src/service-customizations/s3.ts
```diff
@@ -26,3 +45,13 @@ export const S3_HOSTNAME_PATTERN = /^(.+\.)?s3(-fips)?(\.dualstack)?[.-]([a-z0-9
...
+
+export const isArnBucketName = (bucketName: string): boolean => {
+  const [arn, partition, service, region, account, typeOrId] = bucketName.split(":");
+  const isArn = arn === "arn" && bucketName.split(":").length >= 6;
+  const isValidArn = [arn, partition, service, account, typeOrId].filter(Boolean).length === 5;
+  if (isArn && !isValidArn) {
+    throw new Error(`Invalid ARN: ${bucketName} was an invalid ARN.`);
+  }
+  return arn === "arn" && !!partition && !!service && !!account && !!typeOrId;
+};
```
kuhe, line 57: I would expect that we already have such a function somewhere else, but I couldn't find one.

packages/util-endpoints/src/resolveEndpoint.ts
```diff
@@ -31,5 +31,19 @@ export const resolveEndpoint = (ruleSetObject: RuleSetObject, options: EndpointR
...
+    // take protocol and port from custom Endpoint if present.
+    try {
+      const givenEndpoint = new URL(options.endpointParams.Endpoint as string);
+      const { protocol, port } = givenEndpoint;
+      endpoint.url.protocol = protocol;
+      endpoint.url.port = port;
+    } catch (e) {
+      // ignored
+    }
+  }
```
kuhe, line 46: this could simply be a problem with the sample ruleset (not shown in this PR)

the resolved endpoint template starts with a hard-coded `https://{...` but a test case expects the protocol of the custom `endpointParams.Endpoint` to be respected.

trivikr (approved):

PR #4043, opened 2022-10-14
fix(endpoint): endpoints 2.0 all-service TS compilation fixes

kuhe:
- takes the endpoints-2.0 branch for all service models in aws-models repo and tests codegen + compilation
- fixes resulting TS compilation errors


kuhe:
this was paired with and requires https://github.com/awslabs/smithy-typescript/pull/614

trivikr (approved):

PR #4054, opened 2022-10-19
fix(endpoint): use sigv4a signing region override

kuhe:
when a sigv4a authscheme is reslved, set signingRegion using the `*` from `authScheme.signingRegionSet`, allowing S3's MultiRegionSigv4 signer to go into Sigv4a mode.

This fixes a bug uncovered by `S3.ispec.ts` browser mode.

### DIFF
--- a/packages/middleware-signing/src/middleware.ts
+++ b/packages/middleware-signing/src/middleware.ts
@@ -25,7 +25,7 @@ export const awsAuthMiddleware =
       // TODO(identityandauth): call authScheme resolver
       const authScheme: AuthScheme | undefined = context.endpointV2?.properties?.authSchemes?.[0];
       const multiRegionOverride: string | undefined =
-        authScheme?.name === "sigv4a" && authScheme?.signingRegionSet?.[0];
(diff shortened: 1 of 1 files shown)


trivikr (approved):

packages/middleware-signing/src/middleware.ts
```diff
@@ -24,14 +24,16 @@ export const awsAuthMiddleware =
 
       // TODO(identityandauth): call authScheme resolver
       const authScheme: AuthScheme | undefined = context.endpointV2?.properties?.authSchemes?.[0];
+      const multiRegionOverride: string | undefined =
+        authScheme?.name === "sigv4a" && authScheme?.signingRegionSet?.[0];
```
trivikr, line 28: Fix for error
```console
Type 'boolean' is not assignable to type 'string'.
src/middleware.ts(27,13): error TS2322: Type 'string | false | undefined' is not assignable to type 'string | undefined'.
```

```suggestion
      const multiRegionOverride: string | undefined =
        authScheme?.name === "sigv4a" ? authScheme?.signingRegionSet?.[0] : undefined;
```

PR #4087, opened 2022-10-25
chore: fix formatting with prettier

eduardomourar:
### Issue
N/A

### Description
Only fixing formatting on files using prettier.
* Prettier version: v2.7.1 (as specified in package.json)
* Command used: `npx prettier --write '**/*.{ts,js,md,json}'`

---

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -133,22 +133,24 @@ or [Smithy TypeScript generator][smithy typescript repo], you should include
 the generated code change to your PR. Here's how to generate clients:
 
 1. Clone and checkout smithy-typescript repo
-      ```
-      git clone https://github.com/awslabs/smithy-typescript.git && cd smithy-typescript
-      git checkout [working_branch]
-      ```
+
+   ```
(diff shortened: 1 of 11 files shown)


trivikr:
@eduardomourar Thanks for the PR!
Can you remove formatting from aws-models folder? We're in the middle of doing a batch update for clients.

Also, in the PR description can you add prettier version used, and the command run to perform formatting?
We perform auto formatting on file save and precommit. So, I'm curious how these were missed, and will look into it.

eduardomourar:
> @eduardomourar Thanks for the PR! Can you remove formatting from aws-models folder? We're in the middle of doing a batch update for clients.
> 
> Also, in the PR description can you add prettier version used, and the command run to perform formatting? We perform auto formatting on file save and precommit. So, I'm curious how these were missed, and will look into it.

Ideally, prettier/eslint should also run as part of the github workflows to avoid drift because precommit can be misconfigured or bypassed altogether.

trivikr (approved):

Issue #4095, opened 2022-10-26
Cannot connect to localhost with DynamoDBClient "EndpointError: Endpoint URL must be a string, got object"


allanvobraun:
### Checkboxes for prior research

- [X] I've gone through [Developer Guide](https://docs.aws.amazon.com/sdk-for-javascript/v3/developer-guide) and [API reference](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest)
- [X] I've checked [AWS Forums](https://forums.aws.amazon.com) and [StackOverflow](https://stackoverflow.com/questions/tagged/aws-sdk-js).
- [X] I've searched for [previous similar issues](https://github.com/aws/aws-sdk-js-v3/issues) and didn't find any solution.

### Describe the bug

With this configuration:
```
this.dynamoDB = DynamoDBClient({
      region: 'us-east-1',
      endpoint: 'http://localhost:5000/',
      credentials: {
        accessKeyId: 'xxx',
        secretAccessKey: 'yyy',
      },
    });
```
When executing a dynamodb command like:
```
this.dynamoDB.send(
      new ScanCommand({
        TableName: this.constraintTableName,
      }),
    );
```
I got this error:
```
EndpointError: Endpoint URL must be a string, got object
    at getEndpointUrl (/home/allanbraun/fontes/cnm-sphinx/.esbuild/.build/node_modules/@aws-sdk/util-endpoints/dist-cjs/utils/getEndpointUrl.js:19:11)
    at evaluateEndpointRule (/home/allanbraun/fontes/cnm-sphinx/.esbuild/.build/node_modules/@aws-sdk/util-endpoints/dist-cjs/utils/evaluateEndpointRule.js:26:50)
    at evaluateRules (/home/allanbraun/fontes/cnm-sphinx/.esbuild/.build/node_modules/@aws-sdk/util-endpoints/dist-cjs/utils/evaluateRules.js:11:89)
    at evaluateTreeRule (/home/allanbraun/fontes/cnm-sphinx/.esbuild/.build/node_modules/@aws-sdk/util-endpoints/dist-cjs/utils/evaluateTreeRule.js:12:46)
    at evaluateRules (/home/allanbraun/fontes/cnm-sphinx/.esbuild/.build/node_modules/@aws-sdk/util-endpoints/dist-cjs/utils/evaluateRules.js:20:81)
[...]

Issue #4076, opened 2022-10-22
Setting a custom endpoint of an SNS client results in EndpointError


darbio:
### Checkboxes for prior research

- [X] I've gone through [Developer Guide](https://docs.aws.amazon.com/sdk-for-javascript/v3/developer-guide) and [API reference](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest)
- [X] I've checked [AWS Forums](https://forums.aws.amazon.com) and [StackOverflow](https://stackoverflow.com/questions/tagged/aws-sdk-js).
- [X] I've searched for [previous similar issues](https://github.com/aws/aws-sdk-js-v3/issues) and didn't find any solution.

### Describe the bug

Setting a custom endpoint of an SNS client results in `EndpointError: Endpoint URL must be a string, got object`.

### SDK version number

@aws-sdk/sns-client@3.194.0

### Which JavaScript Runtime is this issue in?

Node.js

### Details of the browser/Node.js/ReactNative version

v16.18.0

### Reproduction Steps

```
// sns client
let config: SNSClientConfig = {};

if (process.env.IS_OFFLINE) {
    config.endpoint = "http://127.0.0.1:4002";
    config.region = "us-east-1";
}

const client = new SNSClient(config);

console.log(client.config.endpoint);
// Prints:
// [AsyncFunction (anonymous)]

console.log(await client.config.endpoint());
// Prints:
// {
//     hostname: '127.0.0.1',
//     port: 4002,
//     protocol: 'http:',
//     path: '/',
//     query: undefined
//   }

// create the queue
const { TopicArn } = await client.send(new CreateTopicCommand({
    Name: process.env.STACK_SNS_TOPIC_NAME,
    Attributes: {
        DisplayName: "Some Name",
        FifoTopic: "true"
    }
}));

// throws
// EndpointError: Endpoint URL must be a string, got object
```

### Observed Behavior

Throws an EndpointError: Endpoint URL must be a string, got object

### Expected Behavior

Client uses custom URL

samuelabreu:
Same to me trying to use a custom endpoint on CognitoIdentityProviderClient, sticking to 3.171.0 works fine.

only4lee:
bump. Confirm that this also happens with @aws-sdk/client-sqs@3.194.0
When I revert back to 3.171.0 as per @samuelabreu it works fine.

eau-de-la-seine:
[...]

PR #4098, opened 2022-10-26
fix(endpoint): normalize endpoint to string when passing from client config to endpointParams

kuhe:
### Issue
https://github.com/aws/aws-sdk-js-v3/issues/4076
https://github.com/aws/aws-sdk-js-v3/issues/4095

### Description
convert endpoint to string when passing from client config (where it may have been converted to a v1 endpoint by resolveEndpointConfig) to the `endpointParams` resolver input object.

- `client.config.endpoint` can be a string, V1, or V2 endpoint object
- `resolveEndpointConfig` normalizes all of these to a `Provider<Endpoint>`
- when the `endpointParams` are generated from the config mapping instructions, the provider is resolved but the object is not converted to a `string`, which is expected by the endpoints spec and ruleset.

### Testing
- [x] added unit tests
- [x] manual testing with dynamodb client

### DIFF
--- a/packages/middleware-endpoint/src/adaptors/getEndpointFromInstructions.ts
+++ b/packages/middleware-endpoint/src/adaptors/getEndpointFromInstructions.ts
@@ -1,8 +1,9 @@
-import { Endpoint, EndpointParameters, EndpointV2, HandlerExecutionContext } from "@aws-sdk/types";
+import { EndpointParameters, EndpointV2, HandlerExecutionContext } from "@aws-sdk/types";
 
 import { EndpointResolvedConfig } from "../resolveEndpointConfig";
 import { resolveParamsForS3 } from "../service-customizations";
 import { EndpointParameterInstructions } from "../types";
+import { createConfigValueProvider } from "./createConfigValueProvider";
 
 export type EndpointParameterInstructionsSupplier = Partial<{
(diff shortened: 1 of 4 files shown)


kuhe:
tested various commands with

```js
const client: DynamoDBClient = new DynamoDBClient({
  region: "us-east-1",
  logger: console,
  endpoint: "https://dynamodb.us-east-1.amazonaws.com",
});
```

packages/middleware-endpoint/src/adaptors/getEndpointFromInstructions.ts
```diff
@@ -85,18 +85,42 @@ export const resolveParams = async <
...
+ * @param configKey - the key to look up in config.
+ * @param canonicalEndpointParamKey - this is the name the EndpointRuleSet uses.
+ *                                    it will most likely not contain the config
+ *                                    value, but we use it as a fallback.
+ * @param config - container of the config values.
+ * 
+ * @returns async function that will resolve with the value.
  */
-const createConfigProvider = <Config extends Record<string, unknown>>(
+export const createConfigProvider = <Config extends Record<string, unknown>>(
```
trivikr, line 97: Move `createConfigProvider` to it's own file?

This will export createConfigProvider from index.ts which we don't want as of now. The export was added just for testing the component.

trivikr (approved):

Issue #4111, opened 2022-10-27
All properties of client-s3 GetObjectCommandOutput described same as "Body"


IPCPaul:
### Checkboxes for prior research

- [X] I've gone through [Developer Guide](https://docs.aws.amazon.com/sdk-for-javascript/v3/developer-guide) and [API reference](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest)
- [X] I've checked [AWS Forums](https://forums.aws.amazon.com) and [StackOverflow](https://stackoverflow.com/questions/tagged/aws-sdk-js).
- [X] I've searched for [previous similar issues](https://github.com/aws/aws-sdk-js-v3/issues) and didn't find any solution.

### Describe the bug

Since version 3.188.0 (last version without this bug 3.186.0) the client-s3 package describes all properties of GetObjectCommandOutput as "SdkStream<internal.Readable | ReadableStream | Blob | undefined> | undefined", unless the 'DOM' lib is included in the tsconfig.

I guess the change was introduced here:
https://github.com/aws/aws-sdk-js-v3/pull/3977/files
https://github.com/aws/aws-sdk-js-v3/commit/ad99b66944b6fe1ce83082c1a33193dff033cfaf

### SDK version number

@aws-sdk/client-s3@3.197.0, @aws-sdk/client-s3@3.186.0

### Which JavaScript Runtime is this issue in?

Node.js

### Details of the browser/Node.js/ReactNative version

v16.18.0

### Reproduction Steps

```js
import { GetObjectCommand, S3Client } from "@aws-sdk/client-s3";

const s3Client = new S3Client({});
const getObjectCommand = new GetObjectCommand({
  Bucket: "my-bucket-name",
  Key: "/my/s3/key"
});
(async () => {
  const r = await s3Client.send(getObjectCommand);
  const contentLength: number | undefined = r.ContentLength;
  console.log("GOT CONTENT LENGTH:", contentLength);
})()
  .then(() => {
    console.log("Fetched from s3");
  })
  .catch((error) => {
    console.log("error", error)
  })
  .finally(() => {
    console.log("OK, really done.")
  });
```

### Observed Behavior

The code doesn't compile where properties of GetObjectCommandOutput are expected to be of types other than "SdkStream<internal.Readable | ReadableStream | Blob | undefined> | undefined"

### Expected Behavior

[...]

PR #4116, opened 2022-10-28
fix(types): Fix issue with WithSdkStreamMixin

Lewenhaupt:
Fix an issue where WithSdkStreamMixin would apply to more properties than intended due to Distributive Conditional Types.

### Issue
#4111

### Description
Fix an issue where WithSdkStreamMixin would apply to more properties than intended due to Distributive Conditional Types.

### Testing
Tested locally with client-s3 and it's GetObjectCommand-output which now correctly only applies the mixin on "Body".

---

### DIFF
--- a/packages/types/src/serde.ts
+++ b/packages/types/src/serde.ts
@@ -92,7 +92,7 @@ export type SdkStream<BaseStream> = BaseStream & SdkStreamMixin;
  * with the SdkStreamMixin helper methods.
  */
 export type WithSdkStreamMixin<T, StreamKey extends keyof T> = {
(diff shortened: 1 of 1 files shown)


Lewenhaupt:
Seems the build fails because querystring-builder has no tests?

trivikr:
> Seems the build fails because querystring-builder has no tests?

@Lewenhaupt Yeah, we're experimenting with reducing build times in our CI, and it being followed up in https://github.com/aws/aws-sdk-js-v3/pull/4115
I've updated the buildspec in the backend. Can you please submit again?

Also, please update the PR title to the exact change made as it becomes part of CHANGELOG.

PR #4119, opened 2022-10-28
fix(types): Corrected extends check of generic in WithSdkStreamMixin

Lewenhaupt:
Fix an issue where WithSdkStreamMixin would apply to more properties than intended due to Distributive Conditional Types.

### Issue
https://github.com/aws/aws-sdk-js-v3/issues/4111

### Description
Fix an issue where WithSdkStreamMixin would apply to more properties than intended due to Distributive Conditional Types.

### Testing
Tested locally with client-s3 and it's GetObjectCommand-output which now correctly only applies the mixin on "Body".

---

### DIFF
--- a/packages/types/src/serde.ts
+++ b/packages/types/src/serde.ts
@@ -92,7 +92,7 @@ export type SdkStream<BaseStream> = BaseStream & SdkStreamMixin;
  * with the SdkStreamMixin helper methods.
  */
 export type WithSdkStreamMixin<T, StreamKey extends keyof T> = {
-  [key in keyof T]: T[key] extends T[StreamKey] ? SdkStream<T[StreamKey]> : T[key];
+  [key in keyof T]: [T[key]] extends [T[StreamKey]] ? SdkStream<T[StreamKey]> : T[key];
 };
 
 /**


Lewenhaupt:
@trivikr Does your builds get cached? Seems it's the same build.

Is the title better now?

packages/types/src/serde.ts
```diff
@@ -92,7 +92,7 @@ export type SdkStream<BaseStream> = BaseStream & SdkStreamMixin;
  * with the SdkStreamMixin helper methods.
  */
 export type WithSdkStreamMixin<T, StreamKey extends keyof T> = {
-  [key in keyof T]: T[key] extends T[StreamKey] ? SdkStream<T[StreamKey]> : T[key];
+  [key in keyof T]: [T[key]] extends [T[StreamKey]] ? SdkStream<T[StreamKey]> : T[key];
```
kuhe, line 95:
```suggestion
  [key in keyof T]: key extends StreamKey ? SdkStream<T[StreamKey]> : T[key];
```

kuhe:
```typescript
type Stream = {} | undefined;

type Data = {
    undef: undefined
    num: number
    str: string
    stream: Stream
}

type Mixin<T> = T & {
    someAddedMethod(): void
}

// for v1/v2, the match is on the type of the key indicated. This is too wide if the Stream type includes undefined.
type ApplyConditionalTypeMappingV1<T, Key extends keyof T> = {
    [key in keyof T]: T[key] extends T[Key] ? Mixin<T[Key]> : T[key];
}

type ApplyConditionalTypeMappingV2<T, Key extends keyof T> = {
    [key in keyof T]: [T[key]] extends [T[Key]] ? Mixin<T[Key]> : T[key];
}

// this applies the type mixin only to the specified key.
type ApplyConditionalTypeMappingV3<T, Key extends keyof T> = {
    [key in keyof T]: key extends Key ? Mixin<T[Key]> : T[key];
}

const data: ApplyConditionalTypeMappingV3<Data, 'stream'> = {
    num: 1,
    str: 'a',
    stream: undefined,
    undef: undefined,
}
```

Lewenhaupt: Agreed that this is more robust

Lewenhaupt: Updated docs, however this does change the behaviour and I'm not sure if the change from matching multiple keys to only a single key is expected (even though the signature of the mixin definitely indicates it only should match one key).

kuhe: Yes, there will only be one key. I believe this is guaranteed by Smithy.

Issue #4121, opened 2022-10-29
Use of `Function()` violates `unsafe-eval` CSP rule.


nickhudkins:
### Checkboxes for prior research

- [X] I've gone through [Developer Guide](https://docs.aws.amazon.com/sdk-for-javascript/v3/developer-guide) and [API reference](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest)
- [X] I've checked [AWS Forums](https://forums.aws.amazon.com) and [StackOverflow](https://stackoverflow.com/questions/tagged/aws-sdk-js).
- [X] I've searched for [previous similar issues](https://github.com/aws/aws-sdk-js-v3/issues) and didn't find any solution.

### Describe the bug

The `new Function()` constructor was used as part of `evaluateTemplate.ts` within `@aws-sdk/util-endpoints` as can be seen here: https://github.com/aws/aws-sdk-js-v3/blob/main/packages/util-endpoints/src/utils/evaluateTemplate.ts#L37

This results in CSP errors should someone consume part of the aws-sdk that depends upon this package as part of a browser-based application. This does NOT impact Node.JS consumers.

A simple reproduction of the problem can be found here:
https://github.com/nickhudkins/aws-sdk-strict-csp-violations

### SDK version number

@aws-sdk/package-name@>3.160.0

### Which JavaScript Runtime is this issue in?

Browser

### Details of the browser/Node.js/ReactNative version

🤷‍♀️

### Reproduction Steps

Clone: https://github.com/nickhudkins/aws-sdk-strict-csp-violations
Run `yarn && yarn start`
Visit http://127.0.0.1:3000, become sad.

### Observed Behavior

It breaks strict CSPs.

### Expected Behavior

It doesn't break strict CSPs.

### Possible Solution

🤷‍♀️

RanVaknin:
@nickhudkins ,

Thank you for opening this issue and creating the repo to highlight this issue.

I can confirm that running `DescribeAccountAttributesCommand` from a browser environment causes the following error:
```
EvalError: Refused to evaluate a string as JavaScript because 'unsafe-eval' is not an allowed source of script in the following Content Security Policy directive: "script-src https: 'self' https://*.vscode-cdn.net".
```

Will assign to dev team for further investigation.
Thanks,
Ran~

PR #4135, opened 2022-11-01
fix(endpoint): use regexp to construct template string instead of Function

kuhe:
### Issue
https://github.com/aws/aws-sdk-js-v3/issues/4121

### Description
- use regexp to build template string instead of `Function`.

### Testing
existing unit test coverage.


trivikr:
Superceded by https://github.com/aws/aws-sdk-js-v3/pull/4136

PR #4136, opened 2022-11-01
fix(util-endpoints): evaluateTemplate implementation without RegExp/Function

trivikr:
### Issue
Fixes: https://github.com/aws/aws-sdk-js-v3/issues/4121

### Description
Use a simple custom parser for evaluating template instead of RegExp or Function

### Testing
Unit testing

---

### DIFF
--- a/packages/util-endpoints/src/utils/evaluateTemplate.ts
+++ b/packages/util-endpoints/src/utils/evaluateTemplate.ts
@@ -2,43 +2,54 @@ import { getAttr } from "../lib";
 import { EvaluateOptions } from "../types";
 
 export const evaluateTemplate = (template: string, options: EvaluateOptions) => {
-  const evaluatedTemplateArr: Array<String> = [];
+  const evaluatedTemplateArr: string[] = [];
 
   const templateContext = {
     ...options.endpointParams,
     ...options.referenceRecord,
   } as Record<string, string>;
 
-  for (let i = 0; i < template.length; i++) {
-    const char = template[i];
-    const nextChar = template[i + 1];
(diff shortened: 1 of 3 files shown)


packages/util-endpoints/src/utils/evaluateTemplate.ts
```diff
@@ -1,38 +1,46 @@
...
 
-const ATTR_SHORTHAND_REGEX = new RegExp("\\${([\\w]+)#([\\w]+)}", "g");
-
 export const evaluateTemplate = (template: string, options: EvaluateOptions) => {
-  const templateToEvaluate = template
-    // Replace `{value}` with `${value}`
-    .replace(new RegExp(`\{([^{}]+)\}`, "g"), "${$1}")
-    // Replace `{${value}}` with `{value}`
-    .replace(new RegExp(`\{\\$\{([^{}]+)\}\}`, "g"), "{$1}");
+  const evaluatedTemplateArr: Array<String> = [];
```
kuhe, line 5: should be `string[]` or `Array<string>`?

kuhe (approved):

packages/util-endpoints/src/utils/evaluateTemplate.ts
trivikr: I was using `string[]` in the past, but `Array<string>` is better as it's more readable for complex arrays.

Like `(string | number | boolean)[]` vs `Array<string | number | boolean>`
We should enforce a standard across our codebase.

I'll use `string[]` as of now, as that's use at other places.

kuhe (approved):

PR #4141, opened 2022-11-02
chore(packages): replace util-base64-* with util-base64

trivikr:
### Issue
Internal JS-3705
Respective PR for clients https://github.com/aws/aws-sdk-js-v3/pull/4140

### Description
Replaces dependency util-base64-* with util-base64

### Testing
Node.js and Browser testing was done in https://github.com/aws/aws-sdk-js-v3/pull/4140

---
By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #4142, opened 2022-11-03
fix(fetch-http-handler): do not publish CommonJS version

trivikr:
### Issue
Noticed in https://github.com/aws/aws-sdk-js-v3/pull/4141

### Description
The package fetch-http-handler is only used in the browser, and the commonjs version is not needed.
We may merge node-http-handler and fetch-http-handler in future.

### Testing
N/A

---


trivikr:
Closing as this can be taken up while merging node-http-handler and fetch-http-handler

PR #4148, opened 2022-11-07
chore(types): update @types/node to 14

kuhe:
### Issue
n/a

### Description
A yarn issue is blocking my SSO PR, I think this may fix it. 
The change should be valid considering we've dropped node 10 and 12 support.

### DIFF
--- a/packages/util-stream-node/src/sdk-stream-mixin.ts
+++ b/packages/util-stream-node/src/sdk-stream-mixin.ts
@@ -1,10 +1,8 @@
 import { streamCollector } from "@aws-sdk/node-http-handler";
-import { SdkStream, SdkStreamMixin } from "@aws-sdk/types";
+import { BufferEncoding, SdkStream, SdkStreamMixin } from "@aws-sdk/types";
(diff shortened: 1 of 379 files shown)


packages/util-stream-node/src/sdk-stream-mixin.ts
```diff
@@ -32,7 +32,7 @@ export const sdkStreamMixin = (stream: unknown): SdkStream<Readable> => {
     transformToString: async (encoding?: string) => {
       const buf = await transformToByteArray();
       if (encoding === undefined || Buffer.isEncoding(encoding)) {
-        return fromArrayBuffer(buf.buffer, buf.byteOffset, buf.byteLength).toString(encoding);
+        return fromArrayBuffer(buf.buffer, buf.byteOffset, buf.byteLength).toString(encoding as BufferEncoding);
```
kuhe, line 35: I am telling Buffer.toString to not typecheck the encoding. This should not be applied upward to `encoding?: string` because
- I don't think the enumeration is actually exhaustive, but regardless,
- it makes it very hard to make the browser version consistently typed with this one

trivikr (approved):

Issue #4151, opened 2022-11-08
Improve util-dynamodb unmarshalling performance


Rouby:
### Describe the feature

Unmarshalling of maps retrieved from DynamoDB should be as performant as it can be. Currently, unmarshalling maps is very noisy with respect to garbage collection, due to re-creating objects in the map deserialization reducer.

### Use Case

Having big objects in DynamoDB is a possibility (maps with more than 1000 keys). Retrieving these objects via the node-js client takes an unusually long time (>10sec) but not because they are of big size, but because they have many keys.

### Proposed Solution

Directly mutate the object representing unmarshalled map in https://github.com/aws/aws-sdk-js-v3/blob/03a5b9d9fc3ba695b3245ef13ff02dbca54fba22/packages/util-dynamodb/src/convertToNative.ts#L77:L80

### Acknowledgements

- [X] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### SDK version used

3.204.0

### Environment details (OS name and version, etc.)

MacOS / Lambda, NodeJS

PR #4152, opened 2022-11-08
Improve Performance retrieving large objects from DynamoDB

Rouby:
### Issue
#4151 

### Description
This PR improves performance for very large objects retrieved from DynamoDB. 

### Testing
I added a test with a big data-set and measured the timing before and after the change. I did not include the timing as a hard metric to fail / succeed the test.

---

### DIFF
--- a/clients/client-dynamodb-streams/src/protocols/Aws_json1_0.ts
+++ b/clients/client-dynamodb-streams/src/protocols/Aws_json1_0.ts
@@ -551,10 +551,8 @@ const deserializeAws_json1_0MapAttributeValue = (
     if (value === null) {
       return acc;
     }
-    return {
-      ...acc,
-      [key]: deserializeAws_json1_0AttributeValue(__expectUnion(value), context),
-    };
+    acc[key] = deserializeAws_json1_0AttributeValue(__expectUnion(value), context);
+    return acc;
   }, {});
 };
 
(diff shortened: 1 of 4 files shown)


Rouby:
Results of testing:

#### Before

```sh
 PASS  src/unmarshall.spec.ts
 PASS  src/convertToAttrToNative.spec.ts

 RUNS  src/convertToNative.spec.ts

Test Suites: 2 passed, 2 of 3 total
Tests:       125 passed, 125 total
Snapshots:   0 total
Time:        222 s

Aborted before tests could finish.
```

#### After

```sh
 PASS  src/unmarshall.spec.ts
 PASS  src/convertToAttrToNative.spec.ts
 PASS  src/convertToNative.spec.ts

Test Suites: 3 passed, 3 total
Tests:       134 passed, 134 total
Snapshots:   0 total
Time:        2.479 s
```

kuhe:
I will consider this change when making serialization/deserialization optimizations, but cannot accept this as-is because the change has to be made in code generation, for example in https://github.com/awslabs/smithy-typescript/blob/main/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/integration/HttpBindingProtocolGenerator.java

Rouby:
> I will consider this change when making serialization/deserialization optimizations, but cannot accept this as-is because the change has to be made in code generation, for example in https://github.com/awslabs/smithy-typescript/blob/main/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/integration/HttpBindingProtocolGenerator.java

Hey, thanks. Any idea/estimate when this would land?

kuhe:
redirect comments to https://github.com/aws/aws-sdk-js-v3/issues/4151

PR #4156, opened 2022-11-09
fix(types): add client config interface test with s3 example

kuhe:
### Issue
https://github.com/aws/aws-sdk-js-v3/issues/4122

Our client config types are a mash of many different types due to being piped through many config resolvers.
The endpoints 2.0 changes modified the runtime type of the config and broke the type contract for `resolvedEndpointConfig`, which had assumed `endpoint` would be set.

This test package aims to stabilize the client config resolved interface.

### Description
- this adds a private package that is only used for unit testing
- the tests check that a minimal and maximal configuration for a client (S3 used as example) resolve the expected fields 

### Testing
- added unit tests
- added build test for client api (S3 used as example)


private/client-api-test/src/client-s3-interface.spec.ts
```diff
@@ -0,0 +1,86 @@
+import { S3ClientResolvedConfig } from "@aws-sdk/client-s3";
+
+import { initializeWithMaximalConfiguration, initializeWithMinimalConfiguration } from "./client-s3-interface";
+
+export type FIELD_INIT_TYPE = "resolvedByConfigResolver" | "resolvedOnlyIfProvided" | "neverResolved";
+
+const expectedResolvedFields: Record<keyof S3ClientResolvedConfig, FIELD_INIT_TYPE> = {
```
trivikr, line 7: Move `expectedResolvedFields` to it's own file, say `src/client-s3/expected/RESOLVED_FIELDS.ts`? Also, change the data structure to `Record<FIELD_RESOLVED_TYPE, Array<keyof S3ClientResolvedConfig>>` for ease of readability?

For example:
```js
export enum FIELD_RESOLVED_TYPE {
  CONFIG: "Config",
  CUSTOM: "Custom",
  NEVER: "Never"
}
```

```js
export const RESOLVED_FIELDS_S3 = {
   FIELD_RESOLVED_TYPE.CONFIG: ["requestHandler", "apiVersion" /** others */],
   FIELD_RESOLVED_TYPE.CUSTOM: ["customUserAgent" /** others */]
}
```

We can write generic fields later, like in `src/client/expected/RESOLVED_FIELDS.ts` which other clients (including S3) can extend from, and/or make these tests fully/partly automated during codegen.

private/client-api-test/src/client-s3-interface.ts
```diff
@@ -0,0 +1,111 @@
...
+  resolvedConfig.endpoint = void 0;
+  resolvedConfig.isCustomEndpoint = void 0;
+  resolvedConfig.customUserAgent = void 0;
+};
+
+/**
+ * Successful compilation indicates the client can be initialized
+ * with minimal configuration.
+ */
+export const initializeWithMinimalConfiguration = () => {
```
trivikr, line 52: Move to `src/client-s3/initializeWithMinimalConfiguration.ts`?

private/client-api-test/src/client-s3-interface.ts
```diff
@@ -0,0 +1,111 @@
...
+  return new S3Client({});
+};
+
+/**
+ * Successful compilation indicates the client can be initialized
+ * with maximal configuration.
+ */
+export const initializeWithMaximalConfiguration = () => {
```
trivikr, line 60: Move to `src/client-s3/initializeWithMaximalConfiguration.ts`?

trivikr (approved):

PR #4164, opened 2022-11-10
fix(codegen): trigger endpoints param name map setter on additional events

kuhe:
### Issue
internal JS-3690

### Context
Sometimes this map is not set when the Smithy Command Generator is running, leading to incorrect names being written to the Command classes.

### Description
- because setting the map is no-op, it is ok to trigger it additional times
- this PR adds additional calls to set the map on the Integration event methods "customize" and "preprocessModel".

### Testing

- [x] yarn generate-clients
- [x] yarn build:all

### DIFF
--- a/codegen/smithy-aws-typescript-codegen/src/main/java/software/amazon/smithy/aws/typescript/codegen/AddEndpointsV2ParameterNameMap.java
+++ b/codegen/smithy-aws-typescript-codegen/src/main/java/software/amazon/smithy/aws/typescript/codegen/AddEndpointsV2ParameterNameMap.java
@@ -15,25 +15,47 @@
 
 package software.amazon.smithy.aws.typescript.codegen;
 
+import java.util.Map;
+import software.amazon.smithy.model.Model;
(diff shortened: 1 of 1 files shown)


codegen/smithy-aws-typescript-codegen/src/main/java/software/amazon/smithy/aws/typescript/codegen/AddEndpointsV2ParameterNameMap.java
```diff
@@ -15,25 +15,47 @@
...
+        return TypeScriptIntegration.super.preprocessModel(model, settings);
+    }
+
+    @Override
+    public void customize(TypeScriptCodegenContext codegenContext) {
+        setParameterNameMapForAws();
+    }
+
+    private static void setParameterNameMapForAws() {
+        EndpointsParamNameMap.addNameMapping(NAME_MAP);
```
kuhe, line 59: the fix appears to be to change this to addNameMapping instead of `set`, which calls map::clear.

trivikr (approved):

Issue #4183, opened 2022-11-13
TypeError: Cannot read properties of undefined (reading 'sso_session')


guiaramos:
### Checkboxes for prior research

- [X] I've gone through [Developer Guide](https://docs.aws.amazon.com/sdk-for-javascript/v3/developer-guide) and [API reference](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest)
- [X] I've checked [AWS Forums](https://forums.aws.amazon.com) and [StackOverflow](https://stackoverflow.com/questions/tagged/aws-sdk-js).
- [X] I've searched for [previous similar issues](https://github.com/aws/aws-sdk-js-v3/issues) and didn't find any solution.

### Describe the bug

Start getting this error when deploying to AWS Elastic Beanstalk. I am trying to access Secret Manager.

### SDK version number

"@aws-sdk/client-secrets-manager": "^3.209.0"

### Which JavaScript Runtime is this issue in?

Node.js

### Details of the browser/Node.js/ReactNative version

v16.18.0

### Reproduction Steps

1. Deploy the AWS Secret Manager node.js sample code to Elastic Beanstalk.
2. You will see the error when trying to get the secret.

### Observed Behavior

```
Nov 13 22:30:24 ip-10-0-2-9 web: TypeError: Cannot read properties of undefined (reading 'sso_session')
Nov 13 22:30:24 ip-10-0-2-9 web: at /var/app/current/node_modules/@aws-sdk/credential-provider-sso/dist-cjs/fromSSO.js:15:21
Nov 13 22:30:24 ip-10-0-2-9 web: at async coalesceProvider (/var/app/current/node_modules/@aws-sdk/property-provider/dist-cjs/memoize.js:14:24)
Nov 13 22:30:24 ip-10-0-2-9 web: at async SignatureV4.credentialProvider (/var/app/current/node_modules/@aws-sdk/property-provider/dist-cjs/memoize.js:33:24)
Nov 13 22:30:24 ip-10-0-2-9 web: at async SignatureV4.signRequest (/var/app/current/node_modules/@aws-sdk/signature-v4/dist-cjs/SignatureV4.js:86:29)
Nov 13 22:30:24 ip-10-0-2-9 web: at async /var/app/current/node_modules/@aws-sdk/middleware-signing/dist-cjs/middleware.js:16:18
[...]

PR #4186, opened 2022-11-14
fix(credential-provider-sso): add existence check to profile

kuhe:
### Issue
https://github.com/aws/aws-sdk-js-v3/issues/4183

### Description
Add existence check to `profile.sso_session` access.


kuhe:
I checked the v2 implementation for the same issue and it should not be affected since there is already a check.

trivikr (approved):

PR #4194, opened 2022-11-15
chore(lib-dynamodb): reduce object copying in iterators

kuhe:
### Issue
https://github.com/aws/aws-sdk-js-v3/issues/4151
paired with https://github.com/awslabs/smithy-typescript/pull/638

- reduce object copying in iterators

This reducer style makes pointless object copies via `{ ...acc}`:
```js
.reduce((acc, [key, val]) => ({ ...acc, [key]: val }), {});
```
We know that removing these copies will have no side-effects because the initial accumulator value is `{}`, and not an external object that should not be mutated.

New pattern:
```js
.reduce((acc, [key, val]) => (acc[key] = val, acc), {});
```


kuhe:
- [x] generate clients
- [ ] build:all
- [ ] endpoints tests
- [ ] client unit tests
- [ ] feature tests (cucumber)
- [ ] client integration tests

- [ ] manual dynamodb integration test

kuhe:
split codegen changes into https://github.com/aws/aws-sdk-js-v3/pull/4198

trivikr (approved):

PR #4227, opened 2022-11-25
fix(node-http-handler): throw headerName if incoming header value is undefined

ingestkorea:
### Description
throw headerName if incoming header value is undefined

### Testing
unit test

---


kuhe:
why is this change needed? And, I don't see any throw statement implied by the PR name.

Also, it should be opened against the smithy-typescript repository, since the node-http-handler package has moved.
If you move the PR, please provide an explanation of why it is necessary there.

Issue #4231, opened 2022-11-28
TypeError: Passphrase required for encrypted key 


unchaptered:
### Checkboxes for prior research

- [x] I've gone through [Developer Guide](https://docs.aws.amazon.com/sdk-for-javascript/v3/developer-guide) and [API reference](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest)
- [X] I've checked [AWS Forums](https://forums.aws.amazon.com) and [StackOverflow](https://stackoverflow.com/questions/tagged/aws-sdk-js).
- [X] I've searched for [previous similar issues](https://github.com/aws/aws-sdk-js-v3/issues) and didn't find any solution.

### Need to create new Params of passphrase in `@aws-sdk/cloudfront-signer`.

Iternally, `@aws-sdk/cloudfront-signer`  is using **crypto.sign** to create sign.

For creating **Viewer Access Restrictions of cloudfront**, we must use **RSA key only**.

Also, we need to provide passphrase of RSA key for **crypto.sign** when we use this libs.

getSignedUrl and getSignedCookie can receive **passphrase params** in @aws-sdk/cloudfront-signer/src/sign.ts.

So, these function always throw Error about 'TypeError: Passphrase required for encrypted key'.

### SDK version number

@aws-sdk/cloudfront-signer@3.215.0

### Which JavaScript Runtime is this issue in?

Node.js

### Details of the browser/Node.js/ReactNative version

v16.18.0

### prepare RSA-SHA256 key using ssh-keygen

```sh
ssh-keygen -t rsa -b 2048 -m PEM -f private.pem -P abcdeabcdeabcdeabcde
ssh-keygen -e -m PEM -f private.pem > public.pem
```

### call getSIgnedUrl

```javascript
const privateKey = fs.readFileSync(path.join(process.cwd(), 'private.pem'), 'utf8');
const { getSignedUrl } = require('@aws-sdk/cloudfront-signer');

// Sample Variables
const url= 'sample.com';
const keyPairId= 'KEY_PAIR_ID';

const signedUrl = getSignedUrl({
    url,
    keyPairId,
    privateKey,
    dateLessThan: new Date(Date.now() + 1000 * 60 * 60)
});
```

### Observed Behavior

```cmd
TypeError: Passphrase required for encrypted key 
    at Sign.sign (node:internal/crypto/sig:131:29)
[...]

PR #4232, opened 2022-11-28
fix(cloudfront-signer): accept passphrase when signing

unchaptered:
getSIgnedUrl need new params of passphrase of RSA Key.

### Issue

- Issue #4231

### Description

Add new params of passphrase to fix issue.

### Testing

Add new sample of passpharse to fix issue.
But, in testcode, aws have string type of privateKey.
So, I sugggest new privateKey and passphrase from AWS officials.

### Additional context

Please consider its.
Now, getSignedUrl isn't working...

### DIFF
--- a/packages/cloudfront-signer/src/sign.ts
+++ b/packages/cloudfront-signer/src/sign.ts
@@ -11,6 +11,8 @@ export interface CloudfrontSignInputBase {
   keyPairId: string;
   /** The content of the Cloudfront private key. */
   privateKey: string | Buffer;
+  /** The passphrase of RSA-SHA1 key*/
+  passphrase: string;
   /** The date string for when the signed URL or cookie can no longer be accessed. */
   dateLessThan?: string;
   /** The IP address string to restrict signed URL access to. */
@@ -66,6 +68,7 @@ export function getSignedUrl({
   privateKey,
   ipAddress,
   policy,
+  passphrase,
 }: CloudfrontSignInput): string {
   const parsedUrl = parseUrl(url);
   const queryParams: string[] = [];
@@ -75,6 +78,7 @@ export function getSignedUrl({
   const cloudfrontSignBuilder = new CloudfrontSignBuilder({
     keyPairId,
     privateKey,
+    passphrase,
   });
   if (policy) {
     cloudfrontSignBuilder.setCustomPolicy(policy);
@@ -115,10 +119,12 @@ export function getSignedCookies({
   dateLessThan,
   dateGreaterThan,
   policy,
+  passphrase,
 }: CloudfrontSignInput): CloudfrontSignedCookiesOutput {
   const cloudfrontSignBuilder = new CloudfrontSignBuilder({
     keyPairId,
     privateKey,
+    passphrase,
   });
   if (policy) {
     cloudfrontSignBuilder.setCustomPolicy(policy);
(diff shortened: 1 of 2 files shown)


packages/cloudfront-signer/src/sign.ts
```diff
@@ -327,14 +343,20 @@ class CloudfrontSignBuilder {
     };
   }
 
-  private signData(data: string, privateKey: string | Buffer): string {
+  private signData(data: string, privateKey: string | Buffer, passphrase: string): string {
```
trivikr, line 346:
```suggestion
  private signData(data: string, privateKey: string | Buffer, passphrase?: string): string {
```

packages/cloudfront-signer/src/sign.ts
```diff
@@ -327,14 +343,20 @@ class CloudfrontSignBuilder {
...
+        key: privateKey,
+        passphrase: passphrase,
+      },
+      "base64"
+    );
   }
 
-  private signPolicy(policy: string, privateKey: string | Buffer): string {
-    return this.normalizeBase64(this.signData(policy, privateKey));
+  private signPolicy(policy: string, privateKey: string | Buffer, passphrase: string): string {
```
trivikr, line 358:
```suggestion
  private signPolicy(policy: string, privateKey: string | Buffer, passphrase?: string): string {
```

packages/cloudfront-signer/src/sign.ts
```diff
@@ -327,14 +343,20 @@ class CloudfrontSignBuilder {
...
     const sign = createSign("RSA-SHA1");
     sign.update(data);
-    return sign.sign(privateKey, "base64");
+    return sign.sign(
+      {
+        key: privateKey,
+        passphrase: passphrase,
+      },
+      "base64"
+    );
```
trivikr, line 355: nit: can be shortened
```suggestion
    return sign.sign({ key: privateKey, passphrase }, "base64");
```

The passphrase is optional in PrivateKeyInput, so we don't have to do undefined check

trivikr (changes requested):

packages/cloudfront-signer/src/sign.ts
```diff
@@ -11,6 +11,8 @@ export interface CloudfrontSignInputBase {
   keyPairId: string;
   /** The content of the Cloudfront private key. */
   privateKey: string | Buffer;
+  /** The passphrase of RSA-SHA1 key*/
+  passphrase: string;
```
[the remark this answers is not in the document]
unchaptered: Thanks to your kind guide.
I've never consider that about maintaing backward compatibility.


Thanks! :)

packages/cloudfront-signer/src/sign.ts
unchaptered: Oh, thanks for your recommendation.

unchaptered (commented):
Sorry to late response.
Maybe, I  don't know if I understand open source culture properly.

After chang codes as your requests, am I commit it with mesion to you?

@trivikr

unchaptered (commented):
Sorry to late response.
Maybe, I  don't know if I understand open source culture properly.

After chang codes as your requests, am I commit it with mesion to you?

@trivikr

trivikr (approved):

Issue #3846, opened 2022-08-05
DynamoDBDocument can't unmarshall null


vbanderov:
### Checkboxes for prior research

- [X] I've gone through [Developer Guide](https://docs.aws.amazon.com/sdk-for-javascript/v3/developer-guide) and [API reference](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest)
- [X] I've checked [AWS Forums](https://forums.aws.amazon.com) and [StackOverflow](https://stackoverflow.com/questions/tagged/aws-sdk-js).
- [X] I've searched for [previous similar issues](https://github.com/aws/aws-sdk-js-v3/issues) and didn't find any solution.

### Describe the bug

When trying to query for a record containing `null` value through `DynamoDBDocument` an error is thrown:
```
TypeError: Cannot read properties of null (reading 'S')
```

### SDK version number

@aws-sdk/client-dynamodb@3.142.0

### Which JavaScript Runtime is this issue in?

Node.js

### Details of the browser/Node.js/ReactNative version

v16.16.0

### Reproduction Steps

```typescript
import { DynamoDBClient } from '@aws-sdk/client-dynamodb';
import { DynamoDBDocument } from '@aws-sdk/lib-dynamodb';

const client = new DynamoDBClient({
  // ...
});

const document = DynamoDBDocument.from(client);

await document.put({
  TableName: 'table',
  Item: { 
    HashKey: 'hashkey',
    foo: null
   }
});

await document.query({
  TableName: 'table',
  ExpressionAttributeValues: { ':hk': 'hashkey' },
  KeyConditionExpression: `HashKey = :hk`,
});
```

### Observed Behavior

An uncaught error is thrown:
```
error: TypeError: Cannot read properties of null (reading 'S')
```

### Expected Behavior

The item is successfully returned by the query command.

### Possible Solution

Probably need to check if value is `null` before trying to read property 'S' on it. Not sure where that happens though.

RanVaknin:
@vbanderov ,

Im not sure what example you used for this. But this looks like a mix of V2 and V3?

From what I can see you are trying to:
1. put an item in your table
2. retrieve that item from your table. (You used `query` but it seems like you're asking for a single object, so you can use `get` instead)

**Put your Item in the DynamoDB table:**

```js
import { PutCommand, GetCommand } from "@aws-sdk/lib-dynamodb";
[...]

PR #4249, opened 2022-12-02
fix(lib-dynamodb): fix use of log filters in conjunction with DynamDBDocumentClient

kuhe:
### Issue
https://github.com/aws/aws-sdk-js-v3/issues/3846

### Description
DynamoDBDocument client changes input and output shapes. The logger middleware calls filter functions that expect the base shapes and results in errors.

in this PR:
- in logger middleware, use the overrides for log filtering if they exist
- in dynamodbdocument base command, use the previously implemented middleware to insert overrides for log filtering into the handler context

This ensures filtering functions are called on the original input and output shapes unaffected by dynamodbdocument marshaling. 

### Testing
manual Put and Query command using the dynamodbdocument client.

### DIFF
--- a/lib/lib-dynamodb/src/baseCommand/DynamoDBDocumentClientCommand.spec.ts
+++ b/lib/lib-dynamodb/src/baseCommand/DynamoDBDocumentClientCommand.spec.ts
@@ -14,7 +14,7 @@ class AnyCommand extends DynamoDBDocumentClientCommand<{}, {}, {}, {}, {}> {
   protected readonly clientCommand = {
     middlewareStack: {
       argCaptor: this.argCaptor,
-      add(fn, config) {
+      addRelativeTo(fn, config) {
         this.argCaptor.push([fn, config]);
       },
     },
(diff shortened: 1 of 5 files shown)


lib/lib-dynamodb/src/baseCommand/DynamoDBDocumentClientCommand.spec.ts
```diff
@@ -41,7 +41,8 @@ describe("DynamoDBDocumentClientCommand", () => {
       expect(options).toEqual({
         name: "DocumentMarshall",
         override: true,
-        step: "initialize",
+        relation: "before",
+        toMiddleware: "serializerMiddleware",
```
kuhe, line 45: this isn't strictly necessary since the serializer in in the serialize stage and this was in the initialize stage, but more explicit

lib/lib-dynamodb/src/baseCommand/DynamoDBDocumentClientCommand.spec.ts
```diff
@@ -50,7 +51,8 @@ describe("DynamoDBDocumentClientCommand", () => {
       expect(options).toEqual({
         name: "DocumentUnmarshall",
         override: true,
-        step: "deserialize",
+        relation: "before",
+        toMiddleware: "deserializerMiddleware",
```
kuhe, line 55: safer to be explicit about this, but it was already in this order

siddsriv (approved):

Issue #3396, opened 2022-03-04
@aws-sdk/credential-provider-ini - Enable uncached request for credentials


mdpratt:
### Is your feature request related to a problem? Please describe.
As a user, I want want to automatically refresh credentials in my service if it detects a change in credentials. Invoking `fromIni()` will always return the same results, even if  the `~/.aws/credentials` file has changed between invocations.

### Describe the solution you'd like
I want to be able to retrieve current credentials, regardless of how many times I've invoked `fromIni()`. This could be an extra `{ ignoreCache: true }` property that I can provide.

### Describe alternatives you've considered
I've tried using `@aws-sdk/shared-ini-file-loader` directly, and deleting the require / import but that apparently won't clear the `const fileStatusHash: { [key: string]: FileStatus } = {};` object.

```javascript
// Example, assume wrapped in a function
delete require.cache[require.resolve('@aws-sdk/shared-ini-file-loader')]
let loader = require('@aws-sdk/shared-ini-file-loader');
loader.loadSharedConfigFiles().then((p: any) => console.log(p.credentialsFile['default']))
```

mdpratt:
After thinking some more, since the `fileStatusHash` just uses the filename as the key, I can get around this problem by creating and then deleting temporary symlinks. Full example below, when run, will watch the `~/.aws/credentials` file and then `console.log` the default profile:

```javascript
import chokidar from 'chokidar';
import fs from 'fs';
import { homedir } from 'os';
import { resolve } from 'path';
import { nanoid } from 'nanoid';
import { fromIni } from '@aws-sdk/credential-providers';

const { symlink, unlink } = fs.promises;
const AWS_CREDENTIALS = resolve(`${homedir}/.aws/credentials`);

const getCredentials = () => {
  const filepath = `./${nanoid(10)}`
  symlink(AWS_CREDENTIALS, filepath, 'file')
    .then(() => fromIni({ filepath })())
    .then(console.log) // Or do whatever
    .then(() => unlink(filepath));
}

chokidar.watch(AWS_CREDENTIALS).on('change', getCredentials)
```

Westermann:
:+1: I also thing I need this!

Jay1305:
I am using @aws-sdk/shared-ini-file-loader and this kind of property will be definitely helpful in my case when my creds from credentials file are added/modified using another application from backend and my frontend is receiving only cached credentials.

JoshuaWise:
[...]

PR #4253, opened 2022-12-05
feat(shared-ini-file-loader): enable uncached credential loading

Aeolun:
### Issue
#3396

### Description
Adds an optional property `ignoreCache` to the shared credential loader that allows to ignore the existing cache. So that newly updated credential files can be loaded.

### Testing
A test case has been added to the unit test suite.

### Additional context
-

---

### DIFF
--- a/packages/shared-ini-file-loader/src/loadSharedConfigFiles.ts
+++ b/packages/shared-ini-file-loader/src/loadSharedConfigFiles.ts
@@ -20,6 +20,12 @@ export interface SharedConfigInit {
    * `~/.aws/config` otherwise.
    */
   configFilepath?: string;
+
+  /**
+   * Configuration files are normally cached after the first time they are loaded. When this
+   * property is set, the provider will always reload any configuration files loaded before.
+   */
+  ignoreCache?: boolean;
 }
 
 const swallowError = () => ({});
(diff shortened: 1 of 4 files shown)


Aeolun:
Is it possible for someone to look at this?

packages/shared-ini-file-loader/src/loadSharedConfigFiles.ts
```diff
@@ -20,6 +20,12 @@ export interface SharedConfigInit {
    * `~/.aws/config` otherwise.
    */
   configFilepath?: string;
+
+  /**
+   * Configuration files are normally cached after the first time they are loaded. When this
+   * property is set, the provider will always reload any configuration files loaded before.
+   */
+  ignoreCache?: boolean;
```
justinmk3, line 28: drawing on existing precedent, maybe `forceRefresh` makes sense: https://github.com/aws/aws-sdk-js-v3/blob/a79f962873ff11a3d6f98199224155bd2ddc0b1e/packages/property-provider/src/memoize.ts#L70

btw, is there an existing `refresh()` or `reload()` function somewhere in this area? (I didn't look much)

Aeolun: I don't think an existing function exists. I kind of expected that, but no.

Aeolun:
@syall Sorry, poking random committer directly. If there is something I need to do (other than wait) to get this merged please let me know.

kuhe (approved):

PR #4272, opened 2022-12-12
fix(codegen): generate idempotencyToken default for queryParams

kuhe:
internal P75309097

Currently we provide a default idempotency token (uuid v4) if a request structure member has the smithy trait.
We do not provide one if the member is a query parameter. 

- This PR adds the idempotency default token (`uuid.v4()`) for query parameters. 

from https://smithy.io/1.0/spec/core/behavior-traits.html#idempotencytoken-trait

> A unique identifier (typically a [UUID](https://tools.ietf.org/html/rfc4122)) SHOULD be used by the client when providing the value for the request token member. When the request token is present, the service MUST ensure that the request is not replayed within a service-defined period of time. This allows the client to safely retry operation invocations, including operations that are not read-only, that fail due to networking issues or internal server errors. The service uses the provided request token to identify and discard duplicate requests.

> Client implementations MAY automatically provide a value for a request token member if and only if the member is not explicitly provided.

merge together:
https://github.com/awslabs/smithy-typescript/pull/655
https://github.com/aws/aws-sdk-js-v3/pull/4272

### DIFF
--- a/clients/client-accessanalyzer/src/protocols/Aws_restJson1.ts
+++ b/clients/client-accessanalyzer/src/protocols/Aws_restJson1.ts
@@ -299,7 +299,7 @@ export const serializeAws_restJson1DeleteAnalyzerCommand = async (
     false
   );
   const query: any = map({
-    clientToken: [, input.clientToken! ?? generateIdempotencyToken()],
+    clientToken: [, input.clientToken ?? generateIdempotencyToken()],
   });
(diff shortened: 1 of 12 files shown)


kuhe:
- [x] tested with `ScheduleClient::deleteSchedule`

trivikr (approved):

PR #4276, opened 2022-12-12
fix(aws-client-retry-test): use s3 client in case x-ray is not built

kuhe:
followup to https://github.com/aws/aws-sdk-js-v3/pull/4274

switch the sample client to S3, which is already known to be built since tests in `private/aws-client-api-test` work.

### DIFF
--- a/private/aws-client-retry-test/src/ClientRetryTest.spec.ts
+++ b/private/aws-client-retry-test/src/ClientRetryTest.spec.ts
@@ -18,7 +18,7 @@ describe("Middleware-retry integration tests", () => {
       body: Readable.from(""),
     }),
   };
-  const getInsightCommand = new HeadObjectCommand({
+  const headObjectCommand = new HeadObjectCommand({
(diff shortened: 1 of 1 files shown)


private/aws-client-retry-test/src/ClientRetryTest.spec.ts
```diff
@@ -18,11 +18,12 @@ describe("Middleware-retry integration tests", () => {
       body: Readable.from(""),
     }),
   };
-  const getInsightCommand = new GetInsightCommand({
-    InsightId: "foo",
+  const getInsightCommand = new HeadObjectCommand({
```
trivikr, line 21: Also needs to be updated on lines 32, 49 and so on.
I can't provide suggestion as those files are not changed.

```suggestion
  const headObjectCommand = new HeadObjectCommand({
```

trivikr (changes requested):

trivikr (approved):

Issue #4270, opened 2022-12-12
Incorrect non-null assertion in region config breaks Smithy awsRestJson users


RossWilliams:
### Checkboxes for prior research

- [X] I've gone through [Developer Guide](https://docs.aws.amazon.com/sdk-for-javascript/v3/developer-guide) and [API reference](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest)
- [X] I've checked [AWS Forums](https://forums.aws.amazon.com) and [StackOverflow](https://stackoverflow.com/questions/tagged/aws-sdk-js).
- [X] I've searched for [previous similar issues](https://github.com/aws/aws-sdk-js-v3/issues) and didn't find any solution.

### Describe the bug

packages/config-resolver/src/regionConfig/resolveRegionConfig.ts includes a non-null type assertion on [line 52](https://github.com/aws/aws-sdk-js-v3/blob/e7b1535fb6707bb84241d0bd21ca339037094246/packages/config-resolver/src/regionConfig/resolveRegionConfig.ts#L52).

The input type for this function does not require a `useFipsEndpoint` property, but if this property is not supplied this method will throw an exception since it will try to call a property that is undefined. This causes issues for users of Smithy which do not want users to configure FIPS endpoints.

Suggest returning false if `useFipsEndpoint` is not defined and removing non-null assertion.

### SDK version number

@aws-sdk/config-resolver@3.228.0

### Which JavaScript Runtime is this issue in?

Node.js

### Details of the browser/Node.js/ReactNative version

16.8.3

### Reproduction Steps

delete `useFipsEndpoint` config option from any client runtimeConfig.ts file and attempt to execute command

### Observed Behavior

TypeError

### Expected Behavior

No error, `useFipsEndpoint` is optional

### Possible Solution

return false if `useFipsEndpoint` is not defined and removing non-null assertion.

eduardomourar:
[...]

PR #4279, opened 2022-12-13
fix(config-resolver): replace non-null assertions

eduardomourar:
### Issue
#4270

### Description
We will fallback to a `false` value whenever the client SDKs input properties `useDualstackEndpoint` and `useFipsEndpoint` are not set instead of using the TypeScript Non-null assertion operator (`!`).

### Testing
Unit tests

### Additional context
N/A

---

### DIFF
--- a/packages/config-resolver/src/regionConfig/resolveRegionConfig.ts
+++ b/packages/config-resolver/src/regionConfig/resolveRegionConfig.ts
@@ -49,7 +49,10 @@ export const resolveRegionConfig = <T>(input: T & RegionInputConfig & Previously
       if (isFipsRegion(providedRegion)) {
         return true;
       }
-      return typeof useFipsEndpoint === "boolean" ? Promise.resolve(useFipsEndpoint) : useFipsEndpoint!();
+      if (!useFipsEndpoint) {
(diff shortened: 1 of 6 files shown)


packages/config-resolver/src/regionConfig/resolveRegionConfig.ts
```diff
@@ -49,7 +49,10 @@ export const resolveRegionConfig = <T>(input: T & RegionInputConfig & Previously
       if (isFipsRegion(providedRegion)) {
         return true;
       }
-      return typeof useFipsEndpoint === "boolean" ? Promise.resolve(useFipsEndpoint) : useFipsEndpoint!();
+      if (!useFipsEndpoint) {
+        return Promise.resolve(false);
+      }
+      return typeof useFipsEndpoint === "boolean" ? Promise.resolve(useFipsEndpoint) : useFipsEndpoint();
```
kuhe, line 55: could you change this to 

```suggestion
      return typeof useFipsEndpoint !== "function" ? Promise.resolve(!!useFipsEndpoint) : useFipsEndpoint();
```

kuhe (approved):

Issue #4299, opened 2022-12-19
Minimize ruleset object and convert to JSON to reduce publish/install size


trivikr:
### Describe the feature

Minimize ruleset object to reduce publish/install size

### Use Case

The ruleset.js file in the `@aws-sdk/client-s3` is a 1.4MB pure JSON object that is almost entirely whitespace.

```console
$ yarn add @aws-sdk/client-s3@3.231.0 --exact
...

$ ls -lh node_modules/@aws-sdk/client-s3/dist-cjs/endpoint/ruleset.js
-rw-r--r--  1 trivikr  staff   1.4M Dec 19 12:10 node_modules/@aws-sdk/client-s3/dist-cjs/endpoint/ruleset.js
```

This leads to significant increase in install and publish size as seen on PackagePhobia:
* [@aws-sdk/client-s3@3.190.0](https://packagephobia.com/result?p=@aws-sdk/client-s3@3.190.0)
  * Publish size: 3.36 MB
  * Install size: 9.93 MB
* [@aws-sdk/client-s3@3.192.0](https://packagephobia.com/result?p=@aws-sdk/client-s3@3.192.0)
  * Publish size: 6.31 MB
  * Install size: 12.90 MB

### Proposed Solution

Local testing for minimizing ruleSet object inside `ruleSet.ts`

```console
$ tail node_modules/@aws-sdk/client-s3/dist-cjs/endpoint/ruleset.js
...
type":"error"}]}]}]},{"conditions":[],"error":"A region must be set when sending requests to S3.","type":"error"}]}]};

$ ls -lh node_modules/@aws-sdk/client-s3/dist-cjs/endpoint/ruleset.js
-rw-r--r--  1 trivikr  staff    87K Dec 19 12:22 node_modules/@aws-sdk/client-s3/dist-cjs/endpoint/ruleset.js
```

The JS object can be converted to JSON, so that it's easier to parse for JS runtime.

The experiments can be viewed in `@trivikr-test/*` repo:
* Ruleset JS: [@trivikr-test/client-s3@3.230.0-ruleset-js](https://packagephobia.com/result?p=@trivikr-test/client-s3@3.230.0-ruleset-js)
  * Publish size: 7.39 MB
  * Install size: 11.7 MB
* Ruleset JSON: [@trivikr-test/client-s3@3.230.0-ruleset-json](https://packagephobia.com/result?p=@trivikr-test/client-s3@3.230.0-ruleset-json)
[...]

PR #4300, opened 2022-12-20
chore(scripts): add compression routines for endpoints rulesets

kuhe:
### Issue
https://github.com/aws/aws-sdk-js-v3/issues/4299

### Description
Adds a script that compresses each service's ruleset object with one of two algorithms:
- attempts to use the pattern detection recursive algorithm 
- if that fails, it uses the whitespace removal algorithm

This PR does not integrate this script into the client generation process.

main branch
```
du -sh clients/client-s3/src/endpoint/ruleset.ts
804K	clients/client-s3/src/endpoint/ruleset.ts

npm notice name:          @aws-sdk/client-s3
npm notice version:       3.234.0
npm notice filename:      @aws-sdk/client-s3-3.234.0.tgz
npm notice package size:  355.8 kB
npm notice unpacked size: 6.2 MB (yarn build && npm pack without downlevel types)
```

PR branch
```
du -sh clients/client-s3/src/endpoint/ruleset.ts
28K	clients/client-s3/src/endpoint/ruleset.ts

npm notice name:          @aws-sdk/client-s3
npm notice version:       3.234.0
npm notice filename:      @aws-sdk/client-s3-3.234.0.tgz
npm notice package size:  320.5 kB
npm notice unpacked size: 3.8 MB (yarn build && npm pack without downlevel types)
```

### Testing
each algorithm runs self-verification using the input it is given before it is accepted. Because the whitespace removal algorithm is more or less immune to failure, it is the fallback.

Secondly, to show the two formats have no performance impact, the following performance test loads 
- A) the compressed JS format with patterns represented by variable references
- B) the standard uncompressed JSON format

and then walks the entire object and inspects each value.

```js
const marks = {
  js: [],
  json: [],
};

performanceTest();

function performanceTest() {
  {
    for (let i = 0; i < 10000; ++i) {
      const start = performance.now();
      loadJs();
      const end = performance.now();
      marks.js.push(end - start);
    }
  }

  {
    for (let i = 0; i < 10000; ++i) {
      const start = performance.now();
      loadJson();
      const end = performance.now();
      marks.json.push(end - start);
    }
  }

  console.log("js", marks.js.reduce((a, b) => a + b, 0) / marks.js.length);
  console.log("json", marks.json.reduce((a, b) => a + b, 0) / marks.json.length);
}

function loadJs() {
  const data = require("./s3");
  walkObject(data);
}

function loadJson() {
  const data = require("./s3.json");
  walkObject(data);
}

function walkObject(data) {
  switch (typeof data) {
    case "object":
      if (Array.isArray(data)) {
        data.forEach(walkObject);
        break;
      }
      Object.values(data).forEach(walkObject);
      break;
    default:
      if (data === -13534135) {
        throw new Error("comparison failed");
      }
  }
}
```

output:
```
js 0.5221970861673355
json 0.5317003727197647
```


kuhe:
For codegen changes, see https://github.com/aws/aws-sdk-js-v3/pull/4301

scripts/endpoints-ruleset/compress.js
```diff
@@ -0,0 +1,91 @@
...
+const RemoveWhitespace = require("./compression-algorithms/RemoveWhitespace");
+const PatternDetection = require("./compression-algorithms/PatternDetection");
+
+/**
+ * Run compression on ruleset objects for SDK clients.
+ */
+const main = () => {
+  const root = path.join(__dirname, "..", "..");
+  const clientsFolder = path.join(root, "clients");
+  const modelsFolder = path.join(root, "codegen", "sdk-codegen", "aws-models");
```
trivikr, line 15: This performs compression on all clients, which is time consuming for client generation.
Also, the clients may be generated from any glob, although we primarily use models from `codegen/sdk-codegen/aws-models`

The compression needs to be performed only on generated clients.
May be pass codegen output path to it, and only process the clients generated in that directory?

For example, the value in `CODE_GEN_SDK_OUTPUT_DIR` can be passed. It can be called after prettifyCode https://github.com/aws/aws-sdk-js-v3/blob/b03bc8f09c20173b39be840c83fa6fcc9e25adf8/scripts/generate-clients/index.js#L80

trivikr (changes requested):

scripts/endpoints-ruleset/compress.js
trivikr: Instead of reading the ruleset from model, the compression algorithm can attempt reading it from `ruleset.ts` instead. But that would be time consuming.

I'll resolving this conversation as of now. This is a minor issue, and can be taken up as a fast-follow.

scripts/endpoints-ruleset/compress.js
```diff
@@ -0,0 +1,91 @@
...
+
+    const modifiedSource = `// @ts-nocheck
+// generated code, do not edit
+import { RuleSetObject } from "@aws-sdk/util-endpoints";
+
+/* This file is compressed. Log this object
+   or see "smithy.rules#endpointRuleSet"
+   in codegen/sdk-codegen/aws-models/${client}.json */
+
+${selectedAlgorithm.toCodeString("export const ruleSet: RuleSetObject = $ as any;")}
```
trivikr, line 70: Since TypeScript is not run on this file, `as any` is not required.
```suggestion
${selectedAlgorithm.toCodeString("export const ruleSet: RuleSetObject = $;")}
```

We do need `RuleSetObject` though, as that's a promise from ruleset.ts when exporting the object

scripts/endpoints-ruleset/compress.js
```diff
@@ -0,0 +1,91 @@
...
+`;
+
+    fs.writeFileSync(rulesetTs, modifiedSource, "utf-8");
+    fs.rm(rulesetJson, () => {});
```
trivikr, line 74: I don't think rulesetJson is used anywhere, and this line can be removed.
```suggestion
```

scripts/endpoints-ruleset/compress.js
```diff
@@ -0,0 +1,91 @@
...
+  /**
+   * The first algorithm which passes self-verification will be used.
+   */
+  const compressionAlgorithms = [(data) => new PatternDetection(data), (data) => new RemoveWhitespace(data)];
+
+  for (const serviceName of modelsList) {
+    const client = serviceName.replace(".json", "");
+    const rulesetFolder = path.join(clientsFolder, "client-" + client, "src", "endpoint");
+    const rulesetTs = path.join(rulesetFolder, "ruleset.ts");
+    const rulesetJson = path.join(rulesetFolder, "ruleset.json");
```
trivikr, line 35: I don't think rulesetJson is used anywhere, and this line can be removed.
```suggestion
```

trivikr (approved):

PR #4301, opened 2022-12-20
chore(endpoint): add codegen changes for endpoint ruleset compression

kuhe:
This is a follow-up to https://github.com/aws/aws-sdk-js-v3/pull/4300 that adds the actual codegen changes.

It integrates the compression step into the client generation script and prevents prettier from formatting the ruleset files.

- [x] `yarn generate-clients`
- [x] `yarn build:all`
- [x] `(cd tests/endpoints-2.0 && npx jest)`
- [x] `yarn test:integration:legacy`


clients/client-accessanalyzer/src/endpoint/ruleset.ts
```diff
@@ -1,320 +1,31 @@
...
+export const ruleSet: RuleSetObject = _data as any;
```
trivikr, line 31: Generated clients need to be updated after updates in https://github.com/aws/aws-sdk-js-v3/pull/4300
```suggestion
export const ruleSet: RuleSetObject = _data;
```

.prettierignore
```diff
@@ -3,3 +3,4 @@ CHANGELOG.md
 .github/*
 **/*.hbs
 **/*/report.md
+**/*/ruleset.ts
```
trivikr, line 6: Be explicit

```suggestion
clients/*/src/endpoint/ruleset.ts
```

trivikr (changes requested):

clients/client-accessanalyzer/src/endpoint/ruleset.ts
kuhe: just finished rerunning codegen

.prettierignore
trivikr, line 6: Looks like force push overrode the changes?
```suggestion
clients/*/src/endpoint/ruleset.ts
```

trivikr (approved):

PR #4319, opened 2023-01-01
fix(types): changed NodeJS.ProcessEnv type to Record in node-config-provider

froggydood:
node-config-provider

### Description
Changed the NodeJS.ProcessEnv type (in the node-config-provider package), to a Record<string, string | undefined> type. This removes the dependency on @types/node for the node-config-provider. This means that when the types are compiled there is no reliance on it and it can be therefore used in more non-node environments, for example cloudflare worker environments, without any type mismatches or errrors.

### Testing
Changed the type of `envVarGetter` variable in the `fromEnv.spec.ts` file, to match the new type file. Because its only a type change, there is only minimal test changes.

---

### DIFF
--- a/packages/node-config-provider/src/configLoader.spec.ts
+++ b/packages/node-config-provider/src/configLoader.spec.ts
@@ -25,6 +25,7 @@ describe("loadConfig", () => {
     (fromSharedConfigFiles as jest.Mock).mockReturnValueOnce(mockFromSharedConfigFilesReturn);
     const mockFromStatic = "mockFromStatic";
     (fromStatic as jest.Mock).mockReturnValueOnce(mockFromStatic);
+    // Using Record<string, string | undefined> instead of NodeJS.ProcessEnv, in order to not get type errors in non node environments
     const envVarSelector = (env: Record<string, string | undefined>) => env["AWS_CONFIG_FOO"];
     const configKey = (profile: Profile) => profile["aws_config_foo"];
     const defaultValue = "foo-value";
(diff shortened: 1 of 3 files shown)


froggydood:
Created issue [#4329](https://github.com/aws/aws-sdk-js-v3/issues/4329)

packages/node-config-provider/src/configLoader.spec.ts
```diff
@@ -25,7 +25,7 @@ describe("loadConfig", () => {
     (fromSharedConfigFiles as jest.Mock).mockReturnValueOnce(mockFromSharedConfigFilesReturn);
     const mockFromStatic = "mockFromStatic";
     (fromStatic as jest.Mock).mockReturnValueOnce(mockFromStatic);
-    const envVarSelector = (env: NodeJS.ProcessEnv) => env["AWS_CONFIG_FOO"];
+    const envVarSelector = (env: Record<string, string | undefined>) => env["AWS_CONFIG_FOO"];
```
yenfryherrerafeliz, line 28: I agree with the suggested change. Would you be able to please put some comments about why this change is being done?. For example, you can mention that NodeJS.ProcessEnv causes types error on non node environments.

froggydood: I did another commit, and added comments to the code. I think that was what you meant, or did you mean comments to the PR?

yenfryherrerafeliz: Yeah, that is what I meant.

Thanks!

yenfryherrerafeliz (approved):

kuhe (approved):

PR #4333, opened 2023-01-10
chore: add smithy to codeowners

milesziemer:
Adds Smithy as code owners on packages used by generated Smithy code.
---


syall:
The commit message needs to be in a [conventional commit](https://www.conventionalcommits.org/en/v1.0.0/) format for the failed CI action.

.github/CODEOWNERS
```diff
@@ -1 +1,41 @@
...
+packages/util-defaults-mode-browser/* @aws/smithy
+packages/util-defaults-mode-node/* @aws/smithy
+packages/util-endpoints/* @aws/smithy
+packages/util-middleware/* @aws/smithy
+packages/util-retry/* @aws/smithy
+packages/util-stream-browser/* @aws/smithy
+packages/util-stream-node/* @aws/smithy
+packages/util-utf8-browser/* @aws/smithy
+packages/util-utf8-node/* @aws/smithy
+packages/util-waiter/* @aws/smithy
```
kuhe, line 41: why don't we do this as part of transferring these packages to smithy-typescript, instead of here?

milesziemer: Since we're not sure when that's actually going to happen, this is more of a temporary measure we can take right now, until we're able to move the packages over.

syall (approved):

trivikr (approved):

PR #4378, opened 2023-01-26
fix(codeowners): fix syntax and precedence

milesziemer:
### Description
The CODEOWNERS file was set up with incorrect syntax and precedence rules. Previously, the syntax meant that smithy were owners on only files in the root of each specified directory, but not any subdirectories. Additionally, the precedence rules meant that Smithy would be codeowners on these files, rather than both the Smithy team and JS team. This PR fixes both issues.

### Testing
N/A

### Additional context
[Codeowners documentation](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners)

---

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,40 +1,40 @@
 * @aws/aws-sdk-js-team
 
 # These packages are used by Smithy generated code
-packages/config-resolver/* @aws/smithy
-packages/eventstream-serde-browser/* @aws/smithy
-packages/eventstream-serde-config-resolver/* @aws/smithy
-packages/eventstream-serde-node/* @aws/smithy
-packages/fetch-http-handler/* @aws/smithy
-packages/hash-blob-browser/* @aws/smithy
-packages/hash-node/* @aws/smithy
-packages/hash-stream-node/* @aws/smithy
-packages/invalid-dependency/* @aws/smithy
-packages/md5-js/* @aws/smithy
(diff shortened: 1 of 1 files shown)


.github/CODEOWNERS
```diff
@@ -1,40 +1,40 @@
...
-packages/util-defaults-mode-browser/* @aws/smithy
-packages/util-defaults-mode-node/* @aws/smithy
-packages/util-endpoints/* @aws/smithy
-packages/util-middleware/* @aws/smithy
-packages/util-retry/* @aws/smithy
-packages/util-stream-browser/* @aws/smithy
-packages/util-stream-node/* @aws/smithy
-packages/util-utf8/* @aws/smithy
-packages/util-waiter/* @aws/smithy
+/packages/config-resolver/ @aws/smithy @aws/aws-sdk-js-team @aws/smithy
```
trivikr, line 4: `@aws-smithy` is added twice. Is that on purpose?
```suggestion
/packages/config-resolver/ @aws/aws-sdk-js-team @aws/smithy
```

trivikr (changes requested):

.github/CODEOWNERS
milesziemer: Thanks for catching that, it is a mistake. Pushed a fix.

trivikr (approved):

PR #4379, opened 2023-01-26
feat(smithy-client): parse datetime offsets

srchase:
### Issue

### Description
https://github.com/awslabs/smithy/pull/1502 introduced protocol test coverage for datetimes with offsets for clients.

This PR adds a `parseRfc3339DateTimeWithOffset` function to handle positive and negative offsets.

For now, the function is unused by any clients.

### Testing
This was tested in conjunction with [smithy-typescript changes](https://github.com/awslabs/smithy-typescript/pull/681) to use the `parseRfc3339DateTimeWithOffset` function in place of the existing `parseRfc3339DateTime` function for clients.  `parseRfc3339DateTime` is retained for servers. `yarn generate-clients && yarn test:protocols` was successfully run.

The smithy-typescript changes will be paired with codegen changes to use the new parsing function within clients.

### Additional context

---

### DIFF
--- a/packages/smithy-client/src/date-utils.spec.ts
+++ b/packages/smithy-client/src/date-utils.spec.ts
@@ -1,4 +1,31 @@
-import { parseEpochTimestamp, parseRfc3339DateTime, parseRfc7231DateTime } from "./date-utils";
+import {
+  parseEpochTimestamp,
+  parseRfc3339DateTime,
+  parseRfc3339DateTimeWithOffset,
+  parseRfc7231DateTime,
+} from "./date-utils";
+
+const invalidRfc3339DateTimes = [
+  "85-04-12T23:20:50.52Z",
+  "985-04-12T23:20:50.52Z",
+  "1985-13-12T23:20:50.52Z",
+  "1985-00-12T23:20:50.52Z",
+  "1985-4-12T23:20:50.52Z",
+  "1985-04-32T23:20:50.52Z",
+  "1985-04-00T23:20:50.52Z",
+  "1985-04-05T24:20:50.52Z",
+  "1985-04-05T23:61:50.52Z",
+  "1985-04-05T23:20:61.52Z",
+  "1985-04-31T23:20:50.52Z",
+  "2005-02-29T15:59:59Z",
+  "1996-12-19T16:39:57",
+  "Mon, 31 Dec 1990 15:59:60 GMT",
+  "Monday, 31-Dec-90 15:59:60 GMT",
+  "Mon Dec 31 15:59:60 1990",
+  "1985-04-12T23:20:50.52Z1985-04-12T23:20:50.52Z",
+  "1985-04-12T23:20:50.52ZA",
+  "A1985-04-12T23:20:50.52Z",
+];
 
 describe("parseRfc3339DateTime", () => {
   it.each([null, undefined])("returns undefined for %s", (value) => {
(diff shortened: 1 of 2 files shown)


packages/smithy-client/src/date-utils.spec.ts
```diff
@@ -36,6 +41,80 @@ describe("parseRfc3339DateTime", () => {
     "1985-04-05T23:61:50.52Z",
     "1985-04-05T23:20:61.52Z",
     "1985-04-31T23:20:50.52Z",
+    "1996-12-19T16:39:57",
+    "2019-12-16T22:48:18+02:04",
+    "2019-12-16T22:48:18-01:02",
+    "2005-02-29T15:59:59Z",
```
gosar, line 47: nit: move this to after line 43 (`1985-04-31T23:20:50.52Z`) like it was earlier, to keep the grouping of cases of values being beyond the range of month/min/sec together. it reduces the overhead on a future reader to understand what the different cases are covering.

packages/smithy-client/src/date-utils.spec.ts
```diff
@@ -36,6 +41,80 @@ describe("parseRfc3339DateTime", () => {
...
+    "1985-00-12T23:20:50.52Z",
+    "1985-4-12T23:20:50.52Z",
+    "1985-04-32T23:20:50.52Z",
+    "1985-04-00T23:20:50.52Z",
+    "1985-04-05T24:20:50.52Z",
+    "1985-04-05T23:61:50.52Z",
+    "1985-04-05T23:20:61.52Z",
+    "1985-04-31T23:20:50.52Z",
+    "1996-12-19T16:39:57",
     "2005-02-29T15:59:59Z",
```
gosar, line 118: nit: similar to above comment, move this after `1985-04-31T23:20:50.52Z`.

packages/smithy-client/src/date-utils.spec.ts
```diff
@@ -36,6 +41,80 @@ describe("parseRfc3339DateTime", () => {
     "1985-04-05T23:61:50.52Z",
     "1985-04-05T23:20:61.52Z",
     "1985-04-31T23:20:50.52Z",
+    "1996-12-19T16:39:57",
+    "2019-12-16T22:48:18+02:04",
+    "2019-12-16T22:48:18-01:02",
```
gosar, line 46: Add a `// this is allowed by parseRfc3339DateTimeWithOffset but rejected by parseRfc3339DateTime` to make these stand out.
Moreover, instead of adding these 2 cases in the middle of all the cases that are repeated for `parseRfc3339DateTimeWithOffset`, keep them at the end, so the cases that are common are more easily maintainable (e.g., if a new test case needs to be added to both later, it will be easy to copy-paste). Or even share an array variable in code.

gosar (approved):

packages/smithy-client/src/date-utils.spec.ts
srchase: I went ahead moved the common invalid cases to a shared array.

packages/smithy-client/src/date-utils.spec.ts
srchase: Fixed this ordering.

packages/smithy-client/src/date-utils.spec.ts
srchase: Fixed this ordering.

PR #4386, opened 2023-01-31
feat(types): add DNS interfaces, feat(util-dns): add initial HostResolver implementations

syall:
### Issue

N/A.

---

### Description

Adds a new DNS interface `HostResolver` and implementations for both Node.js and browser-like platforms.

This PR adds 2 commits:

- feat(types): add DNS interfaces
- feat(util-dns): add initial HostResolver implementations

---

### Testing
How was this change tested?

1. GitHub CI passes
2. In `packages/util-dns`:
    - `yarn build`
    - `yarn build:include:deps`
    - `yarn test`

---

### Additional context

N/A.

---

### DIFF
--- a/packages/util-dns/package.json
+++ b/packages/util-dns/package.json
@@ -1,0 +1,70 @@
+{
+  "name": "@aws-sdk/util-dns",
+  "version": "3.0.0",
+  "description": "Implementations of DNS host resolvers.",
+  "main": "./dist-cjs/index.js",
+  "module": "./dist-es/index.js",
+  "scripts": {
+    "build": "concurrently 'yarn:build:cjs' 'yarn:build:es' 'yarn:build:types'",
+    "build:cjs": "tsc -p tsconfig.cjs.json",
+    "build:es": "tsc -p tsconfig.es.json",
+    "build:include:deps": "lerna run --scope $npm_package_name --include-dependencies build",
+    "build:types": "tsc -p tsconfig.types.json",
+    "build:types:downlevel": "downlevel-dts dist-types dist-types/ts3.4",
+    "clean": "rimraf ./dist-* && rimraf *.tsbuildinfo",
+    "test": "jest"
+  },
+  "keywords": [
+    "dns"
+  ],
+  "author": {
+    "name": "AWS SDK for JavaScript Team",
+    "url": "https://aws.amazon.com/javascript/"
+  },
+  "license": "Apache-2.0",
+  "dependencies": {
+    "tslib": "^2.3.1"
+  },
+  "devDependencies": {
+    "@aws-sdk/types": "*",
+    "@tsconfig/recommended": "1.0.1",
+    "@types/node": "^14.14.31",
+    "concurrently": "7.0.0",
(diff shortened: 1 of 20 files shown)


packages/util-dns/src/HostResolver.ts
```diff
@@ -0,0 +1,36 @@
+import { HostAddress, HostResolver as IHostResolver, HostResolverArguments } from "@aws-sdk/types";
+
+import { resolveAddress } from "./resolveAddress";
```
syall, line 3: Need feedback on a cleaner implementation (hacky by binding `resolveAddress()` to `this`).

Maybe it would be better to have per-file class implementations of `HostResolver`, e.g. one for the browser, one for `dns.lookup()`, one for `dns.resolve*()`.

packages/util-dns/src/types.ts
```diff
@@ -0,0 +1,26 @@
...
+export interface ResolveAddressFn {
+  (args: HostResolverArguments): Promise<HostAddress[]>;
+}
+
+/**
+ * {@link ResolveAddressFn} provider for binding a {@link HostResolver}
+ */
+export interface ResolveAddressFnProvider {
+  (_this: HostResolver): ResolveAddressFn;
+}
```
syall, line 15: Hacky provider that uses the `_this` parameter to bind `this`.

packages/util-dns/package.json
```diff
@@ -0,0 +1,70 @@
...
+    "./dist-es/HostResolver": "./dist-es/HostResolver.browser",
+    "./dist-es/NodeDnsLookupHostResolver": false,
+    "./dist-es/NodeDnsResolveHostResolver": false
+  },
+  "react-native": {
+    "./dist-es/index": "./dist-es/index.browser",
+    "./dist-es/HostResolver": "./dist-es/HostResolver.browser",
+    "./dist-es/NodeDnsLookupHostResolver": false,
+    "./dist-es/NodeDnsResolveHostResolver": false
+  }
```
syall, line 69: This section needs review since I'm unfamiliar with how these entries work for exports.

packages/util-dns/src/NodeDnsResolveHostResolver.ts
```diff
@@ -0,0 +1,51 @@
...
+ * @internal
+ * {@link HostResolver} implementation that uses the Node.js dns.resolve*() APIs.
+ *
+ * TODO(dns): should NOT be used or exported until fixed to fit the specified
+ * behavior:
+ * - implement cache
+ * - happy eyeballs support
+ * - identity behavior for a valid input IP address
+ */
+export class NodeDnsResolveHostResolver implements IHostResolver {
```
syall, line 14: This class is here as a reference in case we want to move to the `dns.resolve*()` APIs in the future.

This is NOT exported in `util-dns` in this PR.

kuhe (approved):

packages/util-dns/package.json
kuhe: looks right to me

PR #4402, opened 2023-02-03
feat(util-stream): create util-stream package

milesziemer:
### Description
Merges implementations of util-stream-browser and util-stream-node into a single package, util-stream.
The implementations were left unchanged, but some adjustments were made to the `package.json` to
accommodate both implementations, and in the jest config to test the browser implementation.

### Testing
- In `packages/util-stream`: `yarn test`
- In `clients/client-s3`: 
  - Updated dependencies to only use `util-stream`
  - Updated runtime config to use `util-stream`
  - `yarn ts-mocha test/flexibleChecksums.spec.ts`
- Using the [aws-sdk-js-tests repo](https://github.com/aws-samples/aws-sdk-js-tests), tested the [GetObject command](https://github.com/aws/aws-sdk-js-v3/blob/main/clients/client-s3/src/commands/GetObjectCommand.ts):
  - Configured a bucket and object to get, added S3 client, and set up client to run the command
  - Imported the local util-stream package from the aws-sdk-js-v3 directory
  - Modified the client's `config` object to use the `util-stream` implementations of `getAwsChunkedEncodingStream` and `sdk-stream-mixin`
  - Ran `yarn start:node` and `yarn start:web` to see S3 object returned properly

### Additional context
The `util-stream` package was not added to any clients or middleware. This will be done in subsequent commits.

---


kuhe (approved):

srchase (approved):

packages/util-stream/src/getAwsChunkedEncodingStream.browser.ts
```diff
@@ -0,0 +1,37 @@
+import { GetAwsChunkedEncodingStream, GetAwsChunkedEncodingStreamOptions } from "@aws-sdk/types";
```
syall, line 1: Is there a test file for `getAwsChunkedEncodingStream.browser.ts`?

syall (changes requested):

packages/util-stream/src/getAwsChunkedEncodingStream.browser.ts
milesziemer: No there's not (it's not in util-stream-browser), but these getAwsChunkedEncodingStream implementations are going to eventually get merged into a single runtime-agnostic implementation which can have a test.

syall: Got, approving now given that context.

syall (approved):

Issue #3271, opened 2022-02-01
@aws-sdk/lib-dynamodb unmarshall is converting string set to empty object


chuckntaylor:
## Describe the bug
When I use the `DynamoDBDocument.get()` from the @aws-sdk/lib-dynamodb package. I am getting unusual behavior where a string set type 'SS' is coming back as an empty object `{}`. I have two separate projects using this library and configuration, but only one project seems to be having this issue.

## Your environment
Mac OS 12.2
Node v16.13.0
NPM v8.1.0

#### SDK version number
`@aws-sdk/lib-dynamodb@3.49.0`

#### Is the issue in the browser/Node.js/ReactNative?
Node.js

#### Details of the browser/Node.js/ReactNative version
Node v16.13.0
NPM v8.1.0

## Steps to reproduce
Please share code or minimal repo, and steps to reproduce the behavior.

#### Observed behavior
Attached is an image of an Item in the database - taken from the AWS console. Please note, the `products` attribute is the string set I am having trouble with.
<img width="419" alt="Screen Shot 2022-02-01 at 11 59 03 AM" src="https://user-images.githubusercontent.com/17050325/152014653-9f7f2b4f-c520-4647-b78f-286504c6066b.png">

Here is a sample of the code used to retrieve the item from the database:
```ts
export const dbGetCode2 = async (input: { code: string }) => {
  const getParams: GetCommandInput = {
    TableName: CODES_TABLE,
    Key: {
      code: input.code,
    },
  }

  try {
    const response = await dbDocClient.get(getParams)
    if (!response || !response.Item) {
      return null
    } else {
      console.log('Code Found:', pretty(response.Item)) // pretty is just a util for JSON.stringify(obj, null, 2)
      return response.Item
    }
  } catch (err) {
    console.error(err)
    return null
  }
}
```

The end result ends up being this:
```
{
  "purpose": "INSTANT REDEEM CODE",
  "code": "MSYQC3T",
  "products": {},
  "redemption_email": "batman@gotham.org",
  "created_by": "ADMIN",
  "created": 1635774033596,
  "redeemed": true,
  "redemption_date": 1635774039291
}
```

#### Expected behavior
To have products come back as a `Set<string>`

## Additional context

[...]

PR #4404, opened 2023-02-05
feat(util-dynamodb): add unmarshall option to return sets as arrays

MartinDevillers:
### Issue
Related to #3271

### Description
This adds the option `returnSetsAsArrays` to `unmarshall`. When true, this will return DynamoDB String Sets, Number Sets and Binary Sets as arrays, rather than JavaScript's native Set. This is useful when the result of a read operation has to be serialized to JSON. As JSON doesn't support JavaScript's Set, a serializer like `JSON.stringify` will serialize a Set as an empty object, thus losing the data. There may be other scenario's as well where a developer prefers arrays over sets when post-processing data after reading it from DynamoDB.

### Testing
Created unit tests in the `convertToNative.spec.ts` and `unmarshall.spec.ts` file.

### Additional context
The current behavior of `unmarshall` is preserved when the `returnSetsAsArrays` option is omitted or set to `false`.

---

### DIFF
--- a/packages/util-dynamodb/src/convertToNative.spec.ts
+++ b/packages/util-dynamodb/src/convertToNative.spec.ts
@@ -254,6 +254,37 @@ describe("convertToNative", () => {
       const input = ["one", "two", "three"];
       expect(convertToNative({ SS: input })).toEqual(new Set(input));
     });
+
+    it("string set with options.returnSetsAsArrays=true", () => {
+      const input = ["one", "two", "three"];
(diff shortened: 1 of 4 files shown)


MartinDevillers:
@kuhe @trivikr I see you two are the main contributors to this part of the SDK. Any chance you'd be able to look at this PR? Thank you for your time.

kuhe:
This option is not common enough to warrant integrating into the SDK. Please use an external function that recursively converts an object's Sets to arrays.

Issue #2311, opened 2021-04-28
lib-storage Upload does not retry if one part fails.


alvaromat:
## Describe the bug
I'm using Upload from `lib-storage` to get advantage of multipart uploads regarding large files. I would expect it to retry in case any of failure uploading any part. Imagine we have a 5TB file that is splitted in 5MB parts. That would result in many, many parts, so the chance of individual failure is high.

There is the `leavePartsOnError` option. Setting it to true interrupts the whole upload if any part fails.

#### SDK version number
`@aws-sdk/lib-storage@3.13.1`

#### Is the issue in the browser/Node.js/ReactNative?
Browser

#### Details of the browser/Node.js/ReactNative version
```
  Browsers:
    Chrome: 90.0.4430.85
    Firefox: 78.0.2
    Safari: 14.0.3
```

## Steps to reproduce
Upload a file using Upload, exactly like in the example:
```js
       try {
          const upload = new Upload({
            client: s3,
            params: uploadParams,
            leavePartsOnError: false
          })

          upload.on('httpUploadProgress', progress => {
            console.log(progress)
          })

          const uploadResult = await upload.done()
          console.log('done uploading', uploadResult)
        }
        catch (err) {
          showErrors('There was an error uploading your photo: ', err)
        }
```

#### Observed behavior
If there are no errors with any part, the upload goes right.
If there is an error in one or more parts, that part is skipped and the final file is built without those parts. This results in a corrupt file. This can be reproduced stopping the connection (turning WiFi off) or using browser DevTools to stop the connection.

#### Expected behavior
If a single part fails to upload, what is likely to happen with big files in mobile environments, just retry to send that part.
This could be done for example here:
https://github.com/aws/aws-sdk-js-v3/blob/28008c2a3e30cd234f447484d76777d5db56dad1/lib/lib-storage/src/Upload.ts#L124-L130

Do not send the final build command if any part is missing:
https://github.com/aws/aws-sdk-js-v3/blob/28008c2a3e30cd234f447484d76777d5db56dad1/lib/lib-storage/src/Upload.ts#L152-L162

alexforsyth:
@alvaromat you bring up a really good point. I agree that this is a much better experience. However - this isn't  how upload is designed today so I'll change this to a feature request. As you point out, this isn't a difficult change, and it could also be a good first PR for new contributors!

[...]

PR #4414, opened 2023-02-07
fix(s3): make leavePartsOnError default to true

mifi:
### Issue

#2311

### Description
Changes the unsafe default `leavePartsOnError` to true. The default would leave the upload corrupt without anyone knowing.

### Testing
How was this change tested?

### Additional context

---

### DIFF
--- a/lib/lib-storage/README.md
+++ b/lib/lib-storage/README.md
@@ -21,7 +21,7 @@ try {
     ], // optional tags
     queueSize: 4, // optional concurrency configuration
     partSize: 1024 * 1024 * 5, // optional size of each part, in bytes, at least 5MB
-    leavePartsOnError: false, // optional manually handle dropped parts
+    leavePartsOnError: true, // optional throw error on failed parts
   });
 
   parallelUploads3.on("httpUploadProgress", (progress) => {
--- a/lib/lib-storage/src/Upload.ts
+++ b/lib/lib-storage/src/Upload.ts
@@ -44,7 +44,7 @@ export class Upload extends EventEmitter {
   // Defaults.
   private queueSize = 4;
   private partSize = MIN_PART_SIZE;
-  private leavePartsOnError = false;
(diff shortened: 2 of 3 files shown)


lib/lib-storage/README.md
```diff
@@ -21,7 +21,7 @@ try {
     ], // optional tags
     queueSize: 4, // optional concurrency configuration
     partSize: 1024 * 1024 * 5, // optional size of each part, in bytes, at least 5MB
-    leavePartsOnError: false, // optional manually handle dropped parts
+    leavePartsOnError: true, // optional throw error on failed parts
```
alvaromat, line 24: I think this flag name is confusing. I'm bad with naming but maybe change it to something like 
```suggestion
    throwOnPartsError: true, // optional throw error on failed parts
```

mifi: I agree it's a bad name, but that would be an even more breaking change. maybe a todo for the future?

alvaromat: Yes, you're right

RanVaknin:
Hi there @mifi ,

As mentioned in the comments, changing the name for this config type would be considered a breaking change and not something we can merge. 

Can you amend the PR to be a documentation change only? 

Thanks,
Ran~

mifi:
Hi. I didn't change the name of the variable, just the default value from false to true. do you also consider that a breaking change? I would argue it's not, and that it's rather a bug fix because the default value false leads to such a broken behaviour.

RanVaknin:
Hi @mifi ,

The change in behavior would be considered a breaking change since certain customers might rely on the "incorrect" behavior. 

I'll need to discuss this with the team once again to see if we are willing to accept this and cause a "breaking fix".

Thanks,
Ran~

kuhe:
closing this since a larger fix is needed, will be done in #6112

PR #4433, opened 2023-02-13
feat(util-dns): implement caching for NodeDnsLookupHostResolver

syall:
### Issue

N/A.

### Description

Implements DNS caching for NodeDnsLookupHostResolver, including:

- internal `DnsCache` interfaces
- implementation of the cache with `HostEntryTable`
- updated `NodeDnsLookupHostResolver` to use `DnsCache`

The cache hierarchy is as follows:

- `HostEntryTable` maps host name (string) -> `HostEntry`
- `HostEntry` contains good and failed records for IPv4 and IPv6 addresses in a `HostAddressEntryCollection`
- `HostAddressEntryCollection` is used for LRU behavior, and stores `HostAddress` with expiration timestamps

The address cache workflow is as follows for `NodeDnsLookupHostResolver.resolveAddress()`:

- If there was no previous DNS resolution for a host name, create a new DnsCacheEntry with new addresses
- If there was a previous DNS resolution already cached:
  - Update with a new DNS resolution with new addresses if the next-time-to-update timestamp is reached
  - Otherwise, noop
- Process the records by removing expired addresses (except 1 address of each type in case of a DNS outage)
  - May promote a non-expired failed address if there are no good addresses
- Return the first available IPv4 and IPv6 addresses, and append them to the end of the internal DnsCacheEntryCollection
    as the most recently used addresses.
  - Will throw an error if no addresses are available

### Testing
How was this change tested?

Added tests for cache scenarios (most of the PR are the tests).

1. `yarn build`
2. `yarn test`

### Additional context

Most of the behavior is based on the [AWS CRT `host_resolver.c` implementation](https://github.com/awslabs/aws-c-io/blob/main/source/host_resolver.c).

---


packages/util-dns/src/NodeDnsLookupHostResolver.ts
```diff
@@ -20,20 +23,144 @@ const DNS_LOOKUP_OPTIONS = {
...
+   * @internal
+   */
+  private nodeDnsLookup: NodeDnsLookupFn;
+
+  constructor(config?: NodeDnsLookupHostResolverConfig) {
+    this.ttlMs = config?.ttlMs !== undefined ? config.ttlMs : NodeDnsLookupHostResolver.DEFAULT_TTL_MS;
+    this.cache = config?.cache !== undefined ? config.cache : NodeDnsLookupHostResolver.DEFAULT_CACHE_PROVIDER();
+    this.nodeDnsLookup =
+      config?.nodeDnsLookup !== undefined ? config.nodeDnsLookup : NodeDnsLookupHostResolver.DEFAULT_NODE_DNS_LOOKUP;
+  }
```
kuhe, line 102: this can be destructured with defaults:

```ts
public constructor({
  ttlMs = NodeDnsLookupHostResolver.DEFAULT_TTL_MS,
  cache = NodeDnsLookupHostResolver.DEFAULT_CACHE_PROVIDER(),
  nodeDnsLookup = NodeDnsLookupHostResolver.DEFAULT_NODE_DNS_LOOKUP,
}: NodeDnsLookupHostResolverConfig = {}) {
  this.ttlMs = ttlMs;
  this.cache = cache;
  this.nodeDnsLookup = nodeDnsLookup;
}
```

Also, NodeDnsLookupHostResolver.DEFAULT_CACHE_PROVIDER() is a static method? Probably shouldn't be screaming snake case.

packages/util-dns/src/NodeDnsLookupHostResolver.ts
```diff
@@ -20,20 +23,144 @@ const DNS_LOOKUP_OPTIONS = {
...
+
   /**
    * Node.js resolveAddress() using the dns.lookup() APIs.
    * @see https://nodejs.org/api/dns.html#dnspromiseslookuphostname-options
    * @param args arguments with host name query addresses for
    * @returns promise with a list of {@link HostAddress}
    */
-  async resolveAddress(args: HostResolverArguments): Promise<HostAddress[]> {
+  private async nodeDnsLookupResolveAddress(args: HostResolverArguments): Promise<HostAddress[]> {
```
kuhe, line 161: private methods should go below public methods

packages/util-dns/src/util/LinkedList.ts
```diff
@@ -0,0 +1,109 @@
+/**
+ * LinkedList implementation that implements {@link DnsCacheEntryCollection}
+ * for {@link HostEntry}.
+ * Can use any {@link NodeT} that conforms to the {@link LinkedListNode} interface.
+ * @internal
+ */
+export class LinkedList<NodeT extends LinkedListNode<NodeT>> {
```
kuhe, line 7: JS array is sufficiently optimizable serve as LL. Should be benchmarked if concerning, but I don't think we should have a custom LL implementation for a specific interface.

```ts
class DnsCacheEntryImpl<T> {
    private data: T[] = [];

    public get length(): number {
        return this.data.length;
    }

    public cycle(collection?: DnsCacheEntryImpl<T>) {
        const poll = this.data.shift();
        if (!poll) {
            throw new Error('collection is empty');
        }
        (collection ?? this).data.push(poll);
        return poll;
    }

    public append(item: T) {
        this.data.push(item);
    }

    public remove(item: T): T {
        const index = this.data.indexOf(item);
        const removed = this.data[index];
        this.data.splice(index, 1);
        return removed;
    }
}
```

syall: Resolved in https://github.com/aws/aws-sdk-js-v3/pull/4433/commits/0ac10ffe7f385e67c9693bb44f3391d4cd475eb4

packages/util-dns/src/NodeDnsLookupHostResolver.ts
syall: Resolved in https://github.com/aws/aws-sdk-js-v3/pull/4433/commits/a3eab77afad8781e943b9c0d9c5166d5fb02febb and https://github.com/aws/aws-sdk-js-v3/pull/4433/commits/2cb4adfcb94dc1ad13ba075055e430d1aa50afd7

packages/util-dns/src/NodeDnsLookupHostResolver.ts
syall: Resolved in https://github.com/aws/aws-sdk-js-v3/pull/4433/commits/ed00d144ed0799f846797df9c989a5e03ddbb32d

packages/util-dns/src/util/LinkedList.ts
syall: Also if anyone is curious, did a short analysis of LinkedList vs Array implementations of a `DnsCacheEntryCollection`: https://gist.github.com/syall/4495ee0b8f8648cfdb0e6430d3df45e1

At <100 entries (which is more aligned to reality):

- Arrays perform better for `append()` 
- Arrays perform slightly better for `cycle()`
- Arrays and LinkedLists have negligible differences in performance for `remove()`.

However, it should be noted that at scale (max of ~100,000 entries):

- Arrays perform better for `append()`
- LinkedLists perform better for `cycle()`
- LinkedLists perform better for `remove()`

siddsriv (approved):

Issue #4459, opened 2023-02-22
(short issue description)


JoeWHoward:
### Describe the issue

Client-DynamoDB's QueryOutput type has incorrect wording on the `Count` attribute of the type.  It incorrectly states:  `The number of items evaluated, before any QueryFilter is applied.` 

It should say:  The number of items evaluated, **after** any QueryFilter is applied. 

### Links

https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/clients/client-dynamodb/modules/queryoutput.html#count

The link is correct, but the in-code documentation is not.

![image](https://user-images.githubusercontent.com/9206287/220675759-2b23bf1b-08eb-49ac-9f92-cd7dd9eead3d.png)

RanVaknin:
Hi @JoeWHoward ,

Thank you for opening this issue and your PR. 
Since the SDK is auto-generated from the API of the service which it interacts with, we don't maintain any documentation on the client side. The docs are written and maintained by the service team.  Your PR will get overwritten the next time the service team updates their model.

In order to submit a documentation change I suggest clicking on the `Feedback` button at the top right corner of [this page](https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_Query.html). This would upstream your request to a doc writer associated with the service team and they will be able to review your request and make the necassary changes so ultimately they will be propagated down and amended on the SDK level.

Since there is no action the SDK team can do at the moment I feel confident we can close this. 

Thanks again for taking the time to write up all of this.
Ran~

PR #4460, opened 2023-02-22
fix(client-dynamodb): documentation wording on QueryOutput.Count in r…

JoeWHoward:
…elation to QueryFilter

### Issue
#4459 

### Description
This changes one word, from before to after, reflecting the actual behavior of the Count attribute.

### Testing
N/a, the reasoning was confirmed both by other sources of documentation and inspecting results from my own aws cli ddb query.

### Additional context

---

### DIFF
--- a/clients/client-dynamodb/src/models/models_0.ts
+++ b/clients/client-dynamodb/src/models/models_0.ts
@@ -7821,7 +7821,7 @@ export interface QueryOutput {
   Count?: number;
 
   /**
-   * <p>The number of items evaluated, before any <code>QueryFilter</code> is applied. A high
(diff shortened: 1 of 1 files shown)


RanVaknin:
Hi @JoeWHoward ,

We can't accept this PR.  Everything under the `Clients` folder is auto-generated code. This means that there is an upstream "source of truth" which is the API, only the service team can influence the change in the documentation and would have to be resolved with them first.

You can submit your feedback using [this link](https://docs-feedback.aws.amazon.com/feedback.jsp?hidden_service_name=DynamoDB&topic_url=https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_Query.html).

Thanks again for trying to contribute.

Issue #4490, opened 2023-03-03
Typescript compilation fails for HttpRequest using SignatureV4 and NodeHttpHandler.


samurous:
### Checkboxes for prior research

- [X] I've gone through [Developer Guide](https://docs.aws.amazon.com/sdk-for-javascript/v3/developer-guide) and [API reference](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest)
- [X] I've checked [AWS Forums](https://forums.aws.amazon.com) and [StackOverflow](https://stackoverflow.com/questions/tagged/aws-sdk-js).
- [X] I've searched for [previous similar issues](https://github.com/aws/aws-sdk-js-v3/issues) and didn't find any solution.

### Describe the bug

When trying to use the HttpRequest Object from SignatureV4.sign in NodeHttpHandler, the types do not match.

### SDK version number

"@aws-sdk/credential-provider-node": "^3.282.0",     "@aws-sdk/node-http-handler": "^3.282.0",     "@aws-sdk/protocol-http": "^3.282.0",     "@aws-sdk/signature-v4": "^3.282.0",

### Which JavaScript Runtime is this issue in?

Node.js

### Details of the browser/Node.js/ReactNative version

Node v18.14.2

### Reproduction Steps

Run
```bash
npm i @aws-sdk/credential-provider-node@3.282.0 @aws-sdk/node-http-handler@3.282.0 @aws-sdk/protocol-http@3.282.0 @aws-sdk/signature-v4@3.282.0
```

Try to compile
```typescript
import { SignatureV4 } from "@aws-sdk/signature-v4";
import { NodeHttpHandler } from "@aws-sdk/node-http-handler";
import { Sha256 } from "@aws-crypto/sha256-js";
import { HttpRequest } from "@aws-sdk/protocol-http";
import { defaultProvider as defaultCredentialProvider } from "@aws-sdk/credential-provider-node";

const signer = new SignatureV4({
  service: "execute-api",
  region: "eu-central-1",
  credentials: defaultCredentialProvider(),
  sha256: Sha256,
});

export const handler = async (url: string) => {
  const apiUrl = new URL(url);
  const request = new HttpRequest({
    hostname: apiUrl.hostname.toString(),
    path: apiUrl.pathname,
    method: "GET",
    headers: {
      "Content-Type": "application/json",
      host: apiUrl.hostname.toString(),
    },
  });
  const signedRequest = await signer.sign(request);
[...]

PR #4503, opened 2023-03-08
fix(node-http-handler): fix type signature of NodeHttpHandler::handle

kuhe:
### Issue
https://github.com/aws/aws-sdk-js-v3/issues/4490

### Description
- update signature to use the interface instead of the class. For some reason we have a class and interface with the same name exported from HttpRequest, and the class has additional methods not in the interface.

### Testing
CI

### DIFF
--- a/packages/node-http-handler/src/node-http-handler.ts
+++ b/packages/node-http-handler/src/node-http-handler.ts
@@ -1,6 +1,6 @@
-import { HttpHandler, HttpRequest, HttpResponse } from "@aws-sdk/protocol-http";
+import { HttpHandler, HttpResponse } from "@aws-sdk/protocol-http";
 import { buildQueryString } from "@aws-sdk/querystring-builder";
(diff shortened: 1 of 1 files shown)


packages/node-http-handler/src/node-http-handler.ts
```diff
@@ -75,7 +75,7 @@ export class NodeHttpHandler implements HttpHandler {
     this.config?.httpsAgent?.destroy();
   }
 
-  async handle(request: HttpRequest, { abortSignal }: HttpHandlerOptions = {}): Promise<{ response: HttpResponse }> {
+  async handle(request: IHttpRequest, { abortSignal }: HttpHandlerOptions = {}): Promise<{ response: HttpResponse }> {
```
kuhe, line 78: this is compatible because `HttpRequest` extends `IHttpRequest`.

trivikr (approved):

PR #4508, opened 2023-03-10
feat(node-http-handler): implement connections pool and manager interfaces

AndrewFossAWS:
### Description
Implement new connections pool and manager interfaces

### Testing
yarn test:all

---

### DIFF
--- a/packages/node-http-handler/src/node-http-handler.ts
+++ b/packages/node-http-handler/src/node-http-handler.ts
@@ -6,8 +6,6 @@ import { Agent as hsAgent, request as hsRequest, RequestOptions } from "https";
 
 import { NODEJS_TIMEOUT_ERROR_CODES } from "./constants";
 import { getTransformedHeaders } from "./get-transformed-headers";
-import { setConnectionTimeout } from "./set-connection-timeout";
-import { setSocketTimeout } from "./set-socket-timeout";
 import { writeRequestBody } from "./write-request-body";
 
 /**
@@ -15,12 +13,28 @@ import { writeRequestBody } from "./write-request-body";
  */
 export interface NodeHttpHandlerOptions {
   /**
+   * @deprecated Use {@link requestTimeout}
+   *
+   * Note:{@link NodeHttpHandler} will resolve request timeout via nullish coalescing the following fields:
+   * {@link requestTimeout} ?? {@link connectionTimeout} ?? {@link socketTimeout} ?? {@link DEFAULT_REQUEST_TIMEOUT}
+   *
    * The maximum time in milliseconds that the connection phase of a request
    * may take before the connection attempt is abandoned.
    */
   connectionTimeout?: number;
 
   /**
+   * The maximum time in milliseconds that the connection phase of a request
+   * may take before the connection attempt is abandoned.
+   */
+  requestTimeout?: number;
+
+  /**
+   * @deprecated Use {@link requestTimeout}
+   *
+   * Note:{@link NodeHttpHandler} will resolve request timeout via nullish coalescing the following fields:
+   * {@link requestTimeout} ?? {@link connectionTimeout} ?? {@link socketTimeout} ?? {@link DEFAULT_REQUEST_TIMEOUT}
+   *
    * The maximum time in milliseconds that a socket may remain idle before it
    * is closed.
    */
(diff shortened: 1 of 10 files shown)


packages/node-http-handler/src/node-http-handler.ts
```diff
@@ -59,12 +68,14 @@ export class NodeHttpHandler implements HttpHandler {
...
   private resolveDefaultConfig(options?: NodeHttpHandlerOptions | void): ResolvedNodeHttpHandlerConfig {
-    const { connectionTimeout, socketTimeout, httpAgent, httpsAgent } = options || {};
+    const { requestTimeout, connectionTimeout, socketTimeout, httpAgent, httpsAgent } = options || {};
     const keepAlive = true;
     const maxSockets = 50;
+
     return {
       connectionTimeout,
       socketTimeout,
+      requestTimeout: requestTimeout ?? connectionTimeout ?? socketTimeout ?? DEFAULT_REQUEST_TIMEOUT,
```
kuhe, line 78: this priority order should be mentioned in the deprecation message of the two fields

packages/node-http-handler/src/node-http2-connection-pool.ts
```diff
@@ -0,0 +1,42 @@
...
+  public contains(session: ClientHttp2Session): boolean {
+    return this.sessions.includes(session);
+  }
+
+  public remove(session: ClientHttp2Session): void {
+    this.sessions = this.sessions.filter((s) => s !== session);
+  }
+
+  public [Symbol.iterator]() {
+    return this.sessions.values();
```
kuhe, line 30: I don't think this works unless it is `this.sessions[Symbol.iterator]()`

packages/node-http-handler/src/node-http2-connection-manager.ts
```diff
@@ -0,0 +1,115 @@
...
+        if (!session.destroyed) {
+          session.destroy();
+        }
+        connectionPool.remove(session);
+      }
+      this.sessionCache.delete(key);
+    }
+  }
+
+  public setMaxConcurrency(maxConcurrency: number) {
```
kuhe, line 105: should this enforce a positive number?

packages/types/src/connection/manager.ts
```diff
@@ -0,0 +1,30 @@
+import { RequestContext } from "../transfer";
+import { ConnectConfiguration } from "./config";
+
+export interface ConnectionManagerConfiguration {
+  /**
+   * Maximum number of allowed concurrent requests per connection.
+   */
+  maxConcurrency?: number;
+
+  disableConcurrency?: boolean;
```
kuhe, line 10: this is equivalent to maxConcurrency = 1 and overrides that field?

packages/types/src/connection/pool.ts
```diff
@@ -0,0 +1,28 @@
+export interface ConnectionPool<T> {
```
kuhe, line 1: for this and ConnectionManager, are there any practical limits on what `T` is that should be expressed with `T extends ...`?

packages/types/src/connection/manager.ts
AndrewFossAWS: Yes, the field will be deprecated and handle to maintain backwards compatibility

AndrewFossAWS: No, `disableConcurrency` controls disabling/enabling concurrency. 

For http2, a [`session`](https://nodejs.org/api/http2.html#http2_class_clienthttp2session) contains multiple [stream](https://nodejs.org/api/http2.html#class-http2stream) instances.

The current `NodeHttp2Handler` implementation creates a new http2 `session` per url request when `disableConcurrency` is set to `true`. i.e. not reusing existing session and not utilizing concurrent streams

`maxConcurrency` dictates the number of concurrent streams per session. By default, a http2 session allows a maximum of 2 ^ 31 - 1 stream instances.
If `disableConcurrency` is set to `true`, then `maxConcurrency` becomes meaningless.

packages/types/src/connection/pool.ts
AndrewFossAWS: So we are only implementing this interface for http2 and `T` is essentially `ClientHttp2Session`.

The Node.js http module manages connections internally using httpAgent, and browser does not expose any APIs for managing connections. So the `ConnectionPool` and `ConnectionManager` interfaces might not need generics. 

I will confirm with @trivikr about this.

kuhe (approved):

packages/node-http-handler/src/node-http-handler.ts
```diff
@@ -31,12 +45,15 @@ export interface NodeHttpHandlerOptions {
 }
 
 interface ResolvedNodeHttpHandlerConfig {
+  requestTimeout?: number;
```
milesziemer, line 48: I may be reading it incorrectly, but it seems like `requestTimeout` always gets supplied a value in `resolveDefaultConfig` since it defaults to `DEFAULT_REQUEST_TIMEOUT. Can this be non-optional?

AndrewFossAWS: Good spot. Updated it to non-optional.

PR #4514, opened 2023-03-10
feat(protocol-http): implement SRA HttpRequest

milesziemer:
### Description
This change implements the core functionality of HttpRequest for SRA. New properties were added to the HttpRequest class, old properties and interfaces were deprecated. Proxy was added for headers and query properties to maintain backward compatibility. This change does not include updates to the SDK to use new functionality.

Notes on other changes:
* middleware-host-header and middleware-sdk-transcribe-streaming tests: URL.port always returns '' if port is set to the default for the protocol.
* middleware-sdk-transcribe-streaming: Providing URL.hostname with a string containing the port number does not update the port.
* signature-v4 prepareRequest: See comment in code.
* Field: toString method was changed to only require double quotes on values with commas. Current tests allow spaces without double quotes.
* Fields: Helper methods were added for getting all fields and creating a Fields object.

Additional notes:
* When converting between single string header values and a list of field values, split(",") and join(",") are used to avoid making any modifications to the original value, preserving compatibility.

### Testing
yarn test:all

### Additional context
Because custom middleware can have access to the HttpRequest object, Proxies, getters, and setters are being used
to allow old members of HttpRequest to be used. 

---


packages/protocol-http/src/headersProxy.spec.ts
```diff
@@ -0,0 +1,156 @@
...
+    { name: "foo", values: ["bar"] },
+    { name: "baz", values: ["qux", "quux"] },
+  ]);
+
+  describe("proxy works like a normal Record", () => {
+    describe("access and mutation", () => {
+      it("can get and set individual keys", () => {
+        const headers = getHeadersProxy(new Fields({}));
+        headers["foo"] = "bar";
+        headers.baz = "qux,quux";
```
kuhe, line 19: What if commas come with spaces?

packages/protocol-http/src/httpRequest.ts
```diff
@@ -44,21 +165,31 @@ export class HttpRequest implements HttpMessage, Endpoint {
...
   clone(): HttpRequest {
-    const cloned = new HttpRequest({
+    return HttpRequest.from({
       ...this,
-      headers: { ...this.headers },
+      destination: new URL(this.destination),
+      fields: Fields.from(
+        this.fields.getAll().map((field) => ({
+          ...field,
+          values: [...field.values],
```
kuhe, line 174: Perhaps fields should implement clone. Hard to remember to do this every time.

kuhe (approved):

packages/protocol-http/src/headersProxy.spec.ts
milesziemer: Whitespace is preserved. I have a test for it [here](https://github.com/aws/aws-sdk-js-v3/pull/4514/files#diff-6d4dc8b75cf95d8327296e6bc2c7b4cd3166d09b8215b3972990d6192728b1edR152-R155), but do you think its worth having one in the getHeadersProxy tests?

packages/protocol-http/src/httpRequest.ts
milesziemer: Yea I thought about it. I figured we could implement it later if we needed it rather than exposing it now, but I don't have a strong opinion either way.

packages/protocol-http/src/headersProxy.spec.ts
kuhe: it's ok if you can't think of any problems it would cause

PR #4520, opened 2023-03-13
feat(protocol-http): implement SRA HttpResponse

milesziemer:
### Description
This change implements the core functionality of HttpResponse for SRA. New properties were added to the HttpResponse class, old properties and interfaces were deprecated. To maintain compatibility with old implementation, Proxy was added for headers property, and Object.defineProperties is used with getters/setters for other old properties. This change does not include updates to the SDK to use new functionality.

Additional notes:
* HttpResponse.isInstance was left unchanged to maintain compatibility. Previous implementation only checked for specific properties, so it would return true when passed the HttpResponse interface from types package.
* https://github.com/awslabs/smithy-typescript/pull/698 is a related change that ensures protocol tests are using HttpResponse constructor. Due to type changes in HttpResponse, this is required.

### Testing
CI, `yarn test:integration:legacy`

---

### DIFF
--- a/packages/protocol-http/src/httpResponse.ts
+++ b/packages/protocol-http/src/httpResponse.ts
@@ -1,24 +1,88 @@
-import { HeaderBag, HttpMessage, HttpResponse as IHttpResponse } from "@aws-sdk/types";
+import { HeaderBag, HttpMessage, Response } from "@aws-sdk/types";
 
-type HttpResponseOptions = Partial<HttpMessage> & {
+import { Fields } from "./Fields";
+import { getHeadersProxy, headersToFields } from "./headersProxy";
+
+export type HttpResponseOptions = Partial<HttpMessage> & {
   statusCode: number;
 };
 
-export interface HttpResponse extends IHttpResponse {}
+export type HttpResponseFromOptions = {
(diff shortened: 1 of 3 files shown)


packages/protocol-http/src/httpResponse.ts
```diff
@@ -1,24 +1,88 @@
...
+
+import { Fields } from "./Fields";
+import { getHeadersProxy, headersToFields } from "./headersProxy";
 
 type HttpResponseOptions = Partial<HttpMessage> & {
   statusCode: number;
 };
 
-export interface HttpResponse extends IHttpResponse {}
+type HttpResponseFromOptions = {
```
kuhe, line 10: this looks like it's part of the public API, so it should be exported, along with `HttpResponseOptions`

kuhe (approved):

packages/protocol-http/src/httpResponse.ts
milesziemer: Updated

AndrewFossAWS (approved):

milesziemer:
Closed while issues with header proxy are worked out.

PR #4521, opened 2023-03-13
fix(middleware-sdk-ec2): use endpoints v2 for source region endpoint

kuhe:
### Issue
internal request

### Description
explanation inline with code

### Testing
tested locally with

```js
  const ec2 = new EC2Client({ region: "us-iso-east-1" });
  const copySnapshotCommand = new CopySnapshotCommand({
    SourceRegion: "us-iso-west-1",
    SourceSnapshotId: "snap-0c5f208e931abe79f",
  });
```

and confirmed that the request used the source region's resolved endpoint for the presigned url and the target region's resolved endpoint for the regular request.

### DIFF
--- a/packages/middleware-sdk-ec2/src/index.ts
+++ b/packages/middleware-sdk-ec2/src/index.ts
@@ -1,4 +1,4 @@
-import { toEndpointV1 } from "@aws-sdk/middleware-endpoint";
+import { getEndpointFromInstructions, toEndpointV1 } from "@aws-sdk/middleware-endpoint";
 import { HttpRequest } from "@aws-sdk/protocol-http";
 import { SignatureV4 } from "@aws-sdk/signature-v4";
 import {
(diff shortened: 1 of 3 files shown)


packages/middleware-sdk-ec2/src/index.ts
```diff
@@ -40,17 +40,38 @@ export function copySnapshotPresignedUrlMiddleware(options: PreviouslyResolved):
...
-        const region = await options.region();
-        const resolvedEndpoint =
-          typeof options.endpoint === "function" ? await options.endpoint() : toEndpointV1(context.endpointV2!);
+        const destinationRegion = await options.region();
 
-        if (typeof options.regionInfoProvider === "function") {
-          const regionInfo = await options.regionInfoProvider(input.SourceRegion);
-          resolvedEndpoint.hostname = regionInfo?.hostname || `ec2.${input.SourceRegion}.amazonaws.com`;
-        } else {
-          resolvedEndpoint.hostname = `ec2.${input.SourceRegion}.amazonaws.com`;
```
kuhe, line 52: with regionInfoProvider gone as of endpoints v2, all calls defaulted to this else block.

This domain only works for commercial regions, so the fix is to use the endpoints v2 resolver on both the presigned URL with the source region and the base request with the target region.

siddsriv (approved):

Issue #4524, opened 2023-03-14
@aws-sdk/client-sfn: field.name.toLowerCase is not a function


timoly:
### Checkboxes for prior research

- [X] I've gone through [Developer Guide](https://docs.aws.amazon.com/sdk-for-javascript/v3/developer-guide) and [API reference](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest)
- [X] I've checked [AWS Forums](https://forums.aws.amazon.com) and [StackOverflow](https://stackoverflow.com/questions/tagged/aws-sdk-js).
- [X] I've searched for [previous similar issues](https://github.com/aws/aws-sdk-js-v3/issues) and didn't find any solution.

### Describe the bug

When using @aws-sdk/client-sfn ListExecutionsCommand throws

### SDK version number

@aws-sdk/client-sfn@3.290.0

### Which JavaScript Runtime is this issue in?

Node.js

### Details of the browser/Node.js/ReactNative version

18.15.0

### Reproduction Steps

```js
const params: ListExecutionsCommandInput = {
    stateMachineArn: '{stateMachineArn}' ,
    statusFilter: 'RUNNING',
  }
const command = new ListExecutionsCommand(params)
const res = await sfnClient.send(command)
```

### Observed Behavior

```console
TypeError: field.name.toLowerCase is not a function
    at Fields.setField (/data/{project}/node_modules/@aws-sdk/protocol-http/dist-cjs/Fields.js:12:33)
    at Object.set (/data/{project}/node_modules/@aws-sdk/protocol-http/dist-cjs/headersProxy.js:15:20)
    at wrappedHeaderMw (/data/{project}/node_modules/@newrelic/aws-sdk/lib/v3/smithy-client.js:56:43)
```

### Expected Behavior

command doesn't throw

ahmednaser94:
I confirm this bug, solved by returning to an older version
<img width="697" alt="image" src="https://user-images.githubusercontent.com/39867467/224969939-7b380cf0-c881-4814-84eb-5ce5086e5a0e.png">

GedByrneSpeedy:
We also had the same issue when using 

`"@aws-sdk/client-sqs": "^3.26.0",`

we got it working by reverted to using 

`"@aws-sdk/client-sqs": "3.26.0",`

bliu13:
[...]

PR #4529, opened 2023-03-14
fix(protocol-http): revert SRA HttpRequest

milesziemer:
### Issue

Fixes https://github.com/aws/aws-sdk-js-v3/issues/4524

### Description
This reverts commit d9d24b0a2561dfc739dfbbf5962f3c046bbc2f03.
A bug was found with this commit when using a Symbol as a key in `request.headers`.

### Testing
CI

### Additional context
Opening as a draft while alternative solutions are explored. If none are found, should revert before 03/15 release.

---


packages/protocol-http/src/Fields.ts
```diff
@@ -1,7 +1,7 @@
-import { Field, FieldOptions } from "./Field";
+import { Field } from "./Field";
```
kuhe, line 1: Since Field/Fields will be unused you can leave in the latest version if you want

packages/protocol-http/src/Field.ts
```diff
@@ -1,18 +1,18 @@
 import { FieldPosition } from "./FieldPosition";
 
-export type FieldOptions = {
+export type FieldOptions = {
```
trivikr, line 3: I think prettier was not run on this file, which introduces space and removes semicolon.
The commit might have been created with `--no-verify`

Can you explicitly run prettier on this file? If using VSCode, it should run automatically if you open the file and click save.

milesziemer: There were a few other files with the same issue, I've pushed the update.

trivikr (approved):

srchase (approved):

PR #4561, opened 2023-03-21
fix(node-http-handler): call socket.setKeepAlive if enabled in http(s)Agent

### LINKED ISSUE #3624: TCP Keepalive probes are not sent when using AWS JS SDK
## Describe the bug
AWS NAT Gateway/VPCE have [350 seconds timeout](https://docs.aws.amazon.com/vpc/latest/userguide/nat-gateway-troubleshooting.html#nat-gateway-troubleshooting-timeout) for idle connections. After 350 seconds the connection will be ungracefully closed by sending a RST packet and become unusable. This is not unique to NAT GW/VPCE, this technique is common in other network appliances. TCP supports Keepalive specification in order to prevent idle connections from being closed. Some AWS SDKs/CLI do support enabling TCP Keepalive e.g. see
- https://sdk.amazonaws.com/java/api/latest/software/amazon/awssdk/http/apache/ApacheHttpClient.Builder.html (look for tcpKeepAlive)
- https://boto3.amazonaws.com/v1/documentation/api/latest/guide/configuration.html (look for tcp_keepalive)
- https://docs.aws.amazon.com/cli/latest/topic/config-vars.html (look for tcp_keepalive)

AWS JS SDK doesn't seem to support this capability. The only mention of TCP Keepalive I found was at the very bottom of [this doc article](https://docs.aws.amazon.com/sdk-for-javascript/v2/developer-guide/node-reusing-connections.html), however when I tested it - it didn't work, the request made through NAT GW that lasted for more than 350 seconds never completes, and remains in hanging state. As a result it is impossible to make a request to a resource that returns response after >350seconds, e.g. it is impossible to synchonously invoke Lambda that takes more than 350 seconds to execute. 

(I'm using Lambda as an example, but this would apply to any other connection that has a potential to remain idle for >350seconds).

## Your environment
EC2 instance with AL2

#### SDK version number
`@aws-sdk/client-lambda@3.87.0`

Nodejs

$ node -v
v14.19.3
$ npm -v
6.14.17

## Steps to reproduce
1. Create a Lambda function1 that sleeps for 340 seconds and then returns 'OK'
2. Create a Lambda function2 that sleeps for 360 seconds and then returns 'OK'
3. Create an EC2 in private subnet, expose it to internet via NAT GW in a public subnet
4. Update OS level TCP Keepalive settings in /etc/sysctl.conf of the EC2 instance to values <2 minutes, e.g. 
image
5. Reboot EC2 instance. 
6. [Test1] Create a simple NodeJS app that uses AWS SDK to invoke function1, run it 
7. [Test2] Create a simple NodeJS app that uses AWS SDK to invoke function2, run it 

#### Lambda function code (server side)
```JavaScript
// const waitForSeconds = 1;
// const waitForSeconds = 10;
// const waitForSeconds = 30;
// const waitForSeconds = 70; // To test auto-retry after 60 seconds
// const waitForSeconds = 2*60;
// const waitForSeconds = 5*60; // 300 
const waitForSeconds = 6*60; // 660
// const waitForSeconds = 7*60;
// const waitForSeconds = 10*60;
// const waitForSeconds = 12*60;
// const waitForSeconds = (15*60)-10;

exports.handler = async (event, context) => {
    await new Promise(async (resolve, reject)=> {
        for (let i=0; i<waitForSeconds; i++){
            console.log({i});
            await new Promise((resolve, reject)=>setTimeout(resolve, 1000));
        }
        resolve();
    })
    return 'Hello from F1';
}
```

#### Application code (client side)
```JavaScript
const { LambdaClient, InvokeCommand } = require('@aws-sdk/client-lambda');
const { NodeHttpHandler} = require('@aws-sdk/node-http-handler');
const { Agent } = require('http');


const agent = new Agent({
	keepAlive: true
});

const lambdaClient = new LambdaClient({
	requestHandler: new NodeHttpHandler({
		httpAgent: agent
	})
});

const invokeParams = {
	FunctionName: 'f1',
	Payload: ''
};
const command = new InvokeCommand(invokeParams);
lambdaClient.send(command).then((data)=>{
	console.log(data);
}, (error)=>{
	console.error(error);
});
```

#### Observed behavior
1. Test1 - after 340 seconds: Lambda function1 execution completes successfully; AWS SDK request completes successfully. 
2. Test2 - after 360 seconds: Lambda function 2 exection completes successfully; AWS SDK request remains in a waiting/hanging state and never completes with either success/failure (I waited for several minutes, maybe there's a timeout after longer period of time).

#### Expected behavior
1. Test2 - after 360 seconds invocation completes successfully. 

## Additional context
I've created internal write up regarding addressing this problem with other clients - AWS Java SDK, Python SDK, CLI. Please Slack me internally for the link (@antonaws).

yenfryherrerafeliz:
When using node-http-handler and setting keep-alive to true in the agent configuration, this setup is not being honored by nodejs https library, due to a bug with the https lib from nodejs. To aid this issue at the SDK side what I am doing is enabling keep alive directly in the socket, by adding a listener for when a new socket is created, and there we provide the keep-alive settings to the socket.

### Issue
Fixes #3624 

### Description
what I do is to add a listener for when a new socket is created then we pass the keep alive settings to that socket.
Example:
```js
import { NodeHttpHandler }  from '@aws-sdk/node-http-handler';
import {InvokeCommand, LambdaClient} from "@aws-sdk/client-lambda";

const client = new LambdaClient({
    region: 'us-east-2',
    requestHandler: new NodeHttpHandler({
        httpsAgent: new Agent({
            keepAlive: true,
            keepAliveMsecs: 5 * 1000
        })
    })
})

```
And the configuration above will make the socket to send keep-alive packets every 5 seconds.

### Testing
N/A
### Additional context
I have also created the following [issue](https://github.com/nodejs/node/issues/47137) under nodejs repo.

---


packages/node-http-handler/src/set-socket-keep-alive.ts
```diff
@@ -0,0 +1,18 @@
...
+
+const DEFAULT_INITIAL_DELAY = 10 * 1000;
+
+export interface SocketKeepAliveOptions {
+    keepAlive: boolean;
+    initialDelay: number;
+}
+
+export const setSocketKeepAlive = (request: ClientRequest, socketKeepAlive: SocketKeepAliveOptions) => {
+    if (!socketKeepAlive || socketKeepAlive.keepAlive === false) {
```
kuhe, line 11: this should be `if (!(socketKeepAlive?.keepAlive === true))`

strict type checks should be for the opt-in case

packages/node-http-handler/src/set-socket-keep-alive.ts
```diff
@@ -0,0 +1,18 @@
...
+        return;
+    }
+
+    request.on('socket', (socket) => {
+        socket.setKeepAlive(socketKeepAlive.keepAlive, socketKeepAlive.initialDelay || DEFAULT_INITIAL_DELAY);
```
trivikr, line 16: Looks like `keepAlive` always needs to be true for `initialDelay` to be effective.

Should we simplify option `socketKeepAlive: SocketKeepAliveOptions` to `socketKeepAliveInitialDelay: number`?

packages/node-http-handler/src/node-http-handler.ts
```diff
@@ -7,6 +7,7 @@ import { Agent as hsAgent, request as hsRequest, RequestOptions } from "https";
 import { NODEJS_TIMEOUT_ERROR_CODES } from "./constants";
 import { getTransformedHeaders } from "./get-transformed-headers";
 import { writeRequestBody } from "./write-request-body";
+import {setSocketKeepAlive, SocketKeepAliveOptions} from "./set-socket-keep-alive";
```
kuhe, line 10: `prettier` code formatter should be run on these files

packages/node-http-handler/src/set-socket-keep-alive.ts
yenfryherrerafeliz: Done!

packages/node-http-handler/src/set-socket-keep-alive.ts
yenfryherrerafeliz: What I did was to make the initialDelay parameter optional, so that if is not provided then it falls into the default value.

packages/node-http-handler/src/node-http-handler.ts
yenfryherrerafeliz: Done!

alfaproject:
Is this still a issue? Should we add this work-around in our clients until this is merged, or?

packages/node-http-handler/src/node-http-handler.ts
```diff
@@ -151,7 +156,16 @@ export class NodeHttpHandler implements HttpHandler {
         };
       }
 
+      if (typeof nodeHttpsOptions.agent !== "boolean") {
```
kuhe, line 159: when is the type of `nodeHttpsOptions.agent` boolean?

packages/node-http-handler/src/node-http-handler.ts
```diff
@@ -71,7 +76,7 @@ export class NodeHttpHandler implements HttpHandler {
   }
 
   private resolveDefaultConfig(options?: NodeHttpHandlerOptions | void): ResolvedNodeHttpHandlerConfig {
-    const { requestTimeout, connectionTimeout, socketTimeout, httpAgent, httpsAgent } = options || {};
+    const { requestTimeout, connectionTimeout, socketTimeout, httpAgent, httpsAgent} = options || {};
```
trivikr, line 79: Prettier is still not run on this. Either the prettier plugin is not installed in VSCode, or precommit hook is not being run. Can you manually run prettier?

```console
$ yarn prettier --write packages/node-http-handler/src/node-http-handler.ts
```

packages/node-http-handler/src/set-socket-keep-alive.ts
```diff
@@ -0,0 +1,22 @@
...
+export interface SocketKeepAliveOptions {
+    keepAlive: boolean;
+    keepAliveMsecs?: number;
+}
+
+export const setSocketKeepAlive = (
+    request: ClientRequest,
+    socketKeepAlive: SocketKeepAliveOptions
+) => {
+    if (!(socketKeepAlive?.keepAlive === true)) {
```
trivikr, line 12: Simplify to `!==` check

```suggestion
    if (socketKeepAlive?.keepAlive !== true) {
```

packages/node-http-handler/src/node-http-handler.ts
```diff
@@ -151,7 +156,16 @@ export class NodeHttpHandler implements HttpHandler {
...
       }
 
+      if (typeof nodeHttpsOptions.agent !== "boolean") {
+        // @ts-ignore
+        const keepAliveOptions: SocketKeepAliveOptions = nodeHttpsOptions.agent;
+        setSocketKeepAlive(req, {
+          keepAlive: keepAliveOptions.keepAlive,
+          keepAliveMsecs: keepAliveOptions.keepAliveMsecs
+        });
+      }
```
trivikr, line 166: The variable stored here is httpAgent
```suggestion
        const httpAgent = nodeHttpsOptions.agent;
        setSocketKeepAlive(req, {
          keepAlive: httpAgent.keepAlive,
          keepAliveMsecs: httpAgent.keepAliveMsecs
        });
      }
```

packages/node-http-handler/src/node-http-handler.ts
kuhe:
```js
if (typeof nodeHttpsOptions.agent === "object" && "keepAlive" in nodeHttpsOptions.agent) {
```

packages/node-http-handler/src/set-socket-keep-alive.ts
```diff
@@ -0,0 +1,16 @@
+import { ClientRequest } from "http";
+
+export interface SocketKeepAliveOptions {
+  keepAlive: boolean;
+  keepAliveMsecs?: number;
+}
+
+export const setSocketKeepAlive = (request: ClientRequest, socketKeepAlive: SocketKeepAliveOptions) => {
+  if (socketKeepAlive?.keepAlive !== true) {
```
trivikr, line 9: Optional chaining is not needed as the object is always defined
```suggestion
  if (socketKeepAlive.keepAlive !== true) {
```

packages/node-http-handler/src/set-socket-keep-alive.ts
```diff
@@ -0,0 +1,16 @@
...
+
+export const setSocketKeepAlive = (request: ClientRequest, socketKeepAlive: SocketKeepAliveOptions) => {
+  if (socketKeepAlive.keepAlive !== true) {
+    return;
+  }
+
+  request.on("socket", (socket) => {
+    socket.setKeepAlive(socketKeepAlive.keepAlive, socketKeepAlive.keepAliveMsecs || 0);
+  });
+};
```
trivikr, line 16: Use object destructuring to remove variable `socketKeepAlive`
```suggestion
export const setSocketKeepAlive = (request: ClientRequest, { keepAlive, keepAliveMsecs }: SocketKeepAliveOptions) => {
  if (keepAlive !== true) {
    return;
  }

  request.on("socket", (socket) => {
    socket.setKeepAlive(keepAlive, keepAliveMsecs || 0);
  });
};
```

packages/node-http-handler/src/node-http-handler.ts
```diff
@@ -9,6 +9,7 @@ import { getTransformedHeaders } from "./get-transformed-headers";
 import { setConnectionTimeout } from "./set-connection-timeout";
 import { setSocketTimeout } from "./set-socket-timeout";
 import { writeRequestBody } from "./write-request-body";
+import { setSocketKeepAlive, SocketKeepAliveOptions } from "./set-socket-keep-alive";
```
trivikr, line 12: The SocketKeepAliveOptions import is no longer needed
```suggestion
import { setSocketKeepAlive } from "./set-socket-keep-alive";
```

packages/node-http-handler/src/set-socket-keep-alive.spec.ts
```diff
@@ -0,0 +1,66 @@
...
+
+    expect(socket.isKeepAliveEnabled()).toBeFalsy();
+  });
+
+  it("should not set keepAlive at all when socketKeepAlive is undefined", () => {
+    setSocketKeepAlive(request, undefined);
+    request.emit("socket", socket);
+
+    expect(socket.isKeepAliveEnabled()).toBeFalsy();
+  });
```
trivikr, line 65: Test no longer needed, as options is always defined in our internal API.
```suggestion
```

packages/node-http-handler/src/set-socket-keep-alive.spec.ts
```diff
@@ -0,0 +1,59 @@
+import { setSocketKeepAlive } from "./set-socket-keep-alive";
+import { EventEmitter } from "events";
```
trivikr, line 2: Workspace throws lint error
```suggestion
import { EventEmitter } from "events";

import { setSocketKeepAlive } from "./set-socket-keep-alive";
```

trivikr:
Verified that keepAlive parameter is set on agent

```js
import http from "http";

// Logs false in Node.js 14+
console.log("keepAlive values from default agent:", new http.Agent().keepAlive);

// Logs false in Node.js 14+
console.log(
  "keepAlive from agent with keepAlive=false:",
  new http.Agent({ keepAlive: false }).keepAlive
);

// Logs true in Node.js 14+
console.log(
  "keepAlive from agent with keepAlive=true:",
  new http.Agent({ keepAlive: true }).keepAlive
);
```

packages/node-http-handler/src/set-socket-keep-alive.spec.ts
```diff
@@ -0,0 +1,60 @@
...
+
+  it("should not set keepAlive at all when keepAlive is false", () => {
+    setSocketKeepAlive(request, {
+      keepAlive: false,
+    });
+    request.emit("socket", socket);
+
+    expect(socket.isKeepAliveEnabled()).toBeFalsy();
+  });
+});
```
trivikr, line 60: The internal tests were updated during peer review.
```suggestion
import { EventEmitter } from "events";
import { ClientRequest } from "http";
import { Socket } from "net";

import { setSocketKeepAlive } from "./set-socket-keep-alive";

describe("setSocketKeepAlive", () => {
  let request: ClientRequest;
  let socket: Socket;

  beforeEach(() => {
    request = new EventEmitter() as ClientRequest;
    socket = new Socket();
  });

  it("should set keepAlive to true", () => {
    setSocketKeepAlive(request, { keepAlive: true });

    const setKeepAliveSpy = jest.spyOn(socket, "setKeepAlive");
    request.emit("socket", socket);

    expect(setKeepAliveSpy).toHaveBeenCalled();
    expect(setKeepAliveSpy).toHaveBeenCalledWith(true, 0);
  });

  it("should set keepAlive to true with custom initialDelay", () => {
    const initialDelay = 5 * 1000;
    setSocketKeepAlive(request, { keepAlive: true, keepAliveMsecs: initialDelay });

    const setKeepAliveSpy = jest.spyOn(socket, "setKeepAlive");
    request.emit("socket", socket);

    expect(setKeepAliveSpy).toHaveBeenCalled();
    expect(setKeepAliveSpy).toHaveBeenCalledWith(true, initialDelay);
  });

  it("should not set keepAlive at all when keepAlive is false", () => {
    setSocketKeepAlive(request, { keepAlive: false });

    const setKeepAliveSpy = jest.spyOn(socket, "setKeepAlive");
    request.emit("socket", socket);

    expect(setKeepAliveSpy).not.toHaveBeenCalled();
  });
});

```

trivikr (approved):

packages/node-http-handler/src/node-http-handler.ts
```diff
@@ -151,6 +152,16 @@ export class NodeHttpHandler implements HttpHandler {
         };
       }
 
+      const httpAgent = nodeHttpsOptions.agent;
```
trivikr, line 155: Add link to upstream bug report
```suggestion
      // Workaround for bug in Node.js https://github.com/nodejs/node/issues/47137
      const httpAgent = nodeHttpsOptions.agent;
```

kuhe (approved):

AndrewFossAWS (approved):

Issue #4582, opened 2023-03-27
aws-sdk's AbortSignal can't be given to Response


koSakano:
### Checkboxes for prior research

- [X] I've gone through [Developer Guide](https://docs.aws.amazon.com/sdk-for-javascript/v3/developer-guide) and [API reference](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest)
- [X] I've checked [AWS Forums](https://forums.aws.amazon.com) and [StackOverflow](https://stackoverflow.com/questions/tagged/aws-sdk-js).
- [X] I've searched for [previous similar issues](https://github.com/aws/aws-sdk-js-v3/issues) and didn't find any solution.

### Describe the bug

The `abortController.signal` passed to `client.send` is internally passed to `Respnose`, but an error occurred in that area.
```
const abortController = new AbortController();
const client = new S3Client(clientParams);

const requestPromise = client.send(new CreateBucketCommand(commandParams), {
  abortSignal: abortController.signal,
});
```

### SDK version number

@aws-sdk/ababort-controller@3.296.0 @aws-sdk/client-s3@3.296.0

### Which JavaScript Runtime is this issue in?

Browser

### Details of the browser/Node.js/ReactNative version

Chrome: 111.0.5563.110

### Reproduction Steps

```ts
import { AbortController } from '@aws-sdk/abort-controller';

const controller = new AbortController();
// Type 'AbortSignal' is missing the following properties from type 'AbortSignal': reason, throwIfAborted, addEventListener,  removeEventListener, dispatchEvent
new Response(url, { signal: controller.signal });
```
https://github.com/aws/aws-sdk-js-v3/blob/9152e210c6ec29f34bb070eaf2874039022e6ab7/packages/fetch-http-handler/src/fetch-http-handler.ts#L75

### Observed Behavior

TypeError: Failed to construct 'Request': Failed to read the 'signal' property from 'RequestInit': Failed to convert value to 'AbortSignal'.

### Expected Behavior

Works like a standard AbortController.

RanVaknin:
Hi @koSakano , 

Can confirm this is only happening after bundling for web.

koSakano:
@RanVaknin 
Yes, I can confirm that it probably only happens after bundling for the web.
This did not happen with Node18.

kuhe:
For browsers, don't bundle the AbortController implementation from the `@aws-sdk/abort-controller` package. 

Just use the browser's native AbortController, and it should be fine.

[...]

PR #4591, opened 2023-03-29
docs(abort-controller): add message about browsers

kuhe:
### Issue
https://github.com/aws/aws-sdk-js-v3/issues/4582

### Description
Add message to not use the SDK's AbortController in browsers. Make global AbortController compatible with our interface (type fix)

### Testing
added unit test

### DIFF
--- a/packages/abort-controller/src/AbortController.ts
+++ b/packages/abort-controller/src/AbortController.ts
@@ -2,6 +2,10 @@ import { AbortController as IAbortController } from "@aws-sdk/types";
 
 import { AbortSignal } from "./AbortSignal";
 
+/**
+ * This implementation should not be necessary in browsers.
+ * Use the global native implementation in browsers.
+ */
 export class AbortController implements IAbortController {
   public readonly signal: AbortSignal = new AbortSignal();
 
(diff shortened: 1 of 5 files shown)


packages/abort-controller/src/AbortController.ts
```diff
@@ -2,6 +2,10 @@ import { AbortController as IAbortController } from "@aws-sdk/types";
 
 import { AbortSignal } from "./AbortSignal";
 
+/**
+ * This implementation should not be necessary in browsers.
+ * Use the global native implementation in browsers.
+ */
```
trivikr, line 8:
```suggestion
/**
 * This implementation was added as Node.js didn't support AbortController prior to 15.x
 * Use native implementation in browsers or Node.js >=15.4.0.
 */
```

trivikr (approved):

syall (approved):

packages/types/src/abort.spec.ts
syall, line 1: Not a blocker, but is there a reason why this isn't in a jest test?

syall (approved):

packages/types/src/abort.spec.ts
kuhe: it could be, but there is no runtime component in this package so I thought it's unnecessary

PR #4593, opened 2023-03-30
docs(add-private): add @internal tags to all internal packages

MYoung25:
### Description
Add `@internal` tags to all internal packages.  In some cases this overwrites `@private` which is _not_ officially supported in TSDoc.  There are cases where adding the tag may not be strictly necessary, however we want to make sure any exported item from these packages is not picked up in documentation, or is clearly marked as `@internal` for someone reading the code.

### Testing
`yarn build:all`

---

### DIFF
--- a/packages/middleware-endpoint/src/index.ts
+++ b/packages/middleware-endpoint/src/index.ts
@@ -1,5 +1,20 @@
+/**
+ * @internal
+ */
 export * from "./adaptors";
+/**
+ * @internal
+ */
 export * from "./endpointMiddleware";
+/**
+ * @internal
+ */
 export * from "./getEndpointPlugin";
+/**
+ * @internal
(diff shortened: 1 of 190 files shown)


srchase (approved):

kuhe (approved):

syall (approved):

packages/middleware-endpoint/src/index.ts
```diff
@@ -1,5 +1,20 @@
...
+ * @internal
+ */
 export * from "./getEndpointPlugin";
+/**
+ * @internal
+ */
 export * from "./resolveEndpointConfig";
+/**
+ * @internal
+ */
```
kuhe, line 19: what does this do if there are already tags on the individual components?

PR #4616, opened 2023-04-05
feat(codegen): add serde helper function

kuhe:
### Issue
n/a

### Description
This adds a serde helper called `take`. The anticipated usage is as follows:

current:
```js
  {
    CreationDateTime:
      output.CreationDateTime != null
        ? __expectNonNull(__parseEpochTimestamp(__expectNumber(output.CreationDateTime)))
        : undefined,
    GlobalTableArn: __expectString(output.GlobalTableArn),
    GlobalTableName: __expectString(output.GlobalTableName),
    GlobalTableStatus: __expectString(output.GlobalTableStatus),
    ReplicationGroup:
      output.ReplicationGroup != null
        ? deserializeAws_json1_0ReplicaDescriptionList(output.ReplicationGroup, context)
        : undefined,
  }
```

new:
```js
  take(output, {
    CreationDateTime: [, _ => __expectNonNull(__parseEpochTimestamp(__expectNumber(_)))],
    GlobalTableArn: __expectString,
    GlobalTableName: __expectString, // single fn is equivalent to [,fn,]
    GlobalTableStatus: __expectString,
    ReplicationGroup: [, _ => deserializeAws_json1_0ReplicaDescriptionList(_, context)]
  });
```

The use of tuples as instructions (essentially arguments) is for consistency in reading positional values, and brevity. 
```js
[filter function, mapping function, source key]
```
the default form can be reduced from `[void 0, void 0, void 0]` to `[,,]` and further to `[]`, which is shorter than `void 0`.

### Testing
existing and new unit tests

### DIFF
--- a/packages/smithy-client/src/object-mapping.spec.ts
+++ b/packages/smithy-client/src/object-mapping.spec.ts
@@ -130,6 +130,35 @@ describe("object mapping", () => {
   });
 
   describe("take function", () => {
+    it("will not apply instructions to missing fields", () => {
+      const input = {
+        filteredDefault: null,
+        filteredSupplier: undefined,
+        filteredMapper: void 0,
+        filteredFilter: 43,
+        filteredMapperOnly: null,
+      } as const;
+
+      const output = {} as const;
+
+      const instructions: SourceMappingInstructions = {
+        default: [],
+        filteredDefault: [],
+        supplier: [, () => "x"],
+        filteredSupplier: [, () => "x"],
+        mapper: [, (_) => _ + "x"],
+        filteredMapper: [, (_) => _ + "x"],
+        filter: [(_) => _ === 42],
+        filteredFilter: [(_) => _ === 42],
+        sourceKey: [, , "SOURCE_KEY"],
+        sourceKey2: [, (_) => "mapped" + _, "SOURCE_KEY2"],
+        mapperOnly: (_) => _ + "Only",
+        filteredMapperOnly: (_) => _ + "Only",
+      };
+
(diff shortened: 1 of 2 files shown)


mtdowling:
Would this account for the use of `@jsonName`, or is this only for protocols where the name always matches what's on the wire? Does this belong in smithy-typescript too?

kuhe:
> Does this belong in smithy-typescript too?
Yes, this should migrate over with its module when that happens.

> Would this account for the use of @jsonName, or is this only for protocols where the name always matches what's on the wire?

It should handle name differences between the source and target.

```js
const source = {
  succinct: 'value 1',
  VERBOSE: 'value 2' 
};

const result = take(source, {
  succinct: [], // default instruction
  verbose: [_ => _ === 'value 2', _ => _ + ' append this', 'VERBOSE'] // verbose instruction
});

result == {
  succinct: 'value 1',
  verbose: 'value 2 append this'
}
```

The instruction is a triplet of `[valueFilter, valueMapper, sourceKey]` where the defaults are `[nonNullish, passThrough, same key]`

kuhe:
Applying this for example to Lambda's client, the `npm pack` package size without downlevel types goes from 151kb to 143kb.
The protocol file itself differs in `wc` as follows:

before/after
```
7407  24492 360472 dist-cjs/protocols/Aws_restJson1.js
6393  20720 317862 dist-cjs/protocols/Aws_restJson1.js
```

Example diff: https://github.com/aws/aws-sdk-js-v3/pull/4620

srchase (approved):

trivikr (approved):

PR #4638, opened 2023-04-16
Add ArrayBuffer as an allowed type

Vinlock:
### Description

The SDK outputs errors showing that it will accept an `ArrayBuffer`. When attempted in JavaScript it works fine when passing an `ArrayBuffer`. This will allow TypeScript to be okay with passing an `ArrayBuffer`.

### Testing
How was this change tested?
- Create an `ArrayBuffer` and pass it to the `Body` parameter of the S3 `PutObjectCommand`.

### Additional context

---

### DIFF
--- a/clients/client-s3/src/commands/PutObjectCommand.ts
+++ b/clients/client-s3/src/commands/PutObjectCommand.ts
@@ -34,7 +34,7 @@ export type PutObjectCommandInputType = Omit<PutObjectRequest, "Body"> & {
   /**
    * For *`PutObjectRequest["Body"]`*, see {@link PutObjectRequest.Body}.
    */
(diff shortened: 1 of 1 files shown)


trivikr:
Please create a [bug report](https://github.com/aws/aws-sdk-js-v3/issues/new?assignees=&labels=bug%2Cneeds-triage&template=bug-report.yml&title=TITLE+FOR+BUG+REPORT) which shows the failure in TypeScript.

kuhe:
the change is in generated code which cannot be modifying in this way. 

To increase the accepted types of S3's PutObject, it should instead be made to https://github.com/awslabs/smithy-typescript/blob/main/packages/types/src/streaming-payload/streaming-blob-payload-input-types.ts

PR #4641, opened 2023-04-18
Bugfix: function fromIni not working as intended within a Docker container

HarrisonHemstreet:
### Issue

### Description
When using the `fromIni` function within a Docker container, the user's aws credentials can be sourced incorrectly. Example:
```javascript
this.client = new SNSClient({
      region: "us-west-2",
      credentials: fromIni({
        profile: "default",
        filepath: "~/.aws",
        configFilepath: process.env.AWS_CONFIG_FILEPATH
      })
});
```
Even though the file path is at "~/.aws", when ran in a docker container, the file pathing ends up being wrong.

ALSO: the fromIni function, if not passed a filepath, will assume the file path. The assumed file path, when in Docker, ends up being wrong.

### Testing
1. build a simple javascript app that uses AWS SNS
2. in the app, make sure you connect to AWS SNS using a similar method to the example above
3. start this app in Docker
4. you should see that you cannot connect to AWS SNS
5. now switch to this branch in your docker container and try using your AWS SNS app again
6. you should now see that AWS SNS is working and sending messages properly

### Additional context

---


packages/shared-ini-file-loader/src/slurpFile.ts
```diff
@@ -3,13 +3,18 @@ import { promises as fsPromises } from "fs";
...
+import { execSync } from "child_process";
+
 const filePromisesHash: Record<string, Promise<string>> = {};
 
 interface SlurpFileOptions {
   ignoreCache?: boolean;
 }
 
 export const slurpFile = (path: string, options?: SlurpFileOptions) => {
+  if (execSync("whoami").toString().trim() === "root") {
```
kuhe, line 15: this is OS specific

PR #4644, opened 2023-04-20
test(aws-util-test): create request testing private package

kuhe:
### Description
Create a test helper in a private package.
The test helper accepts matchers and asserts against outgoing requests from smithy clients.

### Testing
This is a testing utility, but the package contains a unit test as an example.


private/aws-util-test/src/requests/test-http-handler.ts
```diff
@@ -0,0 +1,151 @@
...
+    this.check(m.method, request.method);
+
+    throw new TestHttpHandlerSuccess();
+
+    // return {
+    //   response: new HttpResponse({
+    //     statusCode: 200,
+    //     body: Buffer.from(`{ "tests": "ok" }`),
+    //   }),
+    // };
```
trivikr, line 88: nit: remove unused commented code

trivikr (approved):

PR #4649, opened 2023-04-21
feat(credential-provider-args)Specify credentials from argument in defaultProvider of credential-provider-node

huuyafwww:
### Description
Added the ability to specify credentials from arguments in defaultProvider of credential-provider-node.

### Testing
Added tests in the "credential-provider-args" package and in the "credential-provider-node" to support it.
(New tests were created based on the existing "credential-provider-*" tests.

### Additional context
I am currently generating credential providers from defaultProvider via environment variables.
However, I would like to avoid that as it would be an implicit dependency.
I suppose I could place the credentials in "\~/.aws/credentials" as a workaround for the above method.
However, I find that method a bit cumbersome to deal with in some cases.
An example is when you want to run in a container such as Docker and use defaultProvider without installing aws cli.
In such a case, you will need to manually write the settings to "\~/.aws/credentials".
We find it a bit tedious to manually write a configuration file to resolve implicit dependencies.
Therefore, I wanted a way to pass credentials directly from the arguments to defaultProvider, and created this pull request.
Also, [the Python AWS SDK provides a way to pass credential directly](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/core/session.html#boto3.session.Session.resource), so I would expect the JavaScript SDK to provide a similar method.

This pull request will allow you to create a credentials provider as follows.

```js
import { defaultProvider } from '@aws-sdk/credential-provider-node';

const credentialsProvider = defaultProvider({
  accessKeyId,
  secretAccessKey,
  sessionToken,
  expiry: new Date().toISOString(),
});
await credentialsProvider();
```

---


kuhe:
why do you need to call anything if you already have arguments that are a set of credentials?

RanVaknin:
Hi @huuyafwww ,

Since we didn't hear back from you I'll go ahead and close this.

Feel free to raise another PR if needed.
Thanks,
Ran~

Issue #4398, opened 2023-02-03
HeadObjectCommand throws TypeError: Cannot read properties of null (reading 'getReader')


max-lt:
### Checkboxes for prior research

- [X] I've gone through [Developer Guide](https://docs.aws.amazon.com/sdk-for-javascript/v3/developer-guide) and [API reference](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest)
- [X] I've checked [AWS Forums](https://forums.aws.amazon.com) and [StackOverflow](https://stackoverflow.com/questions/tagged/aws-sdk-js).
- [X] I've searched for [previous similar issues](https://github.com/aws/aws-sdk-js-v3/issues) and didn't find any solution.

### Describe the bug

With HeadObjectCommand (and S3Client) 

I got the followinf error: `TypeError: Cannot read properties of null (reading 'getReader')`
Following the stacktrace it's caused by the output body being null

Stacktrace: 
```
collectStream
Object.streamCollector
collectBody
deserializeAws_restXmlHeadObjectCommand
deserialize
```

```javascript
const deserializeAws_restXmlHeadObjectCommand = async (output, context) => {
    if (output.statusCode !== 200 && output.statusCode >= 300) {
        return deserializeAws_restXmlHeadObjectCommandError(output, context);
    }

   // ...

    await collectBody(output.body, context); // <- output.body is null
    return contents;
}
```

```typescript
import { GetObjectCommand, HeadObjectCommand, S3Client } from '@aws-sdk/client-s3';

//...
const s3 = new S3Client(config.s3);

function getObject(key: string) {
  return s3.send(new GetObjectCommand({ Bucket: config.bucket, Key: key }));
}

function headObject(key: string) {
   return s3.send(new HeadObjectCommand({ Bucket: config.bucket, Key: key }))
}
```

### SDK version number

@aws-sdk/package-name@3.262.0

### Which JavaScript Runtime is this issue in?

Browser

### Details of the browser/Node.js/ReactNative version

v18.11.0

### Reproduction Steps

Following the stacktrace it's caused by the output body being null

https://github.com/aws/aws-sdk-js-v3/blob/ceb0ef9f78a2b885c09ec0404d6435d048686fb3/clients/client-s3/src/protocols/Aws_restXml.ts#L5806

```javascript
const deserializeAws_restXmlHeadObjectCommand = async (output, context) => {
    if (output.statusCode !== 200 && output.statusCode >= 300) {
[...]

Issue #4369, opened 2023-01-25
NULL response body incorrectly treated as Uint8Array in `collectBody` method 


limzykenneth:
### Checkboxes for prior research

- [X] I've gone through [Developer Guide](https://docs.aws.amazon.com/sdk-for-javascript/v3/developer-guide) and [API reference](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest)
- [X] I've checked [AWS Forums](https://forums.aws.amazon.com) and [StackOverflow](https://stackoverflow.com/questions/tagged/aws-sdk-js).
- [X] I've searched for [previous similar issues](https://github.com/aws/aws-sdk-js-v3/issues) and didn't find any solution.

### Describe the bug

I'm not very familiar with the overall codebase but I've only track down the offending line so sorry in advanced for the lack of details below.

While trying to use the s3 client in a Cloudflare Workers environment with the `DeleteObjectCommand` I'm getting the following error message: `TypeError: stream.getReader is not a function` with an eventual trace to [this function](https://github.com/aws/aws-sdk-js-v3/blob/main/clients/client-s3/src/protocols/Aws_restXml.ts#L12015), and the problem is that the `streamBody` argument passed here should have defaulted to a blank `UInt8Array` when there is no body but further upstream in the stack trace, the body is set to `null` which does not trigger the default argument value and thus get passed along to `streamCollector` that eventually attempts to call `getReader` on `null`.

### SDK version number

@aws-sdk/client-s3@3.254.0

### Which JavaScript Runtime is this issue in?

Browser

### Details of the browser/Node.js/ReactNative version

Cloudflare Workers, wrangler@2.8.1

### Reproduction Steps

I'm not sure if running in Cloudflare workers are relevant or not in this case.

```
await client.deleteObject({
  Bucket: "my-bucket",
  Key: "filename"
});
```

### Observed Behavior

A more complete stack trace if it helps, although there's no useful line numbers I imagine:
```
TypeError: stream.getReader is not a function
    at collectStream ([...]/dist/worker.js:27440:27)
    at Object.streamCollector ([...]/dist/worker.js:27431:12)
    at collectBody ([...]/dist/worker.js:19480:20)
    at deserializeAws_restXmlDeleteObjectCommand ([...]/dist/worker.js:13067:11)
    at deserialize ([...]/dist/worker.js:20719:16)
    at [...]/dist/worker.js:6766:30
[...]

PR #4705, opened 2023-05-06
fix(fetch-http-handler): check for null fetch response.body

kuhe:
### Issue
https://github.com/aws/aws-sdk-js-v3/issues/4398
https://github.com/aws/aws-sdk-js-v3/issues/4369

### Description
Add null check for `fetch` API response.body. 

### Testing
- [x] add unit test
- [x] manual testing with Chrome Version 113.0.5672.63 (Official Build) (x86_64) 

### Workaround
For affected users' clients, you can apply a code workaround until such time that you upgrade to a version of the AWS SDK for JavaScript v3 that contains this fix.

```ts
    
    const s3 = new S3({
      credentials,
    });

    // this example is on an S3 client. The middlewareStack property is available on all clients.
    // add this middleware to default a fetch response body to an empty Uint8Array, 
    // before making any requests.
    s3.middlewareStack.addRelativeTo(
      (next) => async (args) => {
        const { response } = await next(args);
        if (response.body == null) {
          response.body = new Uint8Array();
        }
        return {
          response,
        };
      },
      {
        name: "nullFetchResponseBodyMiddleware",
        toMiddleware: "deserializerMiddleware",
        relation: "after",
        step: "deserialize",
        override: true,
      }
    );
```

### Additional context
In the API spec for `fetch`'s `Response::body` https://developer.mozilla.org/en-US/docs/Web/API/Response/body,
the body is supposed to be null for responses that have an empty body. 

A note clarifies that this did not happen in browser implementations
> Note: Current browsers don't actually conform to the spec requirement to set the body property to null for responses with no body (for example, responses to [HEAD](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/HEAD) requests, or [204 No Content](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/204) responses).

but with the latest version of Chrome, and possibly other browsers, the spec is now apparently being adhered to. This null check should have been in place originally, but is now necessary.

### DIFF
--- a/packages/fetch-http-handler/src/fetch-http-handler.ts
+++ b/packages/fetch-http-handler/src/fetch-http-handler.ts
@@ -83,7 +83,7 @@ export class FetchHttpHandler implements HttpHandler {
           transformedHeaders[pair[0]] = pair[1];
         }
 
-        const hasReadableStream = response.body !== undefined;
+        const hasReadableStream = response.body != undefined;
 
         // Return the response with buffered body
         if (!hasReadableStream) {
--- a/packages/fetch-http-handler/src/fetch-http-handler.ts
+++ b/packages/fetch-http-handler/src/fetch-http-handler.ts
@@ -83,6 +83,7 @@ export class FetchHttpHandler implements HttpHandler {
           transformedHeaders[pair[0]] = pair[1];
         }
 
+        // Check for undefined as well as null.
         const hasReadableStream = response.body != undefined;
 
         // Return the response with buffered body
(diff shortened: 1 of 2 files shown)


siddsriv (approved):

packages/fetch-http-handler/src/fetch-http-handler.ts
```diff
@@ -83,7 +83,7 @@ export class FetchHttpHandler implements HttpHandler {
           transformedHeaders[pair[0]] = pair[1];
         }
 
-        const hasReadableStream = response.body !== undefined;
+        const hasReadableStream = response.body != undefined;
```
trivikr, line 86: nit: Add a comment that this checks for undefined as well as null.
```suggestion
        // Check for undefined as well as null.
        const hasReadableStream = response.body != undefined;
```

trivikr (approved):

cezarderevlean:
Thank you for providing the workaround.

LorhanSohaky:
When will the fix be made available? @kuhe

jroru:
@LorhanSohaky Looks like it was included in the [v3.329.0 release](https://github.com/aws/aws-sdk-js-v3/releases/tag/v3.329.0)

Issue #4635, opened 2023-04-12
The getAwsChunkedEncodingStream function returns a stream which encodes all streams (can be binary streams) as utf-8 strings


tommn2:
### Checkboxes for prior research

- [X] I've gone through [Developer Guide](https://docs.aws.amazon.com/sdk-for-javascript/v3/developer-guide) and [API reference](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest)
- [X] I've checked [AWS Forums](https://forums.aws.amazon.com) and [StackOverflow](https://stackoverflow.com/questions/tagged/aws-sdk-js).
- [X] I've searched for [previous similar issues](https://github.com/aws/aws-sdk-js-v3/issues) and didn't find any solution.

### Describe the bug

The getAwsChunkedEncodingStream() method is returning a stream that is encoding all streams as utf-8 strings.  Each chunk is being written out with a chunk length that is the correct buffer length but the chunk data is written as a utf-8 string (the default encoding for the Buffer.toString() method).  As a result, the chunk's data length does not equal the length written out at the beginning of the chunk.

Without digging into the code more and looking at everywhere this stream is used, I don't know if there are use-cases where a utf-8 string is desired.  If there are then, at a minimum, the chunk length would be incorrect.

In my case I am uploading a binary file to S3 via a PutObjectCommand object; I have a binary stream which is being incorrectly converted to utf-8 resulting in an error response from S3.  See below for a sample error from CloudTrail.

Here's a suggested fix.
https://github.com/tommn2/aws-sdk-js-v3/commit/8704162760e01c1be2380306249314307083ceb7

### SDK version number

@aws-sdk/client-s3@3.306.0

### Which JavaScript Runtime is this issue in?

Node.js

### Details of the browser/Node.js/ReactNative version

v14.19.0

### Reproduction Steps

Use PutObjectCommand to upload a file to S3 and include the { ChecksumAlgorithm: 'SHA256' } option.  Without the checksum option the upload seems to bypass the getAwsChunkedEncodingStream() stream.

### Observed Behavior

		{
			"eventVersion": "1.08",
			"userIdentity": {
				"type": "IAMUser",
				"principalId": "<redacted>",
				"arn": "arn:aws:iam::<redacted>:user/<redacted>",
				"accountId": "<redacted>",
				"accessKeyId": "<redacted>",
				"userName": "<redacted>"
			},
			"eventTime": "2023-04-06T20:18:53Z",
[...]

PR #4711, opened 2023-05-09
fix(util-stream): do not convert stream data to string in AwsChunkedEncodingStream

kuhe:
### Issue
https://github.com/aws/aws-sdk-js-v3/issues/4635

### Description
Take the suggestion from the issue, changing the `getAwsChunkedEncodingStream` push 

### Testing
- [x] cucmber tests for S3 (e2e)
- [x] integ tests


packages/middleware-flexible-checksums/src/middleware-flexible-checksums.integ.spec.ts
```diff
@@ -80,5 +82,35 @@ describe("middleware-flexible-checksums", () => {
...
+          "content-type": "application/octet-stream",
+          "x-amz-content-sha256": "STREAMING-UNSIGNED-PAYLOAD-TRAILER",
+          "content-length": /undefined/,
+        },
+        query: {
+          "x-id": "PutObject",
+        },
+      });
+
+      const stream = fs.createReadStream(path.join(__dirname, "..", "test", "green.jpeg"));
```
pose, line 104: (minor) Could this be instead be achieved by a base64-encoded representation of an image that could fit the file instead of an additional external resource?

pose (approved):

trivikr (approved):

AndrewFossAWS (approved):

packages/middleware-flexible-checksums/src/middleware-flexible-checksums.integ.spec.ts
kuhe: I'll look into doing this in a revision, I have to fix some more tests anyway.

Issue #4754, opened 2023-05-24
When SDK exhausts retry tokens any API call, it does not make any subsequent retries for other calls


trivikr:
### Checkboxes for prior research

- [X] I've gone through [Developer Guide](https://docs.aws.amazon.com/sdk-for-javascript/v3/developer-guide) and [API reference](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest)
- [X] I've checked [AWS Forums](https://forums.aws.amazon.com) and [StackOverflow](https://stackoverflow.com/questions/tagged/aws-sdk-js).
- [X] I've searched for [previous similar issues](https://github.com/aws/aws-sdk-js-v3/issues) and didn't find any solution.

### Describe the bug

When SDK exhausts retry tokens any API call, it does not make any subsequent retries for other calls

### SDK version number

All, used v3.335.0 in testing

### Which JavaScript Runtime is this issue in?

Node.js

### Details of the browser/Node.js/ReactNative version

All, used v16.20.0 in testing

### Reproduction Steps

Here is a repro which adds custom Handler which returns a Timeout for every call, and a retryStrategy with just 100ms delay between retries.

```js
import { Kinesis } from "@aws-sdk/client-kinesis"; // v3.335.0
import { NodeHttp2Handler } from "@aws-sdk/node-http-handler";
import { ConfiguredRetryStrategy } from "@aws-sdk/util-retry";

// Simlulate a timeout error.
class NodeHttp2HandlerReturnsTimeout extends NodeHttp2Handler {
  async handle(request, options) {
    const timeoutError = new Error("Request timed out");
    timeoutError.name = "TimeoutError";
    throw timeoutError;
  }
}

const client = new Kinesis({
  region: "us-west-2",
  requestHandler: new NodeHttp2HandlerReturnsTimeout(),
  retryStrategy: new ConfiguredRetryStrategy(
    2, // max attempts.
    () => 100 // delay only for 100ms between retries to speed up client side rate limiting.
  ),
});

let calls = 0;
let attempts = 0;
while (true) {
  try {
    await client.listStreams({});
  } catch (error) {
    const currentCallAttempts = error.$metadata.attempts;
    
    calls += 1;
    attempts += currentCallAttempts;
    console.log(`Total: ${calls} calls, ${attempts} attempts`);

    if (error.$metadata.attempts === 1) {
      // No more retries can be made.
      break;
    }
  }
}
```

[...]

PR #4755, opened 2023-05-24
fix(util-retry): use different retry token instances per request

kuhe:
### Issue
https://github.com/aws/aws-sdk-js-v3/issues/4754

### Description
Use different instances of the retry token when new requests are made from the same client. They will however share the token availability if coming from the same client and RetryStrategy instance.

Currently, because the retry token object and its closure persist between calls of the same client using the same StandardRetryStrategy instance, the retry count stays at its max value.

### Testing
```ts
import { HttpResponse } from "@aws-sdk/protocol-http";
import { ConfiguredRetryStrategy } from "@aws-sdk/util-retry";

class MockRequestHandler {
  // Returns 429 error to cause a retry in the client.
  // No network request is made.
  async handle() {
    console.log("    Request (mock) responding with 429 throttling error.");
    return {
      response: new HttpResponse({
        statusCode: 429,
        body: Buffer.from(""),
      }),
    };
  }
}

let n = 0;
console.log("Time:", n++);
const i = setInterval(() => {
  console.log("Time:", n++); // print a time marker to the console to show time between retries.
}, 1000);

// initialize client with customized retry strategy and mock request handler.
const s3 = new S3({
  retryStrategy: new ConfiguredRetryStrategy(3, (_: number) => _ * 1000),
  requestHandler: new MockRequestHandler(),
});

const go = () =>
  s3
    .listBuckets({})
    .then((r) => console.log(r.$metadata))
    .catch((r) => console.error(r.$metadata))
    .then(async () => {
      console.log("Retries exhausted.");
    });

Promise.all([go(), go(), go()]).then(() => clearInterval(i));


// with the bug, multiple requests from the same instance will exhaust each others' retry counts
```


packages/util-retry/src/StandardRetryStrategy.ts
```diff
@@ -20,6 +20,7 @@ export class StandardRetryStrategy implements RetryStrategyV2 {
   }
 
   public async acquireInitialRetryToken(retryTokenScope: string): Promise<StandardRetryToken> {
+    this.retryToken.resetRetryCount();
     return this.retryToken;
```
kuhe, line 24: I considered returning a new default token here, but that would reset the retry token count of 500.

siddsriv (approved):

packages/util-retry/src/StandardRetryStrategy.ts
kuhe: this now returns a new default token but with a pointer to the shared token availability counter.

siddsriv (approved):

packages/util-retry/src/StandardRetryStrategy.ts
```diff
@@ -9,18 +9,17 @@ import { getDefaultRetryToken } from "./defaultRetryToken";
...
   constructor(maxAttempts: number);
   constructor(maxAttemptsProvider: Provider<number>);
   constructor(private readonly maxAttempts: number | Provider<number>) {
-    this.retryToken = getDefaultRetryToken(INITIAL_RETRY_TOKENS, DEFAULT_RETRY_DELAY_BASE);
     this.maxAttemptsProvider = typeof maxAttempts === "function" ? maxAttempts : async () => maxAttempts;
   }
 
   public async acquireInitialRetryToken(retryTokenScope: string): Promise<StandardRetryToken> {
-    return this.retryToken;
+    return getDefaultRetryToken(this.availableCapacityRef, DEFAULT_RETRY_DELAY_BASE);
```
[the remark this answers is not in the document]
kuhe: Tokens exist per request, but need access to a shared capacity value, that is why the number is placed in a container.
At least, that is what I think the intended functionality is. 

That does raise the question for me, of what is supposed to happen under intentional high volume with one client. If you have a 99.9xx% success rate under high load, even a small and expected failure rate will exhaust the default retry tokens.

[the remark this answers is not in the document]
trivikr: Discussions happened offline.

We need a reference to available capacity at StandardRetryStrategy level which need to read and updated in RetryToken:
* Read in `hasRetryTokens`
* Updated (value reduced) in `getRetryTokenCount`
* Updated (value increased) in `releaseRetryTokens`

packages/util-retry/src/defaultRetryToken.ts
```diff
@@ -1,91 +1,53 @@
...
   MAXIMUM_RETRY_DELAY,
   NO_RETRY_INCREMENT,
   RETRY_COST,
-  THROTTLING_RETRY_DELAY_BASE,
   TIMEOUT_RETRY_COST,
 } from "./constants";
-import { getDefaultRetryBackoffStrategy } from "./defaultRetryBackoffStrategy";
-
-/**
- * @public
```
kuhe, line 14: this is marked public but is only used as the input type of an internal function, so I suggest we remove it

packages/util-retry/src/StandardRetryStrategy.ts
```diff
@@ -1,47 +1,79 @@
...
+        lastRetryCost: capacityCost,
+      });
     }
+
     throw new Error("No retry token available");
   }
 
   public recordSuccess(token: StandardRetryToken): void {
-    this.retryToken.releaseRetryTokens(token.getLastRetryCost());
+    token.releaseRetryTokens(token.getLastRetryCost());
```
trivikr, line 73: Discussed offline: The tokens will be released in StandardRetryStrategy, and we're marking `releaseRetryTokens` as deprecated.

trivikr:
In the offline discussion, we decided to deprecate the following functions from `StandardRetryToken`
* `hasRetryTokens`
* `getRetryTokenCount`
* `releaseRetryTokens`

They have access to availableCapacity, and can modify it.
Anything which modifies availableCapacity should be in retryStrategy.

packages/util-retry/src/defaultRetryToken.ts
```diff
@@ -1,91 +1,47 @@
...
-  const timeoutRetryCost = options?.timeoutRetryCost ?? TIMEOUT_RETRY_COST;
-  const retryBackoffStrategy = options?.retryBackoffStrategy ?? getDefaultRetryBackoffStrategy();
-
-  let availableCapacity = initialRetryTokens;
-  let retryDelay = Math.min(MAXIMUM_RETRY_DELAY, initialRetryDelay);
-  let lastRetryCost: number | undefined = undefined;
-  let retryCount = initialRetryCount ?? 0;
-
-  const getCapacityAmount = (errorType: RetryErrorType) => (errorType === "TRANSIENT" ? timeoutRetryCost : retryCost);
+export const createDefaultRetryToken = ({
```
trivikr, line 8: Throw errors in implementations of deprecated functions instead?

That way you don't need to pass `availableCapacity` for creating default RetryToken.

trivikr: I think we should remove the functions altogether from the interface.

It's very unlikely that someone would have written subclass of StandardRetryStrategy which doesn't used methods from parent class. And if they do use methods from parent class, they'll unlikely be calling functions on StandardRetryToken which modifies availableCapacity.

trivikr: We're already removing retryToken member from `StandardRetryStrategy` class, as it should not exist.
The only way to access retryToken is by calling `acquireInitialRetryToken` which is done by callee of the strategy, i.e. middleware.

packages/util-retry/src/defaultRetryToken.ts
```diff
@@ -1,97 +1,27 @@
...
-  let retryCount = initialRetryCount ?? 0;
-
-  const getCapacityAmount = (errorType: RetryErrorType) => (errorType === "TRANSIENT" ? timeoutRetryCost : retryCost);
-
+export const createDefaultRetryToken = ({
+  retryDelay,
+  retryCount,
+  lastRetryCost,
+}: {
+  availableCapacity: number;
```
trivikr, line 13: Type can be removed, as `availableCapacity` is no longer passed
```suggestion
```

packages/util-retry/src/StandardRetryStrategy.ts
```diff
@@ -1,47 +1,89 @@
...
   constructor(maxAttemptsProvider: Provider<number>);
   constructor(private readonly maxAttempts: number | Provider<number>) {
-    this.retryToken = getDefaultRetryToken(INITIAL_RETRY_TOKENS, DEFAULT_RETRY_DELAY_BASE);
     this.maxAttemptsProvider = typeof maxAttempts === "function" ? maxAttempts : async () => maxAttempts;
   }
 
   public async acquireInitialRetryToken(retryTokenScope: string): Promise<StandardRetryToken> {
-    return this.retryToken;
+    return createDefaultRetryToken({
+      availableCapacity: this.capacity,
```
trivikr, line 32: availableCapacity is no longer required to be passed.
```suggestion
```

packages/util-retry/src/StandardRetryStrategy.ts
```diff
@@ -1,47 +1,89 @@
...
+      let retryDelay: number;
+      if (errorInfo.retryAfterHint) {
+        const delayFromRetryAfterHint = errorInfo.retryAfterHint.getTime() - Date.now();
+        retryDelay = Math.max(delayFromRetryAfterHint || 0, delayFromErrorType);
+      } else {
+        retryDelay = delayFromErrorType;
+      }
+      this.capacity -= capacityCost;
+      return createDefaultRetryToken({
+        availableCapacity: this.capacity,
```
trivikr, line 63: availableCapacity is no longer required to be passed
```suggestion
```

trivikr:
Verified the fix by printing `capacityAmount` with the following code which calls two operations for two clients.

```js
import {
  S3Client,
  ListBucketsCommand,
  ListObjectsCommand,
} from "../aws-sdk-js-v3/clients/client-s3/dist-cjs/index.js";
import {
  DynamoDBClient,
  ListTablesCommand,
  ListBackupsCommand,
} from "../aws-sdk-js-v3/clients/client-dynamodb/dist-cjs/index.js";
import { NodeHttpHandler } from "../aws-sdk-js-v3/packages/node-http-handler/dist-cjs/index.js";
import { ConfiguredRetryStrategy } from "../aws-sdk-js-v3/packages/util-retry/dist-cjs/index.js";

// Simulate a timeout error.
class NodeHttpHandlerReturnsTimeout extends NodeHttpHandler {
  async handle(request, options) {
    const timeoutError = new Error("Request timed out");
    timeoutError.name = "TimeoutError";
    throw timeoutError;
  }
}

const s3Client = new S3Client({
  region: "us-west-2",
  requestHandler: new NodeHttpHandlerReturnsTimeout(),
  retryStrategy: new ConfiguredRetryStrategy(2, () => 100),
});

const dynamodbClient = new DynamoDBClient({
  region: "us-west-2",
  requestHandler: new NodeHttpHandlerReturnsTimeout(),
  retryStrategy: new ConfiguredRetryStrategy(3, () => 50),
});

const calls = {};

const callRequest = async (client, command, operation) => {
  if (!calls[operation]) {
    calls[operation] = 0;
  }

  while (true) {
    try {
      await client.send(command);
    } catch (error) {
      const metadata = error.$metadata;
      calls[operation] = calls[operation] + 1;
      console.log(
        `[${operation}]: Call #${calls[operation]}, Attempts: ${metadata.attempts}, Retry delay: ${metadata.totalRetryDelay}ms`
      );

      if (calls[operation] === 3) {
        break;
      }
    }
  }
};

await Promise.all([
  callRequest(s3Client, new ListBucketsCommand({}), "s3:listBuckets"),
  callRequest(
    s3Client,
    new ListObjectsCommand({ Bucket: "foo" }),
    "s3:listObjects"
  ),
  callRequest(dynamodbClient, new ListTablesCommand({}), "ddb:listTables"),
  callRequest(dynamodbClient, new ListBackupsCommand({}), "ddb:listBackups"),
]);

```

trivikr (approved):

packages/util-retry/src/StandardRetryStrategy.ts
```diff
@@ -1,47 +1,88 @@
...
+import { createDefaultRetryToken } from "./defaultRetryToken";
 
 /**
  * @public
  */
 export class StandardRetryStrategy implements RetryStrategyV2 {
   public readonly mode: string = RETRY_MODES.STANDARD;
-  private retryToken: StandardRetryToken;
+  private capacity: number = INITIAL_RETRY_TOKENS;
+  private readonly retryBackoffStrategy = getDefaultRetryBackoffStrategy();
```
trivikr, line 21: nit: The `retryBackoffStrategy` is not required to be a class member, as it's only used in refreshRetryTokenForRetry
```suggestion
  private readonly retryBackoffStrategy = getDefaultRetryBackoffStrategy();
```

packages/util-retry/src/StandardRetryStrategy.ts
```diff
@@ -1,47 +1,88 @@
...
+      const delayBase = errorType === "THROTTLING" ? THROTTLING_RETRY_DELAY_BASE : DEFAULT_RETRY_DELAY_BASE;
+      this.retryBackoffStrategy.setDelayBase(delayBase);
+      const delayFromErrorType = this.retryBackoffStrategy.computeNextBackoffDelay(token.getRetryCount());
+      let retryDelay: number;
+      if (errorInfo.retryAfterHint) {
+        const delayFromRetryAfterHint = errorInfo.retryAfterHint.getTime() - Date.now();
+        retryDelay = Math.max(delayFromRetryAfterHint || 0, delayFromErrorType);
+      } else {
+        retryDelay = delayFromErrorType;
+      }
```
trivikr, line 59: retryDelay can be changed to a const and moved to it's own section for better readability
```suggestion

      const delayFromErrorType = this.retryBackoffStrategy.computeNextBackoffDelay(token.getRetryCount());
      const retryDelay = errorInfo.retryAfterHint
        ? Math.max(errorInfo.retryAfterHint.getTime() - Date.now() || 0, delayFromErrorType)
        : delayFromErrorType;

```

packages/util-retry/src/StandardRetryStrategy.ts
```diff
@@ -1,47 +1,88 @@
...
-      tokenToRenew.getRetryTokenCount(errorInfo);
-      return tokenToRenew;
+    const getCapacityCost = (errorType: RetryErrorType) =>
+      errorType === "TRANSIENT" ? TIMEOUT_RETRY_COST : RETRY_COST;
+
+    if (this.shouldRetry(token, errorInfo, maxAttempts)) {
+      const errorType = errorInfo.errorType;
+      const capacityCost = getCapacityCost(errorType);
+      const delayBase = errorType === "THROTTLING" ? THROTTLING_RETRY_DELAY_BASE : DEFAULT_RETRY_DELAY_BASE;
+      this.retryBackoffStrategy.setDelayBase(delayBase);
```
trivikr, line 51: nit: delayBase variable can be removed as it's used only once
```suggestion

      this.retryBackoffStrategy.setDelayBase(
        errorType === "THROTTLING" ? THROTTLING_RETRY_DELAY_BASE : DEFAULT_RETRY_DELAY_BASE
      );
```

packages/util-retry/src/StandardRetryStrategy.ts
```diff
@@ -1,47 +1,88 @@
...
 
-    if (this.shouldRetry(tokenToRenew, errorInfo, maxAttempts)) {
-      tokenToRenew.getRetryTokenCount(errorInfo);
-      return tokenToRenew;
+    const getCapacityCost = (errorType: RetryErrorType) =>
+      errorType === "TRANSIENT" ? TIMEOUT_RETRY_COST : RETRY_COST;
+
+    if (this.shouldRetry(token, errorInfo, maxAttempts)) {
+      const errorType = errorInfo.errorType;
+      const capacityCost = getCapacityCost(errorType);
```
trivikr, line 49: nit: move capacityCost closest to it's usage, i.e.before line 60?

packages/util-retry/src/StandardRetryStrategy.ts
kuhe: we don't need to regenerate the object every method call though

packages/util-retry/src/StandardRetryStrategy.ts
```diff
@@ -1,47 +1,86 @@
...
+    token: StandardRetryToken,
     errorInfo: RetryErrorInfo
   ): Promise<StandardRetryToken> {
     const maxAttempts = await this.getMaxAttempts();
 
-    if (this.shouldRetry(tokenToRenew, errorInfo, maxAttempts)) {
-      tokenToRenew.getRetryTokenCount(errorInfo);
-      return tokenToRenew;
+    const getCapacityCost = (errorType: RetryErrorType) =>
+      errorType === "TRANSIENT" ? TIMEOUT_RETRY_COST : RETRY_COST;
```
trivikr, line 45: The function `getCapacityCost` inside refreshRetryTokenForRetry is same as `getCapacityAmount` in shouldRetry

Move it to a private function inside class?

trivikr (approved):

AndrewFossAWS (approved):

PR #4787, opened 2023-06-02
feat(util-stream): implement aws-chunked-body

milesziemer:
### Issue

### Description

Implements the getAwsChunkedBody function which converts an
AsyncIterable to the aws-chunked-body encoding stream format,
and will replace the getAwsChunkedEncodingStream function.
The method itself is runtime agnostic, but utility functions
readableToAsyncIterable and asyncIterableToReadable were added
to convert between AsyncIterable and the runtime stream type.
GetAwsChunkedEncodingStream was deprecated.

middlware-flexible-checksum, the only middleware to use the old
implementation was updated to allow callers to optionally provide
an implementation of GetAwsChunkedBody, which will be used instead
of the required GetAwsChunkedEncodingStream implementation.

The code generator was updated to use the util-stream package and
generate the getAwsChunkedBody config property when necessary,
and all the clients were regenerated.

A deprecation message was added to the readmes of util-stream-node
and util-stream-browser.

### Testing
How was this change tested?

- Tests were added to middleware-flexible-checksum to make sure the
GetAwsChunkedBody is used when present, and
GetAwsChunkedEncodingStream is used otherwise. Tests were also added
to util-stream to test the functionality of getAwsChunkedBody. Karma
was added to the package so getAwsChunkedBody can be tested in Node
and browser runtimes.

- `yarn test:protocols` and `yarn test:integration:legacy`

- A test was set up using the aws-sdk-js-tests repo to call S3's GetObject
using the updated code, to verify it works in both node and browser runtimes.

### Additional context

This PR requires the codegen change in https://github.com/awslabs/smithy-typescript/pull/775,
so CI will fail until it is merged. This PR will be converted from a draft when the smithy-typescript
PR is merged.

---


packages/util-stream/src/getAwsChunkedBody.spec.ts
```diff
@@ -0,0 +1,106 @@
...
+  describe("skips checksum computation", () => {
+    const validateStreamWithoutChecksum = (readable: Readable, done: Function) => {
+      let buffer = "";
+      readable.on("data", (data) => {
+        buffer += data.toString();
+      });
+      readable.on("end", () => {
+        expect(mockStreamHasher).not.toHaveBeenCalled();
+        expect(mockBase64Encoder).not.toHaveBeenCalled();
+        expect(buffer).toEqual(`5\r\nHello\r\n5\r\nWorld\r\n0\r\n`);
```
kuhe, line 59: for strings like these you could write it like

```js
[
  '5',
  'Hello',
  '5',
  'World',
  '0',
].join('\r\n') + '\r\n'
```

to make them easier to read and edit, or break the `\n` into actual new lines in the template string

packages/util-stream/src/readableToAsyncIterable.ts
```diff
@@ -0,0 +1,42 @@
...
+ * Convert object stream piped in into an async iterable. This
+ * adaptor should be deprecated when Node stream iterator is stable.
+ * Caveat: this adaptor won't have backpressure to inwards stream
+ *
+ * Reference: https://nodejs.org/docs/latest-v11.x/api/stream.html#stream_readable_symbol_asynciterator
+ */
+export async function* readableToAsyncIterable<T>(readStream: Readable): AsyncIterable<T> {
+  let streamEnded = false;
+  let generationEnded = false;
+  const records = new Array<T>();
```
kuhe, line 15:
```suggestion
  const records: T[] = [];
```

packages/util-stream/src/readableToAsyncIterable.ts
```diff
@@ -0,0 +1,42 @@
...
+    if (!streamEnded) {
+      streamEnded = true;
+    }
+    if (err) {
+      throw err;
+    }
+  });
+
+  readStream.on("data", (data) => {
+    records.push(data);
```
kuhe, line 27: why not yield here?

packages/util-stream/src/readableToAsyncIterable.ts
```diff
@@ -0,0 +1,42 @@
...
+  });
+
+  readStream.on("end", () => {
+    streamEnded = true;
+  });
+
+  while (!generationEnded) {
+    // @ts-ignore TS2345: Argument of type 'T | undefined' is not assignable to type 'T | PromiseLike<T>'.
+    const value = await new Promise<T>((resolve) => setTimeout(() => resolve(records.shift()), 0));
```
kuhe, line 36: this looks like you intend to have the function wait one event loop, can this be... ?

```ts
await 0;
const next = records.shift();
if (next) {
  yield next;
}
```

proof of concept:
```ts
let x = async () => {
  console.log('start');
  await 0;
  console.log('end');
};
x(); x();
```

packages/util-stream/src/readableToAsyncIterable.ts
kuhe: can this function be implemented without this queue?

packages/util-stream/src/readableToAsyncIterable.ts
milesziemer: This works

packages/util-stream/src/readableToAsyncIterable.ts
milesziemer: You can't yield within the closure

packages/util-stream/src/readableToAsyncIterable.ts
milesziemer: I think the reason is the same as https://github.com/aws/aws-sdk-js-v3/pull/4787#discussion_r1221956497, you can't read the stream chunk-by-chunk _and_ yield, but I might be missing something

milesziemer:
Closing in favor of https://github.com/aws/aws-sdk-js-v3/pull/4861. The implementation of getAwsChunkedBody in this PR doesn't work because streamHasher operates on ReadableStream in node runtime. We can move to AsyncIterable in the future, but switching clients to util-stream doesn't have to wait on that.

PR #4850, opened 2023-06-16
chore(packages): consolidate middleware integ tests

srchase:
### Issue

### Description
This PR moves the integration tests from the following packages into a consolidated `aws-middleware-test` package:
* middleware-apply-body-checksum
* middleware-content-length
* middleware-endpoint
* middleware-retry
* middleware-serde

### Testing
Ran `yarn test` from the new package.

### Additional context

---


private/aws-util-test/src/requests/test-http-handler.js
```diff
@@ -0,0 +1,117 @@
+export class TestHttpHandler {
```
kuhe, line 1: do you need to generate this .js and .d.ts file? can they be removed?

private/aws-util-test/src/requests/test-http-handler.d.ts
```diff
@@ -0,0 +1,65 @@
+import { HttpHandler, HttpRequest, HttpResponse } from "@aws-sdk/protocol-http";
```
kuhe, line 1: this and the .js file should not need to be generated

srchase: Those shouldn't have been committed.

private/aws-util-test/src/requests/test-http-handler.js
srchase: Those shouldn't have been committed.

kuhe (approved):

private/aws-middleware-test/package.json
```diff
@@ -0,0 +1,67 @@
...
+  "version": "3.0.0",
+  "scripts": {
+    "build": "exit 0",
+    "build:cjs": "exit 0",
+    "build:es": "exit 0",
+    "build:include:deps": "lerna run --scope $npm_package_name --include-dependencies build",
+    "build:types": "tsc -p tsconfig.types.json",
+    "build:types:downlevel": "downlevel-dts dist-types dist-types/ts3.4",
+    "clean": "rimraf ./dist-* && rimraf *.tsbuildinfo",
+    "test": "jest --coverage --passWithNoTests"
```
kuhe, line 13: these tests need to run in the `test:integration` classification, because they import undeclared fully built clients.

kuhe (changes requested):

kuhe (approved):

PR #4891, opened 2023-06-27
fix(util-retry): correct attempts count on StandardRetryStrategy

srchase:
### Issue

### Description
This PR fixes the attempts count used in `StandardRetryStrategy` in `util-retry`. The number of attempts made is one more than the number of retries. The correct attempts count is then used to compare against `maxAttempts` to determine if retries should be made.

This corrects the behavior of `StandardRetryStrategy` to be congruent with the documentation here: https://docs.aws.amazon.com/sdkref/latest/guide/feature-retry-behavior.html

### Testing
Updated tests to cover case where max attempts is set to 1 which should disable any retries.

### Additional context

---

### DIFF
--- a/packages/util-retry/src/StandardRetryStrategy.spec.ts
+++ b/packages/util-retry/src/StandardRetryStrategy.spec.ts
@@ -22,7 +22,7 @@ describe(StandardRetryStrategy.name, () => {
   });
 
   afterEach(() => {
-    jest.clearAllMocks;
+    jest.clearAllMocks();
   });
 
   it("sets maxAttemptsProvider as a class member variable", () => {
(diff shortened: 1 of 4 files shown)


private/aws-client-retry-test/src/ClientRetryTest.spec.ts
```diff
@@ -89,21 +89,22 @@ describe("util-retry integration tests", () => {
...
+      expect(error.$metadata.attempts).toBe(3);
       expect(error.$metadata.totalRetryDelay).toBeGreaterThan(0);
     }
   });
 
   it("should use a shared capacity for retries", async () => {
     const expectedInitialCapacity = 500;
     const expectedDrainPerAttempt = 5;
-    const expectedRetryAttemptsPerRequest = 7;
+    const expectedRetryAttemptsPerRequest = 6;
```
syall, line 100: nit: create another constant for `maxAttempts` instead of using `expectedRetryAttemptsPerRequest` to calculate it.

packages/util-retry/src/StandardRetryStrategy.spec.ts
```diff
@@ -22,7 +22,7 @@ describe(StandardRetryStrategy.name, () => {
   });
 
   afterEach(() => {
-    jest.clearAllMocks;
+    jest.clearAllMocks();
```
syall, line 25: Nice catch 👍

syall (changes requested):

private/aws-client-retry-test/src/ClientRetryTest.spec.ts
srchase: fixed in 397b004bbb699eb51281598f3c155b61a9eff3ce.

Separating out maxAttempts is much more clear.

syall (approved):

Issue #4933, opened 2023-07-06
create a type alias for http payloads


kuhe:
### Describe the feature

Currently payloads' runtime-acceptable type depends on the request handler implementation. To allow for multiple possible implementations we have a wide http payload type, e.g.

```ts
string | Blob | Uint8Array | Readable | ReadableStream<any> | Buffer | undefined
```

for PutObjectCommandInput.Body.

This is confusing for users, and more-so now that Node.js has made many previously browser-only types available.

### Use Case

improve documentation ease of use for http payload types

### Proposed Solution

Create a type alias for the union of blob types.

The type alias will have attached documentation describing when each of the union types is acceptable or preferred.

Similar to: https://github.com/aws/aws-sdk-js-v3/blob/main/packages/types/src/blob/blob-types.ts

```ts
/**

Write docs here explaining what types are available in what environments (corresponding to request handlers).
 
*/
export type StreamingPayloadTypes = string | Blob | Uint8Array | Readable | ReadableStream<any> | Buffer
```

Codegen the alias to operations with streaming body inputs so that the documentation is easy to access.

### Acknowledgements

- [X] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### SDK version used

*

### Environment details (OS name and version, etc.)

*

kuhe:
Related: https://github.com/aws/aws-sdk-js-v3/issues/4930

kuhe:
related #4973

kuhe:
related request to allow transformation to node stream: https://github.com/aws/aws-sdk-js-v3/issues/4720

kuhe:
PRs: 
https://github.com/awslabs/smithy-typescript/pull/835
https://github.com/awslabs/smithy-typescript/pull/836

TBD: JSv3 codegen

PR #4987, opened 2023-07-19
docs(types): improve streaming payload types

kuhe:
### Issue
closes https://github.com/aws/aws-sdk-js-v3/issues/4933

codegen for https://github.com/awslabs/smithy-typescript/pull/835

### Description
replaces inline union types representing payload blobs with a type alias (equivalent) that centralizes documentation and context for how to use those types.


siddsriv (approved):

kuhe:
- [x] ran integration tests
- [x] ran cucumber e2e tests

scripts/runtime-dependency-version-check/runtime-dep-version-check.js
```diff
@@ -108,3 +108,36 @@ if (errors.length) {
   const violations = errors.join(", ");
   throw new Error(violations + " have inconsistent declared versions.");
 }
+
+for (const pkg of nonClientPackages) {
```
srchase, line 112: The generated protocol test packages in `/private` *should* have their dependencies updated the same way as client packages. Should those get handled separately? They shouldn't ever be out of sync with the clients anyways, but this could potentially paper over an issue where they get updated independently of the right codegen changes being applied to those test packages.

kuhe: I've moved the private entries into the client list so they get checked and contribute to the canonical version that gets written to the rest of the packages.

trivikr (approved):

PR #5028, opened 2023-07-31
Changed statuses case

poph2:
### Issue
aws/aws-sdk#569

### Description
Updated the case for the ApplicationVersionStatuses to match what is being returned by AWS APIs

---


kuhe:
As described in the original issue, this fix needs to come from the service model, it cannot be changed here in generated code.

PR #5049, opened 2023-08-03
fix(token-providers): break dependency cycle with sso-oidc

kuhe:
### Description
client-sso-oidc is used in token-providers. However, that client may use token providers in its default credential chain. This would create a cyclical dependency chain that breaks `yarn build:all` and turbopack.

This PR proposes that a minimal bundle of the parts of the sso-oidc client are included within token-providers for its use. This increases (before minification) the size of the token-providers package by 76K. To compare, the full sso-oidc client is 112K. 

This means a savings of 36K when not using the full sso-oidc client, but a redundancy of 76K when doing so. 

The alternative of making the client a peer dependency would be better, but the loss of a required dependency may be too disruptive of a breaking change. 

### Testing
existing unit tests

### DIFF
--- a/package.json
+++ b/package.json
@@ -80,6 +80,7 @@
     "concurrently": "7.0.0",
     "decomment": "0.9.5",
     "downlevel-dts": "0.10.1",
+    "esbuild": "0.18.17",
     "eslint": "8.36.0",
     "eslint-config-prettier": "8.5.0",
     "eslint-plugin-prettier": "4.0.0",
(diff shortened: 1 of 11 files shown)


kuhe:
@aws-sdk/client-sso-oidc weekly downloads: 8,217,279
@aws-sdk/token-providers weekly downloads: 8,211,169

this implies only 5k direct downloads of client-sso-oidc

trivikr (approved):

Issue #5056, opened 2023-08-07
Undeclared Dependencies in token-providers@3.385.0


zaq42:
### Checkboxes for prior research

- [X] I've gone through [Developer Guide](https://docs.aws.amazon.com/sdk-for-javascript/v3/developer-guide) and [API reference](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest)
- [X] I've checked [AWS Forums](https://forums.aws.amazon.com) and [StackOverflow](https://stackoverflow.com/questions/tagged/aws-sdk-js).
- [X] I've searched for [previous similar issues](https://github.com/aws/aws-sdk-js-v3/issues) and didn't find any solution.

### Describe the bug

After updating to 3.385.0, using Yarn 3.6.1, there are a number of (21) unresolved dependencies for token-providers. 

For example:
```
main: Error: @aws-sdk/token-providers tried to access @aws-sdk/util-user-agent-node, but it isn't declared in its dependencies; this makes the require call ambiguous and unsound.
Required package: @aws-sdk/util-user-agent-node
Required by: @aws-sdk/token-providers@npm:3.385.0 (via ./.yarn/cache/@aws-sdk-token-providers-npm-3.385.0-dd849ebcdb-8ee04c9433.zip/node_modules/@aws-sdk/token-providers/dist-cjs/client-sso-oidc-bundle/)
```

This may be related to https://github.com/aws/aws-sdk-js-v3/issues/5049

Workaround was ultimately the following packageExtension in .yarnrc.yml:

```
packageExtensions:
  "@aws-sdk/token-providers@3.385.0":
    dependencies:
      "@aws-sdk/middleware-host-header": "*"
      "@aws-sdk/middleware-logger": "*"
      "@aws-sdk/middleware-recursion-detection": "*"
      "@aws-sdk/middleware-user-agent": "*"
      "@aws-sdk/util-endpoints": "*"
      "@aws-sdk/util-user-agent-node": "*"
      "@smithy/node-config-provider": "*"
      "@smithy/smithy-client": "*"
      "@smithy/config-resolver": "*"
      "@smithy/hash-node": "*"
      "@smithy/middleware-content-length": "*"
      "@smithy/middleware-endpoint": "*"
      "@smithy/middleware-retry": "*"
      "@smithy/middleware-serde": "*"
      "@smithy/node-http-handler": "*"
      "@smithy/protocol-http": "*"
      "@smithy/url-parser": "*"
[...]

PR #5067, opened 2023-08-08
fix(token-providers): use static version of bundled sso oidc client

kuhe:
### Issue
fixes https://github.com/aws/aws-sdk-js-v3/issues/5056

### Description
use a static version of the client sso oidc bundle, instead of trying to generate it at build time.

### DIFF
--- a/packages/token-providers/package.json
+++ b/packages/token-providers/package.json
@@ -6,8 +6,8 @@
   "module": "./dist-es/index.js",
   "sideEffects": false,
   "scripts": {
-    "build:client-dist": "node ./scripts/esbuild",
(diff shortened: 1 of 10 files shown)


packages/token-providers/package.json
```diff
@@ -6,8 +6,8 @@
   "module": "./dist-es/index.js",
   "sideEffects": false,
   "scripts": {
-    "build:client-dist": "node ./scripts/esbuild",
-    "build": "yarn build:client-dist && concurrently 'yarn:build:cjs' 'yarn:build:es' 'yarn:build:types'",
```
kuhe, line 10: this is the main difference. Instead of trying to build a bundle during the build of this package, the bundle is static. 

While it may fall out of sync with the SSO OIDC client, this is necessary anyway to maintain a minimal version of the client that does not need the credential provider chain.

Issue #4288, opened 2022-12-14
Add `--loglevel` option in `generate-clients` CLI, which passes it to subsequent CLIs.


trivikr:
### Describe the feature

Add `--loglevel` option in `generate-clients` CLI, which passes it to subsequent CLIs.

### Use Case

When running yarn generate-clients, it prints lot of output

On 2022-12-14, when `generate-clients` was run it, the output was 25000 lines as can be viewed in: [generate-clients-output.txt](https://github.com/aws/aws-sdk-js-v3/files/10230688/generate-clients-output.txt)
* First 500 lines are from java codegen.
* Then ~3000 lines are from eslint.
* The rest  21000 lines are from prettier.

### Proposed Solution

Add `--loglevel` option in `generate-clients` CLI. This can be similar to one in prettier

```console
$ prettier --help | grep log     
  --loglevel <silent|error|warn|log|debug>
                           What level of logs to report.
                           Defaults to log.
```

It can default to log, but developer has an option to pass which level they want.
The CLI should pass options to gradle, eslint and prettier, if one exists.

### Acknowledgements

- [X] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### SDK version used

main

### Environment details (OS name and version, etc.)

Node.js 16

PR #5088, opened 2023-08-15
feat(generate-client) Added --loglevel option in generate-clients CLI

NextThread:
#4288 
Added --loglevel option in generate-clients CLI, which passes it to subsequent CLIs


kuhe:
Why is this change needed?

NextThread:
#4288

trivikr:
The loglevel first needs to be introduced in downstream dependencies, and a difference between all logs needs to be provided under testing.

@NextThread Can you first comment on https://github.com/aws/aws-sdk-js-v3/issues/4288 on how different log levels would look like? And the output will look like?

PR #5106, opened 2023-08-18
feat(rekognitionstreaming): add oval fit timeout configuration option

kuhe:
add model update and codegen for manually managed service


codegen/sdk-codegen/aws-models/rekognitionstreaming.json
```diff
@@ -152,6 +152,9 @@
           "traits": {
             "smithy.api#documentation": "<p> Threshold for height during face IOU (Intersection over Union) match.</p>"
           }
+        },
+        "OvalFitTimeout": {
+          "target": "com.amazonaws.rekognitionstreaming#TimeoutInMs"
```
kuhe, line 157: This member needs documentation trait. Users cannot see that it targets something called `TimeoutInMs`.

kuhe: The max of 10000 is also not enforced nor visible as a trait.

Issue #5125, opened 2023-08-22
getSignedUrl with ChecksumMode ENABLED Fails


iheffernan:
### Checkboxes for prior research

- [X] I've gone through [Developer Guide](https://docs.aws.amazon.com/sdk-for-javascript/v3/developer-guide) and [API reference](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest)
- [X] I've checked [AWS Forums](https://forums.aws.amazon.com) and [StackOverflow](https://stackoverflow.com/questions/tagged/aws-sdk-js).
- [X] I've searched for [previous similar issues](https://github.com/aws/aws-sdk-js-v3/issues) and didn't find any solution.

### Describe the bug

When creating an S3 presigned URL for a GET with `ChecksumMode: 'ENABLED'` an error is thrown.

Error is 
```
Cannot read properties of undefined (reading 'x-amz-checksum-crc32')
TypeError: Cannot read properties of undefined (reading 'x-amz-checksum-crc32')
    at validateChecksumFromResponse (/Users/ian/WebstormProjects/byrst-web/node_modules/@aws-sdk/middleware-flexible-checksums/dist-cjs/validateChecksumFromResponse.js:13:53)
    at /Users/ian/WebstormProjects/byrst-web/node_modules/@aws-sdk/middleware-flexible-checksums/dist-cjs/flexibleChecksumsMiddleware.js:68:79
    at processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async /Users/ian/WebstormProjects/byrst-web/node_modules/@aws-sdk/middleware-logger/dist-cjs/loggerMiddleware.js:7:26
    at async getSignedUrl (/Users/ian/WebstormProjects/byrst-web/node_modules/@aws-sdk/s3-request-presigner/dist-cjs/getSignedUrl.js:54:24)
```

### SDK version number

@aws-sdk/s3-request-presigner@3.391.0

### Which JavaScript Runtime is this issue in?

Node.js

### Details of the browser/Node.js/ReactNative version

Node.js v18.17.1

### Reproduction Steps

```
const s3Client = new S3Client();
const command = new GetObjectCommand({
    Bucket: 'my-bucket',
    Key: 'test-key.txt',
    ChecksumMode: "ENABLED"
});
const url = await getSignedUrl(s3Client, command, {
    expiresIn: 10,
    unhoistableHeaders: new Set(['x-amz-checksum-mode'])
});
console.log(url);
```

[...]

PR #5126, opened 2023-08-22
fix(middleware-flexible-checksums): split stream when validating response checksum

kuhe:
### Issue
fixes https://github.com/aws/aws-sdk-js-v3/issues/5125

### Description
If checksum validation of the response is requested but the response is a stream, we fully gather the stream before passing it to the checksum validator and then the user. 

This has performance downsides due to collecting the stream in memory. However, if the user is opting into a checksum validation of the response, do we have an alternative? I'm not entirely certain after reading the internal spec for this.

### Alternative
revert the await from [3.383.0](https://github.com/aws/aws-sdk-js-v3/pull/5043), but that creates a potentially uncatchable error, which may crash user applications.

### Testing
update unit tests

manual test
```js
import { GetObjectCommand, S3 } from "@aws-sdk/client-s3";

const s3Client = new S3({
  region: "us-west-2",
});
const bucket = "......-us-west-2";
const key = "checksum.txt";

const putResult = await s3Client.putObject({
  Bucket: bucket,
  Key: key,
  Body: "this is a test!!!",
  ContentType: "text/plain",
  ChecksumAlgorithm: "SHA256",
});

var get;
const getResult = await s3Client.send(
  (get = new GetObjectCommand({
    Bucket: bucket,
    Key: key,
    ChecksumMode: "ENABLED",
  }))
);

console.log("body ctor", getResult.Body.constructor.name);
console.log("body contents:", await getResult.Body?.transformToString("utf-8"));

console.log(get.middlewareStack.identify());
```

### DIFF
--- a/packages/middleware-flexible-checksums/src/flexibleChecksumsMiddleware.ts
+++ b/packages/middleware-flexible-checksums/src/flexibleChecksumsMiddleware.ts
@@ -1,4 +1,4 @@
-import { HttpRequest, HttpResponse } from "@smithy/protocol-http";
+import { HttpRequest } from "@smithy/protocol-http";
 import {
   BuildHandler,
   BuildHandlerArguments,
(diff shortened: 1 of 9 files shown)


packages/middleware-flexible-checksums/src/flexibleChecksumsMiddleware.ts
```diff
@@ -78,14 +80,5 @@ export const flexibleChecksumsMiddleware =
...
 
-    const { requestValidationModeMember, responseAlgorithms } = middlewareConfig;
-    // @ts-ignore Element implicitly has an 'any' type for input[requestValidationModeMember]
-    if (requestValidationModeMember && input[requestValidationModeMember] === "ENABLED") {
-      await validateChecksumFromResponse(result.response as HttpResponse, {
-        config,
-        responseAlgorithms,
-      });
-    }
-
```
kuhe, line 89: the response portion of this middleware is moved to a separate middleware that goes after the deser stage. 

the plugin installs both of these new middleware to preserve functionality

trivikr (approved):

PR #5157, opened 2023-08-30
feat(region-config): add aws region runtime extension codegen

AndrewFossAWS:
### Description
Add aws region runtime extension client changes

### Testing
yarn test:all

---


packages/token-providers/package.json
```diff
@@ -61,6 +61,7 @@
     "@smithy/util-defaults-mode-node": "^2.0.10",
     "@smithy/util-retry": "^2.0.0",
     "@smithy/util-utf8": "^2.0.0",
+    "@smithy/querystring-builder": "^2.0.6",
```
syall: Why is this dependency needed?

syall (approved):
Make sure the codegen are up-to-date with the latest commits.

packages/token-providers/package.json
AndrewFossAWS: It's not needed. I removed in the second commit.

kuhe (approved):

Issue #4464, opened 2023-02-23
When using vite to build the browser environment package, an unrelated package（@aws-sdk/signature-v4-crt） is added


ZengTianShengZ:
### Checkboxes for prior research

- [X] I've gone through [Developer Guide](https://docs.aws.amazon.com/sdk-for-javascript/v3/developer-guide) and [API reference](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest)
- [X] I've checked [AWS Forums](https://forums.aws.amazon.com) and [StackOverflow](https://stackoverflow.com/questions/tagged/aws-sdk-js).
- [X] I've searched for [previous similar issues](https://github.com/aws/aws-sdk-js-v3/issues) and didn't find any solution.

### Describe the bug

I use aws-sdk-js-v3 sdk as a file upload tool. When I use vite to build this tool, the build product adds an unrelated package (@aws-sdk/signature-v4-crt). I see that this package is used in the nodejs environment. But I use the browser environment. Why is this package used, and how can I troubleshoot this package. And this package（@aws-sdk/signature-v4-crt） is built in commonjs module mode, what I need is es module mode

<img width="1101" alt="image" src="https://user-images.githubusercontent.com/15622519/220856975-da0b8f12-8250-4178-a5ab-8631b1cec842.png">
![image](https://user-images.githubusercontent.com/15622519/220857069-7237f6f3-86c2-483e-89a8-e9f11479e919.png)

### SDK version number

 "@aws-sdk/client-s3": "^3.137.0", "vite": "^2.9.13",

### Which JavaScript Runtime is this issue in?

Browser

### Details of the browser/Node.js/ReactNative version

node v16.15.1

### Reproduction Steps

my vite build config

```js
import { defineConfig } from 'vite';
import lerna from '../../lerna.json';
import dts from 'vite-plugin-dts';

export default defineConfig({
    plugins: [
        dts(),
    ],
    build: {
        rollupOptions: {
            external: ['axios'],
        },
        lib: {
            entry: './src/index.ts',
            formats: ['es'],
            fileName: () => {
                return 'index.es.js';
            },
        },
    },
});
```

### Observed Behavior

```js
Please check if you have installed "@aws-sdk/signature-v4-crt" package explicitly
```

### Expected Behavior

Do not import the `@aws-sdk/signature-v4-crt` package, or the imported `@aws-sdk/signature-v4-crt` package is in es module mode

RanVaknin:
[...]

Issue #4373, opened 2023-01-26
util-user-agent-node and signature-v4-multi-region fails with ESM


ciscoheat:
### Checkboxes for prior research

- [X] I've gone through [Developer Guide](https://docs.aws.amazon.com/sdk-for-javascript/v3/developer-guide) and [API reference](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest)
- [X] I've checked [AWS Forums](https://forums.aws.amazon.com) and [StackOverflow](https://stackoverflow.com/questions/tagged/aws-sdk-js).
- [X] I've searched for [previous similar issues](https://github.com/aws/aws-sdk-js-v3/issues) and didn't find any solution.

### Describe the bug

In version 3.257.0 I suddenly get an ESM-related error when using vitest with a SvelteKit site. By configuring vite to inline the packages, it works again.

### SDK version number

@aws-sdk/util-user-agent-node@3.257.0

### Which JavaScript Runtime is this issue in?

Node.js

### Details of the browser/Node.js/ReactNative version

v18.10.0

### Reproduction Steps

Sorry but I cannot reproduce this in an easy way. I'm hoping that the error is well-known and has an easy fix.

### Observed Behavior

```
SyntaxError: Cannot use import statement outside a module
 ❯ Object.compileFunction node:vm:360:18

Module .../node_modules/.pnpm/@aws-sdk+util-user-agent-node@3.257.0/node_modules/@aws-sdk/util-user-agent-node/dist-es/is-crt-available.js:1 seems to be an ES Module but shipped in a CommonJS package. You might want to create an issue to the package "@aws-sdk/util-user-agent-node" asking them to ship the file in .mjs extension or add "type": "module" in their package.json.

As a temporary workaround you can try to inline the package by updating your config:

// vitest.config.js
export default {
  test: {
    deps: {
      inline: [
        "@aws-sdk/signature-v4-multi-region"
      ]
    }
  }
}
```

### Expected Behavior

No error message, it used to work fine.

### Additional Information/Context

The suggested workaround works, when adding this to `vite.config.js`

```ts
import { defineConfig } from 'vite';

export default defineConfig(({ command }) => {
  return {
    test: {
      deps: {
        inline: ['@aws-sdk/util-user-agent-node', '@aws-sdk/signature-v4-multi-region']
      }
    }
  }
}
```

RanVaknin:
Hi @ciscoheat ,

Thanks for opening this issue.

[...]

Issue #5135, opened 2023-08-19
Module not found


DrillableBit:
### Describe the bug

Start a fresh next.js application ( create-next-app myApp) - use app router. 

 npm install @aws-sdk/client-s3 @aws-sdk/s3-request-presigner
 
 
 import into a file (don't use): 
 import { getSignedUrl } from "@aws-sdk/s3-request-presigner";
import { S3Client, PutObjectCommand } from "@aws-sdk/client-s3";

### Expected Behavior

- warn ./node_modules/@aws-sdk/signature-v4-multi-region/dist-cjs/SignatureV4MultiRegion.js
Module not found: Can't resolve '@aws-sdk/signature-v4-crt' in 'C:\Users\aapli\Desktop\Programming\help-s3\node_modules\@aws-sdk\signature-v4-multi-region\dist-cjs'

Import trace for requested module:
./node_modules/@aws-sdk/signature-v4-multi-region/dist-cjs/SignatureV4MultiRegion.js
./node_modules/@aws-sdk/signature-v4-multi-region/dist-cjs/index.js
./node_modules/@aws-sdk/s3-request-presigner/dist-cjs/presigner.js
./node_modules/@aws-sdk/s3-request-presigner/dist-cjs/index.js
./src/_lib/afile.jsx
./src/app/page.tsx

in the console

### Current Behavior

- warn ./node_modules/@aws-sdk/signature-v4-multi-region/dist-cjs/SignatureV4MultiRegion.js
Module not found: Can't resolve '@aws-sdk/signature-v4-crt' in 'C:\Users\aapli\Desktop\Programming\help-s3\node_modules\@aws-sdk\signature-v4-multi-region\dist-cjs'

Import trace for requested module:
./node_modules/@aws-sdk/signature-v4-multi-region/dist-cjs/SignatureV4MultiRegion.js
./node_modules/@aws-sdk/signature-v4-multi-region/dist-cjs/index.js
./node_modules/@aws-sdk/s3-request-presigner/dist-cjs/presigner.js
./node_modules/@aws-sdk/s3-request-presigner/dist-cjs/index.js
./src/_lib/afile.jsx
./src/app/page.tsx

### Reproduction Steps

npx create-next-app 
typescript, tailwind, approuter

 npm install @aws-sdk/client-s3 @aws-sdk/s3-request-presigner

 import into a file (don't use): 
 import { getSignedUrl } from "@aws-sdk/s3-request-presigner";
[...]

PR #5203, opened 2023-09-12
fix(signature-v4-multi-region): make conditional require call in node.js only

kuhe:
related to https://github.com/aws/aws-sdk-js-v3/issues/5135
mentioned in https://github.com/aws/aws-sdk-js-v3/issues/4797
fixes https://github.com/aws/aws-sdk-js-v3/issues/4464
related to https://github.com/aws/aws-sdk-js-v3/issues/2822
partially addresses https://github.com/aws/aws-sdk-js-v3/issues/4126
fixes https://github.com/aws/aws-sdk-js-v3/issues/4373

- inverts the dependency between `signature-v4-crt` and `signature-v4-multi-region`. `crt` now depends on `multi-region`, and `multi-region` does not depend on `crt`, even optionally. 
- why? To avoid the conditional runtime require being analyzed as a static import by bundlers. Now, if you install `signature-v4-crt`, it will inject itself into a container that is used by `signature-v4-multi-region`.

revision 1:
- created node and browser files to load the CRT package as suggested in https://github.com/aws/aws-sdk-js-v3/issues/4126

### DIFF
--- a/packages/signature-v4-crt/package.json
+++ b/packages/signature-v4-crt/package.json
@@ -22,6 +22,7 @@
   },
   "license": "Apache-2.0",
   "dependencies": {
+    "@aws-sdk/signature-v4-multi-region": "*",
     "@smithy/querystring-parser": "^2.0.0",
     "@smithy/signature-v4": "^2.0.0",
     "@smithy/util-middleware": "^2.0.0",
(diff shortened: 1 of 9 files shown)


siddsriv (approved):

AllanZhengYP (approved):
Very nice to have this fixed! Thank you @kuhe

packages/signature-v4-multi-region/src/load-crt.node.ts
```diff
@@ -0,0 +1,20 @@
+import { signatureV4CrtContainer } from "./signature-v4-crt-container";
```
trivikr, line 1: We don't explicitly add `.node` suffix.

Rename this to `load-crt.ts`?
The change needs to be done in package.json overrides too.

trivikr (approved):

PR #5242, opened 2023-09-21
feat(credential-providers): add fromHttp credential provider

kuhe:
### Description
Adds an HTTP credential provider. 

### Testing
- [x] manual testing
- [x] implement retry options and retry wrapper
- [x] write unit tests
- [ ] write integration test?

### DIFF
--- a/packages/credential-provider-http/.gitignore
+++ b/packages/credential-provider-http/.gitignore
@@ -1,0 +1,8 @@
+/node_modules/
+/build/
+/coverage/
+/docs/
+*.tsbuildinfo
+*.tgz
+*.log
+package-lock.json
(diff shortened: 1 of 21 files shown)


packages/credential-providers/src/fromHttp/fromHttp.ts
```diff
@@ -0,0 +1,77 @@
...
+  return retryWrapper(
+    async (): Promise<AwsCredentialIdentity> => {
+      const request = createGetRequest(url);
+
+      if (token) {
+        request.headers.Authorization = token;
+      } else if (tokenFile) {
+        // Note: specification requires a file read on each request
+        // to allow for updates to the file contents.
+        request.headers.Authorization = fs.readFileSync(tokenFile).toString();
```
trivikr, line 69: Use async readFile from "fs/promises"?

```suggestion
        request.headers.Authorization = await fs.readFile(tokenFile).toString();
```

kuhe: done, changed import and await in parens () before toString

PR #5318, opened 2023-10-04
feat(middleware-endpoint): read service specific endpoints

trivikr:
### Issue
Internal JS-4281

### Description
Adds a utility to read service specific endpoints from env and ini files

### Testing
CI

---


trivikr:
Decision: we'll move getEndpointUrlConfig to `@aws-sdk/util-endpoints` for the following reasons:
* getEndpointUrlConfig is a utility, and not a middleware
* `@aws-sdk/middleware-endpoint` was deprecated in favor of `@smithy/middleware-endpoint`.  We don't want to revive, as it can create confusion.

PR #5343, opened 2023-10-10
test(client-s3): update S3.ispec.ts

kuhe:
restore s3 ispec to 1mb payload after fixes to fetch keepalive

- [x] this should wait until a smithy sync PR, which will likely happen during the identityAndAuth releases


kuhe:
- [x] yarn test:e2e passes

siddsriv (approved):

PR #5347, opened 2023-10-11
test(lib-dynamodb): add timestamp to table name in e2e tests

kuhe:
### Issue
internal JS-4767

Adds a timestamp component to the table name. While this unbounds the limit on the number of possible tables, the test deletes its table in `afterAll`, so there should only be leftover tables in unusual cases like process termination, which can be cleaned up manually.

### DIFF
--- a/lib/lib-dynamodb/src/test/lib-dynamodb.e2e.spec.ts
+++ b/lib/lib-dynamodb/src/test/lib-dynamodb.e2e.spec.ts
@@ -46,8 +46,9 @@ describe(DynamoDBDocument.name, () => {
   // don't delete the table in afterAll().
   // The table will in that case be re-used.
   const randId = String((Math.random() * 99) | 0);
+  const timestamp = (Date.now() / 1000) | 0;
 
-  const TableName = `js-sdk-dynamodb-test-${randId}`;
+  const TableName = `js-sdk-dynamodb-test-${timestamp}-${randId}`;
 
   const log = {
     describe: null as null | DescribeTableCommandOutput,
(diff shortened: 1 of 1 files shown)


lib/lib-dynamodb/src/test/lib-dynamodb.e2e.spec.ts
```diff
@@ -46,8 +46,9 @@ describe(DynamoDBDocument.name, () => {
   // don't delete the table in afterAll().
   // The table will in that case be re-used.
   const randId = String((Math.random() * 99) | 0);
+  const timestamp = (Date.now() / 1000) | 0;
 
-  const TableName = `js-sdk-dynamodb-test-${randId}`;
+  const TableName = `js-sdk-dynamodb-test-${timestamp}-${randId}`;
```
kuhe, line 51: this gets deleted at the end anyway

trivikr: Instead of using timestamp, create a random string of length 10 instead?

Current setup:
```console
> String((Math.random() * 99) | 0)
'32'
> String((Math.random() * 99) | 0)
'39'
> String((Math.random() * 99) | 0)
'36'
```

Suggestion:
```console
> (Math.random() + 1).toString(36).substring(2)
'kxp7dywlmb'
> (Math.random() + 1).toString(36).substring(1)
'.hu82ey92t7f'
> (Math.random() + 1).toString(36).substring(2)
'g0d90jwi1z'
```

kuhe: I prefer including both a timestamp and a random id so that the age of the table can be checked without additional effort beyond having its name.

kuhe:
```
 PASS  src/test/lib-dynamodb.e2e.spec.ts (30.32 s)
  DynamoDBDocument
    ✓ initializes using the static constructor
    ✓ is using a random TableName=js-sdk-dynamodb-test-1697046751-b6yu
```

trivikr (approved):

Issue #4278, opened 2022-12-13
Remove ` | string` from enums


everett1992:
### Describe the feature

smithy-typescript and aws-sdk-js-v3 generate clients with `| string` for any enum type in the model. It breaks auto complete, which is annoying, and compile time type checks, which is unsafe.

### Use Case

TypeScript simplifies the documented [^1] type from `ExecutionStatus | string` to `string` in error messages, auto complete, and type hints so it's hard to discover the generated enum values from an editor or commandline.

The inclusion of `string` in output types prevents me from writing exhaustiveness checks that would fail at compile time if my code did not handle all cases. For example handling step functions execution status[^1].

```typescript
// According to aws-sdk docs[^1] the type is ExecutionStatus | string | undefined
// but according to typescript the type is `string | undefined`.
//  ExecutionStatus | string is simplified to `string`.
const {status} = await sfn.describeExecution({executionArn})

switch (status) {
  case ExecutionStatus.ABORTED:
  case ExecutionStatus.FAILED:
  case ExecutionStatus.TIMED_OUT:
    return false;
  case ExecutionStatus.RUNNING:
    return undefined;
  case ExecutionStatus.SUCCEEDED:
    return true;
    console.log(`some known status ${status}`)
  case undefined:
    // https://github.com/aws/aws-sdk-js-v3/issues/1613
    throw Error('undefined')
  default:
    // This is a type error because status could be any string other than the enumerated values we checked above.
    assertNever(status)       ● Argument of type 'string' is not assignable to parameter of type 'never'.
}
```

The inclusion of `string` in input types allows me to pass invalid values[^2]
```typescript
await sfn.createStateMachine({
  // no type error here :(
  type: 'nope',
  ...args
})
```

I understand that smithy requires[^3] clients to allow unmodeled enum values but it does not require the current implementation.

[^1]: https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/clients/client-sfn/interfaces/describeexecutioncommandoutput.html#status
https://github.com/aws/aws-sdk-js-v3/blob/main/codegen/sdk-codegen/aws-models/sfn.json#L3479

[^2]: https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/clients/client-sfn/interfaces/createstatemachinecommandinput.html#type
https://github.com/aws/aws-sdk-js-v3/blob/main/codegen/sdk-codegen/aws-models/sfn.json#L5890

[...]

PR #5356, opened 2023-10-13
chore(codegen): update smithy commit to closed enums

kuhe:
### Issue
closes https://github.com/aws/aws-sdk-js-v3/issues/4278

### Description
Shapes with members targeting string and number enums (TypeScript unions) have been closed, no longer having `| string` or `| number` appended.

### Testing
This change has no runtime effect.

For user guidance, see https://github.com/awslabs/smithy-typescript/pull/1028


trivikr (approved):

trivikr (approved):

danielrbradley:
Just FYI @trivikr @kuhe this just caused us to have compilation failures as we were passing in variables as strings to be used in these properties typed as enums. 

IMO this should have been flagged as a breaking change as it could require users to rewrite their code to be compatible.

Issue #5360, opened 2023-10-15
client-dynamodb/lib-dynamodb v3.428.0: Sending an UpdateCommand fails with TypeError


maplesteve:
### Checkboxes for prior research

- [X] I've gone through [Developer Guide](https://docs.aws.amazon.com/sdk-for-javascript/v3/developer-guide) and [API reference](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest)
- [X] I've checked [AWS Forums](https://forums.aws.amazon.com) and [StackOverflow](https://stackoverflow.com/questions/tagged/aws-sdk-js).
- [X] I've searched for [previous similar issues](https://github.com/aws/aws-sdk-js-v3/issues) and didn't find any solution.

### Describe the bug

Since v3.428.0 updates on a DynamoDB table fail with a TypeError. It worked up to v3.427.0. The call of `new UpdateCommand(...)` is ok, but sending the call to DynamoDB fails. The properties of the `UpdateCommand` result are different between v3.428.0 and v3.427.0, which may lead to this error.

### SDK version number

@aws-sdk/lib-dynamodb@3.428.0

### Which JavaScript Runtime is this issue in?

Node.js

### Details of the browser/Node.js/ReactNative version

Node.js 18.x

### Reproduction Steps

```ts
let UpdateExpression = "SET autoDeployEnabled = :autoDeployEnabled";
if (args.config.manualDeploymentInfo !== undefined) {
  UpdateExpression += ", manualDeploymentInfo = :manualDeploymentInfo";
}
const cmd = new UpdateCommand({
  ExpressionAttributeValues: {
    ":autoDeployEnabled": args.config.autoDeployEnabled,
    ":manualDeploymentInfo": args.config.manualDeploymentInfo,
  },
  Key: { dataItemType: TestDeploymentDataItemType, repoName: args.repoName },
  ReturnValues: "ALL_NEW",
  TableName: GlobalConstants.DataItemsTableName,
  UpdateExpression,
});
const result = await DynamoDBDocumentClient.send(cmd);
```

### Observed Behavior

The call `await DynamoDBDocumentClient.send(cmd)` fails with a TypeError:
```
TypeError: Cannot read properties of undefined (reading 'S')
    at AttributeValue.visit (node_modules/@aws-sdk/client-dynamodb/dist-cjs/models/models_0.js:620:19)
[...]

PR #5306, opened 2023-10-02
fix(lib-dynamodb): add e2e suite and bug fixes for lib-dynamodb

kuhe:
### Issue
fixes https://github.com/aws/aws-sdk-js-v3/issues/3849
fixes https://github.com/aws/aws-sdk-js-v3/issues/4703
fixes https://github.com/aws/aws-sdk-js-v3/issues/4155

### Description
- [x] adding a new e2e suite
- [x] bug fixes

### Testing
- the new lib-dynamodb e2e suite in the PR

PR #5365, opened 2023-10-16
fix(lib-dynamodb): backwards compat for undefined columns

kuhe:
related to https://github.com/aws/aws-sdk-js-v3/issues/5360

A previous PR #5306 corrected behavior around what was considered an AttributeValue.

This excluded undefined row columns from processing because an undefined column is not an undefined map or set member and cannot be resolved to any particular AttributeValue type. 

For backwards compatibility, this PR makes the marshaller ignore undefined columns.


siddsriv (approved):

mikebrantbison:
@kuhe Not sure this PR fully fixes the issue.  See related ticket - https://github.com/aws/aws-sdk-js-v3/issues/5368

kuhe:
P.S. "column" simply refers to top level data fields, not to imply DynamoDB is columnar.

PR #5424, opened 2023-10-27
test(karma-credential-loader): use workspace version of STS in karma credential loader

kuhe:
experiencing an unclear error with karma in E2E tests, trying this as a fix


siddsriv (approved):

packages/karma-credential-loader/package.json
```diff
@@ -21,7 +21,7 @@
   },
   "license": "Apache-2.0",
   "dependencies": {
-    "@aws-sdk/client-sts": "file:../../clients/client-sts",
+    "@aws-sdk/client-sts": "*",
```
kuhe, line 24: this was originally to break a circular dependency, but there is, at least currently, no cycle related to this package

Issue #5363, opened 2023-10-16
(lib-dynamodb): DocumentClient should handle i/o of large numbers e.g. BigInt


alanb-sony:
### Checkboxes for prior research

- [X] I've gone through [Developer Guide](https://docs.aws.amazon.com/sdk-for-javascript/v3/developer-guide) and [API reference](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest)
- [X] I've checked [AWS Forums](https://forums.aws.amazon.com) and [StackOverflow](https://stackoverflow.com/questions/tagged/aws-sdk-js).
- [X] I've searched for [previous similar issues](https://github.com/aws/aws-sdk-js-v3/issues) and didn't find any solution.

### Describe the bug

The following code:
```
const doc = DynamoDBDocument.from(new DynamoDB());
await doc.put({
  TableName: 'test',
  Item: { num: 1.23e100 },
});
```
Raises the error:
```
Error: Number 1.23e+100 is greater than Number.MAX_SAFE_INTEGER. Use BigInt.
```
We could change `num` to `BigInt` but it is actually a floating point value so we'd have to only convert large numbers, as our numbers are already stored as `Number`, changing them to `BigInt` isn't going to increase their precision

### SDK version number

@aws-sdk/lib-dynamodb@3.428.0

### Which JavaScript Runtime is this issue in?

Node.js

### Details of the browser/Node.js/ReactNative version

v18.18.0

### Reproduction Steps

```
const doc = DynamoDBDocument.from(new DynamoDB());
await doc.put({
  TableName: 'test',
  Item: { num: 1.23e100 },
});
```

### Observed Behavior

An error is thrown:
```
Error: Number 1.23e+100 is greater than Number.MAX_SAFE_INTEGER. Use BigInt.
```

### Expected Behavior

Large numbers should be stored in the database

alanb-sony:
Even converting a number to a BigInt isn't a trivial task as dynamo wont accept numbers with more than 38 digits of precision

kuhe:
For now, use the base DynamoDB client to `put` with large numbers.

kuhe:
A way to mark large or precise numbers for input and output will be included in the next release.

Instructions are at https://github.com/aws/aws-sdk-js-v3/pull/5427 or https://github.com/aws/aws-sdk-js-v3/blob/main/lib/lib-dynamodb/README.md#large-numbers-and-numbervalue

alanb-sony:
[...]

PR #5427, opened 2023-10-27
feat(lib-dynamodb): large number handling

kuhe:
### Issue
fixes https://github.com/aws/aws-sdk-js-v3/issues/5363

### Description
This PR introduces an input side container for large/precise numbers not representable by the Native JS number type.

```ts
  const dynamodb = new DynamoDB({ region: "us-west-2", maxAttempts: 10 });
  const doc = DynamoDBDocument.from(dynamodb, {
    unmarshallOptions: {
      wrapNumbers: true,
    },
  });
```

To write a large or precise number: 

```ts
import { DynamoDB } from "@aws-sdk/client-dynamodb";
import { NumberValue, DynamoDBDocument } from "@aws-sdk/lib-dynamodb";

// note, the client will not validate the acceptability of the number.
// it is only to preserve its precise representation.
DynamoDBDocument.from(new DynamoDB({})).put({
  Item: {
    id: NumberValue.from("1000000000000000000000.000000000001")
  }
});
```
(You can also use this wrapper for small numbers, but that is optional).
e.g.
```js
new Set([
  1, "1234567890", 2, 3, 4
].map(NumberValue.from));
```

To read large or precise numbers, set 
```
    unmarshallOptions: {
      wrapNumbers: true,
    },
```
as shown above. All numbers will be returned in the `NumberValue` wrapper.

### Testing
new unit and e2e tests

### Checklist
- [x] update lib-dynamodb documentation


packages/util-dynamodb/src/convertToNative.ts
```diff
@@ -41,22 +42,40 @@ export const convertToNative = (data: AttributeValue, options?: unmarshallOption
...
 };
 
-const convertNumber = (numString: string, options?: unmarshallOptions): number | bigint | NumberValue => {
-  if (options?.wrapNumbers) {
+const convertNumber = (
+  numString: string,
+  options?: unmarshallOptions
+): number | bigint | NumberValue | DynamoDBNumber => {
+  if (options?.wrapNumbers && !options?.useDynamoDBNumberWrapper) {
     return { value: numString };
```
kuhe, line 50: use `wrapNumbers` directly instead of `useDynamoDBNumberWrapper` and wrap it with DynamoDBNumber renamed to `NumberValue` (see util-dynamodb/models).

trivikr (approved):

PR #5468, opened 2023-11-08
chore: add package entrypoints enforcement script

kuhe:
paired with https://github.com/awslabs/smithy-typescript/pull/1066 (see explanation there)

### DIFF
--- a/packages/credential-provider-web-identity/package.json
+++ b/packages/credential-provider-web-identity/package.json
@@ -15,8 +15,7 @@
     "test": "jest"
   },
   "browser": {
-    "./dist-es/fromTokenFile": false,
-    "./dist-cjs/fromTokenFile": false
+    "./dist-es/fromTokenFile": false
   },
   "react-native": {
     "./dist-es/fromTokenFile": false,
--- a/packages/credential-provider-web-identity/package.json
+++ b/packages/credential-provider-web-identity/package.json
@@ -15,8 +15,8 @@
     "test": "jest"
   },
   "browser": {
-    "./dist-cjs/fromTokenFile": false,
-    "./dist-es/fromTokenFile": false
+    "./dist-es/fromTokenFile": false,
+    "./dist-cjs/fromTokenFile": false
   },
   "react-native": {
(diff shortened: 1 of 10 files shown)


syall (approved):

packages/credential-provider-web-identity/package.json
```diff
@@ -15,7 +15,6 @@
     "test": "jest"
   },
   "browser": {
-    "./dist-cjs/fromTokenFile": false,
```
trivikr, line 18: Make script more restrictive to check if the value is a string, and ignore if it's boolean?

This is needed as for webpack support https://github.com/aws/aws-sdk-js-v3/pull/3005

packages/rds-signer/package.json
```diff
@@ -60,9 +60,7 @@
   "browser": {
     "./dist-es/runtimeConfig": "./dist-es/runtimeConfig.browser"
   },
-  "react-native": {
-    "./dist-es/runtimeConfig": "./dist-es/runtimeConfig.native"
```
trivikr, line 64: Is this removal on purpose? Does React Native has any other bundlers which might require this configuration?

If not, can we remove the `react-native` key itself?

packages/credential-provider-web-identity/package.json
kuhe: we don't provide dist-cjs corrections in other packages for the browser environment

packages/rds-signer/package.json
kuhe: this removal is on purpose because you asked that we do not explicitly add the replacements that would be automatic. 

react-native is left as an object because of the consistency enforcement that `browser` and `react-native` are of the same type: object or string or undefined.

this prevents errors like forgetting to add the react-native directive when creating a browser replacement

packages/credential-provider-web-identity/package.json
kuhe: added condition to skip dropping boolean directives in browser with CJS to avoid unforeseen consequences

trivikr (approved):

PR #5648, opened 2024-01-05
chore: add migration issue template for SDK v2 to v3 upgrade

RanVaknin:
Adding a new issue template targeting migration related issues.

### DIFF
--- a/.github/ISSUE_TEMPLATE/v2-to-v3-migration.yml
+++ b/.github/ISSUE_TEMPLATE/v2-to-v3-migration.yml
@@ -28,10 +28,20 @@ body:
     attributes:
       label: Which JavaScript Runtime is this issue in?
       options:
-        - Node.js
+        - Node.js (includes AWS Lambda)
         - Browser
-        - Lambda
         - React Native
+    validations:
+      required: true
+
+  - type: checkboxes
+    id: lambda-use
+    attributes:
+      label: AWS Lambda Usage
+      description: Are you running your application on AWS Lambda?
+      options:
+        - label: Yes, my application is running on AWS Lambda.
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/v2-to-v3-migration.yml
```diff
@@ -0,0 +1,77 @@
...
+          required: true
+
+  - type: dropdown
+    id: runtime
+    attributes:
+      label: Which JavaScript Runtime is this issue in?
+      options:
+        - Node.js
+        - Browser
+        - Lambda
```
trivikr, line 33: Users might get confused between Node.js and Lambda.
```suggestion
        - Node.js
        - Browser
```

kuhe: If you really want to include the word Lambda, it could be 

`Node.js (includes AWS Lambda)`

or add a yes or no question

> if you selected Node.js, is your application running on AWS Lambda?

trivikr:
```suggestion
        - Node.js (includes AWS Lambda)
        - Browser
```

RanVaknin: I often find myself 5-6 back and fourth messages deep into a thread conversation to only then realize customers are talking about Lambda specific issues. 

I think having Lambda as an option is super helpful for OSDS to form an opinion right from the get-go.

How about this?

```
        - Node.js
        - Node.js on Lambda
        - Browser
        - React Native
```

trivikr (approved):

Issue #5632, opened 2024-01-02
Transcribe Client has broken dependencies


qhenkart:
### Checkboxes for prior research

- [X] I've gone through [Developer Guide](https://docs.aws.amazon.com/sdk-for-javascript/v3/developer-guide) and [API reference](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest)
- [X] I've checked [AWS Forums](https://forums.aws.amazon.com) and [StackOverflow](https://stackoverflow.com/questions/tagged/aws-sdk-js).
- [X] I've searched for [previous similar issues](https://github.com/aws/aws-sdk-js-v3/issues) and didn't find any solution.

### Describe the bug

Node version 20.10.0

package.json dependency:
```
    "@aws-sdk/client-transcribe-streaming": "^3.484.0",
```

When running using yarn 4.0.2 on pnp mode

I resolved it by manually installing @smithy/eventstream-codec

```console
 ERROR in ./.yarn/cache/@aws-sdk-middleware-websocket-npm-3.468.0-1d6a9d1e6e-ab5c9b242e.zip/node_modules/@aws-sdk/middleware-websocket/dist-es/EventStreamPayloadHandler.js 1:0-61
 Module not found: Error: Can't resolve '@smithy/eventstream-codec' in '/usr/src/app/.yarn/cache/@aws-sdk-middleware-websocket-npm-3.468.0-1d6a9d1e6e-ab5c9b242e.zip/node_modules/@aws-sdk/middleware-websocket/dist-es'
 resolve '@smithy/eventstream-codec' in '/usr/src/app/.yarn/cache/@aws-sdk-middleware-websocket-npm-3.468.0-1d6a9d1e6e-ab5c9b242e.zip/node_modules/@aws-sdk/middleware-websocket/dist-es'
   Parsed request is a module
   using description file: /usr/src/app/.yarn/cache/@aws-sdk-middleware-websocket-npm-3.468.0-1d6a9d1e6e-ab5c9b242e.zip/node_modules/@aws-sdk/middleware-websocket/package.json (relative path: ./dist-es)
     Field 'browser' doesn't contain a valid alias configuration
     resolve as module
       looking for modules in /usr/src/app/src
         single file module
           using description file: /usr/src/app/package.json (relative path: ./src/@smithy/eventstream-codec)
             no extension
               Field 'browser' doesn't contain a valid alias configuration
[...]

PR #5657, opened 2024-01-09
fix(middleware-websocket): add missing codec dependency from smithy

RanVaknin:
Adresses #5632 , adding the `eventstream-codec` pacakge as a direct dependency since we rely on it in our code in https://github.com/aws/aws-sdk-js-v3/blob/main/packages/middleware-websocket/src/EventStreamPayloadHandler.ts

### DIFF
--- a/packages/middleware-websocket/package.json
+++ b/packages/middleware-websocket/package.json
@@ -30,6 +30,7 @@
     "@smithy/signature-v4": "^2.0.0",
     "@smithy/types": "^2.8.0",
     "@smithy/util-hex-encoding": "^2.0.0",
+    "@smithy/eventstream-codec": "^2.0.16", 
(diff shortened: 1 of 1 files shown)


packages/middleware-websocket/package.json
```diff
@@ -30,6 +30,7 @@
     "@smithy/signature-v4": "^2.0.0",
     "@smithy/types": "^2.8.0",
     "@smithy/util-hex-encoding": "^2.0.0",
+    "@smithy/eventstream-codec": "^2.0.16",
```
kuhe, line 33: I don't think this passes our pre-commit hooks, did you do this in github UI?

kuhe:
use this one: https://github.com/aws/aws-sdk-js-v3/pull/5658

PR #5658, opened 2024-01-09
fix: dependency check script and undeclared dependencies

kuhe:
- fix the dependency check script, a `.slice(1)` was omitting one of the matches that should have been checked.
- then add any detected missing dependencies.


kuhe:
fixes https://github.com/aws/aws-sdk-js-v3/issues/5632

siddsriv (approved):

PR #5664, opened 2024-01-10
chore: add smoke test traits dep

milesziemer:
### Issue

### Description

Adds a dependency on smithy-smoke-test-traits in the sdk code gen package so aws service models that use the smokeTests trait won't fail to build because of unknown traits.

### Testing
How was this change tested?

Generated a client with a smoke test and confirmed there was no diff.

### Additional context

Adding the dependency in the code gen package since it's the models that need the dependency, not the code generator. When smoke tests are implemented in the generator, we can add the dep there.

### Checklist

---

### DIFF
--- a/codegen/sdk-codegen/build.gradle.kts
+++ b/codegen/sdk-codegen/build.gradle.kts
@@ -43,7 +43,9 @@ plugins {
 }
 
 dependencies {
+    val smithyVersion: String by project
     implementation(project(":smithy-aws-typescript-codegen"))
+    implementation("software.amazon.smithy:smithy-smoke-test-traits:$smithyVersion")
 }
 
 // This project doesn't produce a JAR.


kuhe (approved):

syall (approved):

hpmellema (changes requested):
Should this be added to the same place as the other trait dependencies? 
https://github.com/aws/aws-sdk-js-v3/blob/main/codegen/smithy-aws-typescript-codegen/build.gradle.kts#L37

milesziemer:
> Should this be added to the same place as the other trait dependencies? https://github.com/aws/aws-sdk-js-v3/blob/main/codegen/smithy-aws-typescript-codegen/build.gradle.kts#L37

Since this is a model dependency right now, i.e. the code generator implementation isn't using it, we're just adding it as a dep where the models are loaded. If anything it is a little funky that the model dependencies would be available through the code generator implementation package.

hpmellema (approved):

Issue #5651, opened 2024-01-09
[S3] Impossible to set SSECustomerKey with binary AES256 key


tmccombs:
### Checkboxes for prior research

- [X] I've gone through [Developer Guide](https://docs.aws.amazon.com/sdk-for-javascript/v3/developer-guide) and [API reference](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest)
- [X] I've checked [AWS Forums](https://forums.aws.amazon.com) and [StackOverflow](https://stackoverflow.com/questions/tagged/aws-sdk-js).
- [X] I've searched for [previous similar issues](https://github.com/aws/aws-sdk-js-v3/issues) and didn't find any solution.

### Describe the bug

This is yet another reiteration of https://github.com/aws/aws-sdk-js-v3/issues/4736. Because I'm pretty sure something is not working as intend here, and cannot figure out how to convert a `UInt8Array` to a string that is a valid key for SSECustomerKey in the S3 APIs. 

Basically, if I have a `UInt8Array` of 32 random bytes for an AES256 key, there doesn't seem to be any way to pass that to `SSECustomerKey` that complies with the tyepscript types for `CopyObjectCommandInput`, `PutObjectCommandInput`, `GetObjectCommandInput`, etc. 

### SDK version number

@aws-sdk/package-name@version, ...

### Which JavaScript Runtime is this issue in?

Node.js

### Details of the browser/Node.js/ReactNative version

18.18.0

### Reproduction Steps

```typescript
import {S3Client, PutObjectCommand} from '@aws-sdk/client-s3';
import * as crypto from 'crypto';
import {promisify} from 'util';

const randomBytes = promisify(crypto.randomBytes);

const client = new S3Client();
key = await randomBytes(32);
const response = await client.send(new PutObjectCommnd({Bucket: 'mybucket', Body: 'This is a test', SSECustomerKey: key.toString(), SSECustomerAlgorithm: 'AES256});

```

### Observed Behavior

The request fails with a 400, with the error "InvalidArgument: The secret key was invalid for the specified algorithm."

This happens regardless of the encoding used. I have tried calling `toString` on the key with the following encodings:

- utf8
- binary
- base64
- hex

All with the same result.

### Expected Behavior

It should be possible to use an arbitrary, randomly generated key for the SSE customer key, even if it isn't a valid UTF-8 string.

### Possible Solution

[...]

Issue #4736, opened 2023-05-19
PutObjectCommandInput.SSECustomerKey type is incorrect


DavidZbarsky-at:
### Checkboxes for prior research

- [X] I've gone through [Developer Guide](https://docs.aws.amazon.com/sdk-for-javascript/v3/developer-guide) and [API reference](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest)
- [X] I've checked [AWS Forums](https://forums.aws.amazon.com) and [StackOverflow](https://stackoverflow.com/questions/tagged/aws-sdk-js).
- [X] I've searched for [previous similar issues](https://github.com/aws/aws-sdk-js-v3/issues) and didn't find any solution.

### Describe the bug

This is a dupe of https://github.com/aws/aws-sdk-js-v3/issues/2081 that I am reopening because the explanation didn't make sense to me. I am still seeing this issue, but I am using `PutObjectCommandInput` as advised, rather than `PutObjectRequest`. See also generated types at https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/clients/client-s3/interfaces/putobjectcommandinput.html#ssecustomerkey-7

Is there something in the codegen process that is supposed to expand the type to account for conversions?

### SDK version number

@aws-sdk/client-v3@latest

### Which JavaScript Runtime is this issue in?

Node.js

### Details of the browser/Node.js/ReactNative version

16

### Reproduction Steps

```
const data: PutObjectCommandInput = {
    SSECustomerKey: Buffer.from('foo')
}
```

### Observed Behavior

```(property) SSECustomerKey?: string | undefined
Type 'Buffer' is not assignable to type 'string'.ts(2322)
models_0.d.ts(8835, 5): The expected type comes from property 'SSECustomerKey' which is declared here on type 'PutObjectCommandInput'```

### Expected Behavior

no type error

### Possible Solution

_No response_

### Additional Information/Context

_No response_

yenfryherrerafeliz:
Hi @DavidZbarsky-at, sorry to hear about your issues. Right now, SSECustomerKey parameter is expected to be string, so that it needs to be provided as so. You can see it's definition [here](https://github.com/aws/aws-sdk-js-v3/blob/45df38afdaf09468648553c9ce36abaca54653d0/clients/client-s3/src/models/models_1.ts#L1057). However, I can mark this as a feature request, but, I would need you to provide more details about why this change is needed?, and what exactly is your use case?.

[...]

PR #5676, opened 2024-01-11
fix(middleware-ssec): add logic to handle string input as specified b…

RanVaknin:
### Background:
middleware-ssec was ported over from v2, however the implementation prevents from customer to supply `SSECustomerKey` in a string format which is problematic as `putObject#SSECustomerKey` as modeled as a string. Instead customer are required to provide `SSECustomerKey` as binary. The middleware logic does not cover cases in which customers want to supply `SSECustomerKey` as a base 64 string directly as the S3 docs suggest.

### Issue
#5651 #4736 

### Reproduction:
```js
import { S3Client, PutObjectCommand } from '@aws-sdk/client-s3';
import * as crypto from 'crypto';
import { promisify } from 'util';

const randomBytes = promisify(crypto.randomBytes);

const client = new S3Client({region: "us-east-1"});

async function putWithSSECInBinaryFormat(){
  try {
    const key = await randomBytes(32);
    const response = await client.send(new PutObjectCommand({
      Bucket: 'testbucket',
      Body: 'This is a test',
      Key: "foo",
      SSECustomerKey: key,
      SSECustomerAlgorithm: 'AES256'
    }));
    console.log('PutObject with SSEC in binary format successful:', response.$metadata.httpStatusCode);
  } catch (error) {
    console.error('PutObject with SSEC in binary format failed:', error);
  }

}

async function putWithSSECInBase64StringFormat(){
  try {
    const key = await randomBytes(32);
    const response = await client.send(new PutObjectCommand({
      Bucket: 'testbucket',
      Body: 'This is a test',
      Key: "foo",
      SSECustomerKey: key.toString('base64'),
      SSECustomerAlgorithm: 'AES256'
    }));
    console.log('PutObject with SSEC in base64 string format successful:', response.$metadata.httpStatusCode);
  } catch (error) {
    console.error('PutObject with SSEC in base64 string format failed:', error);
  }
}

putWithSSECInBinaryFormat();
putWithSSECInBase64StringFormat();
```

Will Result in:

```
PutObject with SSEC in binary format successful: 200
PutObject with SSEC in base64 string format failed: InvalidArgument: The secret key was invalid for the specified algorithm.
```

### Changes Made:
**Middleware Logic Update:**

- Regex Adjustment for Base64 Detection: now using a regular expression used to determine if a string is base64-encoded.
- Handling of Different Input Types: Modified the logic to distinguish between base64-encoded strings, non-base64 strings, and binary data. Each type is now appropriately handled:
  -  Base64 strings are directly converted to binary for MD5 hashing without re-encoding.
Non-base64 strings are decoded to binary, encoded to base64, and their MD5 hashes are calculated.
  - Binary data is directly used for MD5 hashing and then encoded to base64.

**Test Suite Enhancement:**
- Test for Base64-Encoded Key Processing: This test confirms that the middleware correctly handles base64-encoded keys by verifying that it properly maintains the key in base64 format and accurately calculates and sets the corresponding base64-encoded MD5 hash.

### Behavior after change 
```
PutObject with SSEC in binary format successful: 200
PutObject with SSEC in base64 string format successful: 200
```
---

### DIFF
--- a/packages/middleware-ssec/src/index.ts
+++ b/packages/middleware-ssec/src/index.ts
@@ -21,7 +21,7 @@ interface PreviouslyResolved {
 export function ssecMiddleware(options: PreviouslyResolved): InitializeMiddleware<any, any> {
   return <Output extends MetadataBearer>(next: InitializeHandler<any, Output>): InitializeHandler<any, Output> =>
     async (args: InitializeHandlerArguments<any>): Promise<InitializeHandlerOutput<Output>> => {
-      let input = { ...args.input };
+      const input = { ...args.input };
       const properties = [
         {
           target: "SSECustomerKey",
@@ -36,19 +36,25 @@ export function ssecMiddleware(options: PreviouslyResolved): InitializeMiddlewar
       for (const prop of properties) {
         const value: SourceData | undefined = (input as any)[prop.target];
         if (value) {
-          const valueView: Uint8Array = ArrayBuffer.isView(value)
-            ? new Uint8Array(value.buffer, value.byteOffset, value.byteLength)
-            : typeof value === "string"
-            ? options.utf8Decoder(value)
-            : new Uint8Array(value);
-          const encoded = options.base64Encoder(valueView);
+          let valueForHash: Uint8Array;
+          if (typeof value === "string") {
+            const isBase64Encoded = /^(?:[A-Za-z0-9+/]{4})*([A-Za-z0-9+/]{2}==|[A-Za-z0-9+/]{3}=)?$/.test(value);
+            if (isBase64Encoded) {
+              valueForHash = new Uint8Array(Buffer.from(value, "base64"));
+            } else {
+              valueForHash = options.utf8Decoder(value);
+              input[prop.target] = options.base64Encoder(valueForHash);
+            }
+          } else {
+            valueForHash = ArrayBuffer.isView(value)
+              ? new Uint8Array(value.buffer, value.byteOffset, value.byteLength)
+              : new Uint8Array(value);
+            input[prop.target] = options.base64Encoder(valueForHash);
+          }
+
           const hash = new options.md5();
-          hash.update(valueView);
-          input = {
-            ...(input as any),
-            [prop.target]: encoded,
-            [prop.hash]: options.base64Encoder(await hash.digest()),
-          };
+          hash.update(valueForHash);
+          input[prop.hash] = options.base64Encoder(await hash.digest());
         }
       }
 
--- a/packages/middleware-ssec/src/index.ts
+++ b/packages/middleware-ssec/src/index.ts
@@ -40,7 +40,7 @@ export function ssecMiddleware(options: PreviouslyResolved): InitializeMiddlewar
           if (typeof value === "string") {
             const isBase64Encoded = /^(?:[A-Za-z0-9+/]{4})*([A-Za-z0-9+/]{2}==|[A-Za-z0-9+/]{3}=)?$/.test(value);
             if (isBase64Encoded) {
-              valueForHash = new Uint8Array(Buffer.from(value, "base64"));
+              valueForHash = base64ToUint8Array(value);
             } else {
               valueForHash = options.utf8Decoder(value);
               input[prop.target] = options.base64Encoder(valueForHash);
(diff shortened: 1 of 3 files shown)


packages/middleware-ssec/src/index.ts
```diff
@@ -36,19 +36,25 @@ export function ssecMiddleware(options: PreviouslyResolved): InitializeMiddlewar
       for (const prop of properties) {
         const value: SourceData | undefined = (input as any)[prop.target];
         if (value) {
-          const valueView: Uint8Array = ArrayBuffer.isView(value)
-            ? new Uint8Array(value.buffer, value.byteOffset, value.byteLength)
-            : typeof value === "string"
-            ? options.utf8Decoder(value)
-            : new Uint8Array(value);
-          const encoded = options.base64Encoder(valueView);
+          let valueForHash;
```
kuhe, line 39: create a type annotation for this var init. It looks like `Uint8Array`.

kuhe (changes requested):
fix commit metadata

packages/middleware-ssec/src/index.ts
```diff
@@ -36,19 +36,25 @@ export function ssecMiddleware(options: PreviouslyResolved): InitializeMiddlewar
...
-            ? new Uint8Array(value.buffer, value.byteOffset, value.byteLength)
-            : typeof value === "string"
-            ? options.utf8Decoder(value)
-            : new Uint8Array(value);
-          const encoded = options.base64Encoder(valueView);
+          let valueForHash: Uint8Array;
+          if (typeof value === "string") {
+            const isBase64Encoded = /^(?:[A-Za-z0-9+/]{4})*([A-Za-z0-9+/]{2}==|[A-Za-z0-9+/]{3}=)?$/.test(value);
+            if (isBase64Encoded) {
+              valueForHash = new Uint8Array(Buffer.from(value, "base64"));
```
kuhe, line 43: Buffer.from isn't compatible with browsers

RanVaknin: Because that value is already base64 encoded, and does not need to be re-set on that target. 

Otherwise it would be 
```js
if (isBase64Encoded) {
              valueForHash = base64ToUint8Array(value);
              input[prop.target] = value; // and value is already set to be the target on line 37
}
```

packages/middleware-ssec/src/index.ts
```diff
@@ -59,6 +65,16 @@ export function ssecMiddleware(options: PreviouslyResolved): InitializeMiddlewar
     };
 }
 
+function base64ToUint8Array(base64String: string) {
+  const binaryString = atob(base64String);
```
kuhe, line 69: test failing with ReferenceError: atob is not defined

RanVaknin: Added logic to check if Buffer is available (Nodejs), and if not use atob which is supported on browsers.

packages/middleware-ssec/src/index.ts
```diff
@@ -59,6 +65,21 @@ export function ssecMiddleware(options: PreviouslyResolved): InitializeMiddlewar
     };
 }
 
+function base64ToUint8Array(base64String: string) {
```
kuhe, line 68: try options.base64Decoder

kuhe (approved):

kuhe (approved):

packages/middleware-ssec/src/middleware-ssec.integ.spec.ts
```diff
@@ -1,4 +1,6 @@
 import { S3 } from "@aws-sdk/client-s3";
+import * as crypto from "crypto";
+import { isMainThread } from "worker_threads";
```
kuhe, line 3: remove this

packages/middleware-ssec/src/index.ts
```diff
@@ -36,19 +37,25 @@ export function ssecMiddleware(options: PreviouslyResolved): InitializeMiddlewar
...
-          const valueView: Uint8Array = ArrayBuffer.isView(value)
-            ? new Uint8Array(value.buffer, value.byteOffset, value.byteLength)
-            : typeof value === "string"
-            ? options.utf8Decoder(value)
-            : new Uint8Array(value);
-          const encoded = options.base64Encoder(valueView);
+          let valueForHash: Uint8Array;
+          if (typeof value === "string") {
+            const isBase64Encoded = /^(?:[A-Za-z0-9+/]{4})*([A-Za-z0-9+/]{2}==|[A-Za-z0-9+/]{3}=)?$/.test(value);
+            if (isBase64Encoded) {
```
tmccombs, line 43: What if the value is a valid base64 string, but isn't actually a base64 encoded value. 

I think probably the best way to check this would be to look at the length of the string. If the length of the string is 44 characters, then it is probably a base64 encoded value (for 32 bytes of data). Assuming that AES256 is the encryption algorithm.

PR #5677, opened 2024-01-11
feat(credential-provider-node): use dynamic import for credential providers

kuhe:
### Description

- this PR uses dynamic imports in the node default credential provider chain
  - this reduces the number of files loaded during app initialization
  - Node.js has good support for dynamic imports whether via `await import()` or simple dynamic `require()` calls.

This avoids the issue of browser/bundler and react-native compatibility. 

### Testing
- [x] cucumber-js e2e
- [x] `yarn test:e2e`
- [x] `yarn test:integration`
- [x] check Node.js 14.x support (yes, supported, but also TS compiles `import()` to a `require`) so there is no difference. ESBuild if used in the future does not convert `import()` to `require()`, but it does not need to.

effects on SDK initialization times:
- dynamic imports that are avoided save time for SDK init
- dynamic imports are not slower than regular imports (in dist-cjs they are all require at this point anyway)


kuhe:
Example in conjunction with "package-bundling":
Loading S3 and making 1 request. Initially only 135 files are loaded. After 3 providers are invoked, the loaded number of files reaches 151.

```
files 135
AWS SDK loading env provider
AWS SDK loading sso provider
AWS SDK loading ini provider
{
  httpStatusCode: 200,
  requestId: 'FJ4845PS5MW1VSB4',
  extendedRequestId: '...+...+...+...=',
  cfId: undefined,
  attempts: 1,
  totalRetryDelay: 0
}
files after init 151
```

trivikr (approved):

Issue #5686, opened 2024-01-16
DynamoDBDocumentClient throws on marshalling of class instances where class includes bound functions or arrow functions, despite convertClassInstanceToMap = true


firekart:
### Checkboxes for prior research

- [X] I've gone through [Developer Guide](https://docs.aws.amazon.com/sdk-for-javascript/v3/developer-guide) and [API reference](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest)
- [X] I've checked [AWS Forums](https://forums.aws.amazon.com) and [StackOverflow](https://stackoverflow.com/questions/tagged/aws-sdk-js).
- [X] I've searched for [previous similar issues](https://github.com/aws/aws-sdk-js-v3/issues) and didn't find any solution.

### Describe the bug
[ Edited to include bound functions ]
An error is now thrown If  bound or arrow functions are included as part of a class, when an instance of that class is used in PutCommand, BatchWriteCommand, DeleteCommand, etc. Example of error 'Unsupported type passed: () => this.props. Pass options.convertClassInstanceToMap=true to marshall typeof object as map attribute.'

### SDK version number

@aws-sdk/lib-dynamodb@3.490.0

### Which JavaScript Runtime is this issue in?

Node.js

### Details of the browser/Node.js/ReactNative version

v18.17.1

### Reproduction Steps

```
import { DynamoDBClient, DynamoDBClientConfig, } from "@aws-sdk/client-dynamodb";
import { DynamoDBDocumentClient, PutCommand, TranslateConfig } from "@aws-sdk/lib-dynamodb";

const datastoreParams: DynamoDBClientConfig = {
    region: 'ap-southeast-2'
};

const marshallOptions: TranslateConfig[ 'marshallOptions' ] = {
    convertEmptyValues: false,
    removeUndefinedValues: true,
    convertClassInstanceToMap: true,
};

const unmarshallOptions: TranslateConfig[ 'unmarshallOptions' ] = {
    wrapNumbers: false,
};

const translateConfig: TranslateConfig = { marshallOptions, unmarshallOptions };
const dynamoDb = new DynamoDBClient(datastoreParams);
const dynamoDbConfig = { client: dynamoDb };

const docClient = DynamoDBDocumentClient.from(dynamoDbConfig.client, translateConfig);

/** a function to be bound to X */
function boundGetProps( this: X ) {
    return this.props
}

/** a dummy class that conforms to our schema */
class X {
[...]

PR #5697, opened 2024-01-19
fix(lib-dynamodb): skip function properties in data objects for DynamoDB

siddsriv:
### Issue
#5686 

### Description
skipping function properties when processing data objects for DynamoDB

### Testing
- [x] Unit tests
- [x] E2E test

---

### DIFF
--- a/lib/lib-dynamodb/src/commands/utils.ts
+++ b/lib/lib-dynamodb/src/commands/utils.ts
@@ -77,7 +77,7 @@ const processKeysInObj = (obj: any, processFunc: Function, keyNodes: KeyNodeChil
       continue;
     }
     const processedValue = processObj(obj[nodeKey], processFunc, nodes);
-    if (processedValue !== undefined) {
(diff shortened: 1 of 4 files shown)


kuhe (approved):

lib/lib-dynamodb/src/commands/utils.ts
```diff
@@ -90,12 +90,41 @@ const processAllKeysInObj = (obj: any, processFunc: Function, keyNodes: KeyNodes
   }, {} as any);
 };
 
+function copyWithoutFunctions(o: any, depth = 0): any {
```
kuhe, line 93: we can actually remove all the other places that handle `convertClassInstanceToMap` since this is exhaustive

siddsriv: not many other places where we handle that option, can open a new PR for it if we decide to remove the checks for that option

Issue #5516, opened 2023-11-22
Bundling for Lambda requires unnecessary dependencies


perpil:
### Checkboxes for prior research

- [X] I've gone through [Developer Guide](https://docs.aws.amazon.com/sdk-for-javascript/v3/developer-guide) and [API reference](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest)
- [X] I've checked [AWS Forums](https://forums.aws.amazon.com) and [StackOverflow](https://stackoverflow.com/questions/tagged/aws-sdk-js).
- [X] I've searched for [previous similar issues](https://github.com/aws/aws-sdk-js-v3/issues) and didn't find any solution.

### Describe the bug

For Lambda coldstarts, it is well known creating a minified, tree-shaken bundle measurably improves performance vs. using the V3 JavaScript SDK version bundled with Lambda.  

| coldstart time (ms) | bundled/minified |
| ------------ | ------ |
| 675        | no|
| 241       | yes|

In a minified bundle, if you use any AWS client, `credential-provider-node` starts pulling in `client-sso` and other packages. These packages are unused but add > 32KB to the minified bundle.  If you set the `credentials` parameter for the AWS client, it still pulls in `client-sso` and can't be tree-shaken out. Since loading credentials from environment variables is what most developers need for Lambda, there should be a way to only bundle the `credential-provider-env` credentials provider.  Arguably this should be the default in the Lambda environment.

### SDK version number

@aws-sdk/client-sts@3.454.0

### Which JavaScript Runtime is this issue in?

Node.js

### Details of the browser/Node.js/ReactNative version

v20.9.0

### Reproduction Steps

1. Create a lambda with any AWS client, i.e. STS:
  ```javascript
  import { GetCallerIdentityCommand, STSClient } from '@aws-sdk/client-sts';

  const stsClient = new STSClient({region: process.env.AWS_REGION});

  export async function handler(event, context) {
    const result = await stsClient.send(new GetCallerIdentityCommand({}));
    return {
      statusCode: 200,
      body: JSON.stringify({
        result
      }),
      headers: {
        'Content-Type': 'application/json',
      },
    };
  }
  ```
2. Minify, treeshake and create metafile with esbuild. If you use [NodeJsFunction](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_lambda_nodejs.NodejsFunction.html) in the CDK your bundling looks like: 
```javascript
  bundling: {
[...]

PR #5698, opened 2024-01-19
feat(credential-provider-node): use dynamic import for credential providers

kuhe:
This re-stages https://github.com/aws/aws-sdk-js-v3/pull/5677 which was merged too early.

- [x] integ
- [x] e2e
- [x] e2e legacy

- [x] test Node.js platform target bundlers (esbuild, webpack)


siddsriv (approved):

mskonf:
@kuhe  I am not sure if this is the right place to mention this:
But we are using AWS SDK and after the latest release we are facing issues maybe due to this change, while the code is bundled using webpack.
`2024-01-25T08:09:47.489Z	85c2f086-1ee8-4b53-bbc7-7389da5df30d	ERROR	Error: Cannot find module '../564/index.js'
Require stack:
- /var/task/listEvents/index.js
- /var/runtime/index.mjs
    at Module._resolveFilename (node:internal/modules/cjs/loader:1144:15)
    at Module._load (node:internal/modules/cjs/loader:985:27)
    at Module.require (node:internal/modules/cjs/loader:1235:19)
    at require (node:internal/modules/helpers:176:18)
    at __webpack_require__.f.require (/var/task/listEvents/index.js:34787:28)
    at /var/task/listEvents/index.js:34726:40
    at Array.reduce (<anonymous>)
    at __webpack_require__.e (/var/task/listEvents/index.js:34725:67)
    at fromSSO.fromSSO (/var/task/listEvents/index.js:21092:59)
    at /var/task/listEvents/index.js:25522:39 {
  code: 'MODULE_NOT_FOUND',
  requireStack: [ '/var/task/listEvents/index.js', '/var/runtime/index.mjs' ]
}
`
Please let me know if there is a more appropriate place to post the issue...

kuhe:
@mskonf could you create a new bug report with instructions on how to reproduce the issue?

mskonf:
> @mskonf could you create a new bug report with instructions on how to reproduce the issue?

Sure

mskonf:
@kuhe have created the issue. This is my first time creating an issue, I have included all the context to the best of my ablity. Please let me know if I have left something out and if you need additional context on the same!
Thanks

perpil:
Not sure if this is the best place to comment, but I think this change was made in response to https://github.com/aws/aws-sdk-js-v3/issues/5516 which is locked so I update it.

In the lambda environment this didn't result in the performance gains in coldstarts I was expecting.  I updated my test repo: https://github.com/perpil/sso-coldstart-benchmark-cdk to use 3.499

The most surprising thing to me was the delta between the new provider and the new provider when the sso packages are excluded from the bundle so they would be loaded from disk.  I would expect the performance to be the same because the sso packages shouldn't be loaded, but there was a 285 ms delta seemingly indicating those packages are still getting used and being loaded from disk.

If I figure out what is going on, should I follow up here?

Issue #5721, opened 2024-01-24
Invoke Bedrock Agent not returning valid response


Dinuda:
### Checkboxes for prior research

- [X] I've gone through [Developer Guide](https://docs.aws.amazon.com/sdk-for-javascript/v3/developer-guide) and [API reference](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest)
- [X] I've checked [AWS Forums](https://forums.aws.amazon.com) and [StackOverflow](https://stackoverflow.com/questions/tagged/aws-sdk-js).
- [X] I've searched for [previous similar issues](https://github.com/aws/aws-sdk-js-v3/issues) and didn't find any solution.

### Describe the bug

It seems like the response returned from the agent is not what is expected to return. I have tried invoking the model and knowledge base seperately so I can verify that the `client` is configured correctly

### SDK version number

@aws-sdk/client-bedrock-agent-runtime": "^3.496.0"

### Which JavaScript Runtime is this issue in?

Node.js

### Details of the browser/Node.js/ReactNative version

v21.2.0

### Reproduction Steps

Here's my code so far:
```js
const { BedrockAgentRuntimeClient, InvokeAgentCommand } = require("@aws-sdk/client-bedrock-agent-runtime");

const config = {
    accessKeyId: process.env.AWS_BEDROCK_ACCESS_KEY_ID,
    secretAccessKey: process.env.AWS_BEDROCK_SECRET_ACCESS_KEY,
};


const client = new BedrockAgentRuntimeClient(config);

const input = { 
    agentId: "agentId", 
    agentAliasId: "agentAliasId", 
    sessionId: "abc123", 
    inputText: "Hello",
};

async invokeAgent() {
   return new Promise(async (resolve, reject) => {
      const command = new InvokeAgentCommand(input);
      const response = await client.send(command);
      resolve(response);
   });
}
```

### Observed Behavior

The response I get currently from the endpoint is:
```
{
    "$metadata": {
        "httpStatusCode": 200,
        "requestId": "reqId",
        "attempts": 1,
        "totalRetryDelay": 0
    },
    "contentType": "application/json",
    "sessionId": "abc123",
    "completion": {
        "options": {
            "messageStream": {
                "options": {
                    "inputStream": {},
                    "decoder": {
                        "headerMarshaller": {},
                        "messageBuffer": [],
                        "isEndOfStream": false
                    }
                }
            }
        }
    }
}
```

### Expected Behavior
```
[...]

PR #5729, opened 2024-01-26
docs(client-bedrock-agent-runtime): edit example for bare-bone client…

Dinuda:
… to invoke Agent API

### Issue
#5721 

### Description
In the example(invoke a bedrock agent using the API) that was given there was no high level documentation informing users about the need to properly handle the asynchronous nature of streaming data in this case.

### Testing
- Created a simple project which invokes the API correctly
- Ran package test using lerna

### Checklist

---


kuhe:
this is a generated file and cannot be edited like this. 

The relevant code is found in https://github.com/smithy-lang/smithy-typescript/blob/main/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/documentation/StructureExampleGenerator.java

Issue #5639, opened 2024-01-04
Requesting a signed url from MRAP returns a url where region is 'undefined' 


Giovanni-Ilacqua:
### Checkboxes for prior research

- [X] I've gone through [Developer Guide](https://docs.aws.amazon.com/sdk-for-javascript/v3/developer-guide) and [API reference](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest)
- [X] I've checked [AWS Forums](https://forums.aws.amazon.com) and [StackOverflow](https://stackoverflow.com/questions/tagged/aws-sdk-js).
- [X] I've searched for [previous similar issues](https://github.com/aws/aws-sdk-js-v3/issues) and didn't find any solution.

### Describe the bug

The getSignedUrl function from @aws-sdk/s3-request-presigner returns a url with region 'undefined' when pointing to an MRAP ARN.
This bug is related to an already closed issue:
https://github.com/aws/aws-sdk-js-v3/issues/4335#issuecomment-1589863800

A workaround was suggested, but a proper solution has not been released yet.

### SDK version number

@aws-sdk/s3-request-presigner: "3.470.0"

### Which JavaScript Runtime is this issue in?

Node.js

### Details of the browser/Node.js/ReactNative version

node v18.18.2

### Reproduction Steps

```
import { everything_relevant } from '@aws-sdk/various-sdks';

const MRAP_ARN = 'arn:aws:s3::0123456789:accesspoint/my_mrap_id.mrap';

const client = new S3Client({
  region: 'us-east-2,
  disableMultiregionAccessPoints: false,
  credentials: {
    accessKeyId: env.ACCESS_KEY_ID as string,
    secretAccessKey: env.SECRET_ACCESS_KEY_ID as string
  }
});

export async function getPreSignedURLToGetFile(key: string): Promise<string> {
  const command = new GetObjectCommand({
    Bucket: MRAP_ARN,
    Key: key,
  });

  const url = await getSignedUrl(client, command);
  return url;
}
```

### Observed Behavior

The generated url has an invalid X-Amz-Credential parameter, because the region is 'undefined'
When the signed url is accessed, the following XML error is returned:

```
<Error>
<Code>AuthorizationQueryParametersError</Code>
<Message>Error parsing the X-Amz-Credential parameter; the region 'undefined' is wrong; expecting 'us-east-2'</Message>
<Region>us-east-2</Region>
<RequestId>GFYNHFPQSME0S905</RequestId>
[...]

PR #5742, opened 2024-01-29
Request presigner mrap fix

RanVaknin:
### Issue
#5639 

### Description
When supplying an MRAP as a bucket the SDK should determine the authscheme as Sigv4a and set the signing region to be *. Before the change it was undefined.

### DIFF
--- a/packages/s3-request-presigner/src/getSignedUrl.spec.ts
+++ b/packages/s3-request-presigner/src/getSignedUrl.spec.ts
@@ -23,9 +23,24 @@ jest.mock("@aws-sdk/util-format-url", () => ({
   formatUrl: (url: any) => url,
 }));
 
+import { expectByte } from "@smithy/smithy-client";
 import { RequestPresigningArguments } from "@smithy/types";
 
 import { getSignedUrl } from "./getSignedUrl";
+
+jest.mock("@smithy/middleware-endpoint", () => {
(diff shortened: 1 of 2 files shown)


packages/s3-request-presigner/src/getSignedUrl.ts
```diff
@@ -56,9 +61,13 @@ export const getSignedUrl = async <
...
 
       let presigned: IHttpRequest;
+      let effectiveSigningRegion = options.signingRegion;
+      if (!effectiveSigningRegion) {
+        effectiveSigningRegion = context["signing_region"] ?? region;
+      }
       const presignerOptions = {
         ...options,
-        signingRegion: options.signingRegion ?? context["signing_region"],
+        signingRegion: effectiveSigningRegion,
```
kuhe, line 70: options.signingRegion ?? context["signing_region"] ?? region

kuhe (approved):

packages/s3-request-presigner/src/getSignedUrl.spec.ts
```diff
@@ -23,10 +23,25 @@ jest.mock("@aws-sdk/util-format-url", () => ({
   formatUrl: (url: any) => url,
 }));
 
+import { expectByte } from "@smithy/smithy-client";
```
kuhe, line 26: remove this

PR #5795, opened 2024-02-14
fix(client-ec2): Add missing value to NetworkInterfaceType

xXPOLYGONXx:
### Issue
No issue was created beforehand.

### Description
The Type "NetworkInterfaceType" of the ec2-client is missing the possible value "efs".
The value is listed in the [AWS API Reference](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_NetworkInterface.html#API_NetworkInterface) and is automatically used by all enis, created by an efs-volume.

This PR adds the value to the possible values.

### Testing
Only the tests running for this PR and no new ones.
Adding the additional property to the enum should not change any behaviour.

### Additional context
None

---

### DIFF
--- a/clients/client-ec2/src/models/models_1.ts
+++ b/clients/client-ec2/src/models/models_1.ts
@@ -11061,6 +11061,7 @@ export const NetworkInterfaceType = {
   aws_codestar_connections_managed: "aws_codestar_connections_managed",
   branch: "branch",
   efa: "efa",
+  efs: "efs",
   gateway_load_balancer: "gateway_load_balancer",
   gateway_load_balancer_endpoint: "gateway_load_balancer_endpoint",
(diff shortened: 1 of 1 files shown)


xXPOLYGONXx:
Do I need to do something more to get this merged? Is an issue required for pull requests?

kuhe:
We cannot accept this PR because the file you have edited is generated from a service model to all AWS SDKs of multiple programming languages.

Unfortunately, this is easy to miss, but it is a note at the top of the file. 

For your usage, please override the typechecking of the value at the call site.

```ts
const value = "efs" as unknown as NetworkInterfaceType;
```

The addition of this enum value must come from the EC2 service team via deploying a model update to the public SDKs.

Issue #5799, opened 2024-02-15
Inner STSClient doesn't longer use the provided region


Grmiade:
### Checkboxes for prior research

- [X] I've gone through [Developer Guide](https://docs.aws.amazon.com/sdk-for-javascript/v3/developer-guide) and [API reference](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest)
- [X] I've checked [AWS Forums](https://forums.aws.amazon.com) and [StackOverflow](https://stackoverflow.com/questions/tagged/aws-sdk-js).
- [X] I've searched for [previous similar issues](https://github.com/aws/aws-sdk-js-v3/issues) and didn't find any solution.

### Describe the bug

Since the version 3.502.0, the inner STSClient doesn't longer use the provided region in the DynamoDBClient instance when we use assume role with web identity.

### SDK version number

@aws-sdk/client-dynamodb@3.502.0

### Which JavaScript Runtime is this issue in?

Node.js

### Details of the browser/Node.js/ReactNative version

v20.9.0

### Reproduction Steps

- Configure envs like AWS_REGION (`eu-west-3` by example), AWS_ROLE_ARN, AWS_STS_REGIONAL_ENDPOINTS and AWS_WEB_IDENTITY_TOKEN_FILE to use assume role with web identity.
- Create a DynamoDBClient instance like `new DynamoDBClient()`
- Perform any command.

### Observed Behavior

The STSClient will use the `us-east-1` region by default.

### Expected Behavior

The STSClient should use the region specified in AWS_REGION by default (like it was before the version 3.502.0).

### Possible Solution

Downgrade to 3.501.0. 
Or specify explicitly the region with `new DynamoDBClient({ credentials: fromNodeProviderChain({ clientConfig: { region: process.env.AWS_REGION } }) })`.

### Additional Information/Context

Still doesn't work with `3.514.0`

kuhe:
Thanks for reporting this. I've confirmed this bug was a result of my misunderstanding of the defaulting behavior of the global fallback us-east-1. 

A fix was merged in #5800 with new tests.

The fix should be released in the next version expected later today as v3.515.0.

Grmiade:
Thanks @kuhe for the reactivity 👍

PR #5800, opened 2024-02-15
fix(client-sts): use outer client region before default STS global region

kuhe:
### Issue
https://github.com/aws/aws-sdk-js-v3/issues/5799

### Description
This fixes the behavior of the STS role assumer internal client to use the parent client's region if available in any region, not just non-AWS partitions.

### Testing
Added integ test case


kuhe:
merging since integ, e2e, e2e:legacy are passing

PR #5850, opened 2024-03-04
chore(codegen): add sigv4a trait detection

kuhe:
### Issue
n/a

### Description
Add detection of the sigv4a auth trait and use the multiregion signer if that is the case.

Some services are already using sigv4a (hard-coded) but do not have the trait. This PR does not remove sigv4a from those services.

### Testing
- [x] no effect on current code generation since the trait is not deployed to models

### DIFF
--- a/codegen/smithy-aws-typescript-codegen/src/main/java/software/amazon/smithy/aws/typescript/codegen/AddSigv4aPlugin.java
+++ b/codegen/smithy-aws-typescript-codegen/src/main/java/software/amazon/smithy/aws/typescript/codegen/AddSigv4aPlugin.java
@@ -10,11 +10,11 @@ import java.util.Collections;
 import java.util.Map;
 import java.util.Set;
 import java.util.function.Consumer;
+
+import software.amazon.smithy.aws.traits.auth.SigV4ATrait;
 import software.amazon.smithy.codegen.core.SymbolProvider;
 import software.amazon.smithy.model.Model;
 import software.amazon.smithy.model.knowledge.ServiceIndex;
-import software.amazon.smithy.model.knowledge.TopDownIndex;
-import software.amazon.smithy.model.shapes.OperationShape;
 import software.amazon.smithy.model.shapes.ServiceShape;
 import software.amazon.smithy.model.shapes.ShapeId;
 import software.amazon.smithy.model.traits.Trait;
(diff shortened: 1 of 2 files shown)


codegen/smithy-aws-typescript-codegen/src/main/java/software/amazon/smithy/aws/typescript/codegen/AddSigv4aPlugin.java
```diff
@@ -0,0 +1,67 @@
...
+ */
+@SmithyInternalApi
+public final class AddSigv4aPlugin implements TypeScriptIntegration {
+    public Map<String, Consumer<TypeScriptWriter>> getRuntimeConfigWriters(
+        TypeScriptSettings settings,
+        Model model,
+        SymbolProvider symbolProvider,
+        LanguageTarget target
+    ) {
+        boolean useSigv4aSigner = false;
```
kuhe, line 34: note, this does not mean to use sigv4a signing. It only means that a sigv4a-capable signer, that is also capable of sigv4, will be provided as the signer implementation.

The decision to use sigv4 vs sigv4a is still left to the endpoints 2.0 resolution's `authschemes` property.

codegen/smithy-aws-typescript-codegen/src/main/java/software/amazon/smithy/aws/typescript/codegen/AddSigv4aPlugin.java
```diff
@@ -0,0 +1,67 @@
...
+
+        ServiceIndex serviceIndex = ServiceIndex.of(model);
+        Set<ServiceShape> services = model.getServiceShapes();
+        for (ServiceShape service of services) {
+            Map<ShapeId, Trait> authSchemes = serviceIndex.getAuthSchemes(service);
+            useSigv4aSigner = useSigv4aSigner || authSchemes.values().stream()
+                .anyMatch(trait -> trait.toString().equals("aws.auth#sigv4a"));
```
kuhe, line 41: I'm using a string because there doesn't seem to be a SigV4ATrait class yet

codegen/smithy-aws-typescript-codegen/src/main/java/software/amazon/smithy/aws/typescript/codegen/AddSigv4aPlugin.java
```diff
@@ -0,0 +1,84 @@
...
+        if (!useSigv4aCapableSigner) {
+            return Collections.emptyMap();
+        }
+
+        switch (target) {
+            case SHARED:
+                return MapUtils.of("signerConstructor", writer -> {
+                    writer.addDependency(AwsDependency.SIGNATURE_V4_MULTIREGION)
+                        .addImport("SignatureV4MultiRegion", "SignatureV4MultiRegion",
+                                AwsDependency.SIGNATURE_V4_MULTIREGION)
```
syall, line 68: nit: I think the second `"SignatureV4MultiRegion"` is redundant

```suggestion
                        .addImport("SignatureV4MultiRegion", null,
                                AwsDependency.SIGNATURE_V4_MULTIREGION)
```

codegen/smithy-aws-typescript-codegen/src/main/java/software/amazon/smithy/aws/typescript/codegen/AddSigv4aPlugin.java
syall: The `SigV4ATrait` is available in the repo's smithy version 1.45.0: https://smithy.io/javadoc/1.45.0/software/amazon/smithy/aws/traits/auth/SigV4ATrait.html

codegen/smithy-aws-typescript-codegen/src/main/java/software/amazon/smithy/aws/typescript/codegen/AddSigv4aPlugin.java
syall: nit: can we rename this to `addSigv4aSignerToConfig` instead?

codegen/smithy-aws-typescript-codegen/src/main/java/software/amazon/smithy/aws/typescript/codegen/AddSigv4aPlugin.java
```diff
@@ -0,0 +1,84 @@
...
+            }
+        }
+
+        if (!useSigv4aCapableSigner) {
+            return Collections.emptyMap();
+        }
+
+        switch (target) {
+            case SHARED:
+                return MapUtils.of("signerConstructor", writer -> {
```
syall, line 65: Clarifying question: is it safe to include `SignatureV4MultiRegion` in the shared config?

codegen/smithy-aws-typescript-codegen/src/main/java/software/amazon/smithy/aws/typescript/codegen/AddSigv4aPlugin.java
```diff
@@ -0,0 +1,84 @@
...
+            }
+
+            TopDownIndex topDownIndex = TopDownIndex.of(model);
+            for (OperationShape operationShape : topDownIndex.getContainedOperations(service)) {
+                useSigv4aCapableSigner = usesSigv4a(operationShape.getAllTraits());
+                if (useSigv4aCapableSigner) {
+                    break;
+                }
+            }
+        }
```
syall, line 57: nit: I think a better way of checking this is:

1. If the service auth schemes don't include `sigv4a`, return empty
2. If all operation auth schemes don't have `sigv4a`, return empty
3. return `signerConstructor` with `SignatureV4MultiRegion`

This is because operation auth schemes are equal to or a subset of the service auth schemes.

syall (changes requested):

codegen/smithy-aws-typescript-codegen/src/main/java/software/amazon/smithy/aws/typescript/codegen/AddSigv4aPlugin.java
kuhe: yes, it throws a clear error in the browser and may be implemented in JS in the future

syall (approved):

syall (approved):

trivikr (approved):

Issue #5863, opened 2024-03-07
S3 SSECustomerKey no longer working since 3.498.0


john-yick:
### Checkboxes for prior research

- [X] I've gone through [Developer Guide](https://docs.aws.amazon.com/sdk-for-javascript/v3/developer-guide) and [API reference](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest)
- [X] I've checked [AWS Forums](https://forums.aws.amazon.com) and [StackOverflow](https://stackoverflow.com/questions/tagged/aws-sdk-js).
- [X] I've searched for [previous similar issues](https://github.com/aws/aws-sdk-js-v3/issues) and didn't find any solution.

### Describe the bug

Versions prior to 3.498.0 seem to work as expected for SSECustomerKey, but something within 3.498.0 has broken it for us.

I have crafted a small script below that works fine for 3.496.0 which we have running in production. But updating to any AWS SDK version higher than 3.496.0 causes an 403 Access Denied error to be thrown.

### SDK version number

@aws-sdk/client-s3@3.498.0

### Which JavaScript Runtime is this issue in?

Node.js

### Details of the browser/Node.js/ReactNative version

v20.7.0

### Reproduction Steps

Run the following code for a file that exists with SSECustomerKey encryption

Variables you will need to provide
- customerManagedKey
- key
- bucket

```js
const crypto = require("crypto");
const { S3Client, GetObjectCommand } = require("@aws-sdk/client-s3");
const { fromEnv } = require("@aws-sdk/credential-providers");

const SSECustomerAlgorithm = "AES256";
const SSECustomerKey = crypto
  .createHash("md5")
  .update(customerManagedKey)
  .digest("hex");
const SSECustomerKeyMD5 = crypto
  .createHash("md5")
  .update(SSECustomerKey)
  .digest("hex");

const s3Client = new S3Client({
  region: process.env.AWS_REGION,
  credentials: fromEnv(),
});

// Works under 3.496.0
// Broken under 3.498.0
async function main() {
  try {
    const s3GetObjectResponse = await s3Client.send(
      new GetObjectCommand({
        Bucket: bucket,
        Key: key,
        SSECustomerAlgorithm: SSECustomerAlgorithm,
        SSECustomerKey: SSECustomerKey,
        SSECustomerKeyMD5: SSECustomerKeyMD5,
      })
    );
    const data = new Promise((resolve, reject) => {
      s3GetObjectResponse.Body.pipe(res);
[...]

PR #5875, opened 2024-03-08
fix(middleware-ssec): ssecMiddleware with strict base64 validation

RanVaknin:
### Issue
#5863

### Description

Making a a more robust check for if the provided key is indeed a base64  by comparing the byte length to 32, since AES-256 specifies a cipher must be 256bits (32 bytes) we can assert that the cipher provided must have a byte length of 32.

> AES-256() The block cipher specifed in this Standard with 256-bit keys

https://csrc.nist.gov/pubs/fips/197/final

### Testing
Added unit test coverage to cover the more of the middleware changes from https://github.com/aws/aws-sdk-js-v3/pull/5676
And additionally unit tests to cover the `isValidBase64EncodedSSECustomerKey` helper function.

### Driver code:
Borrowed from edge case mentioned in #5863:

```js
async function putObjectWithMD5EncodedSSECKey(){
  try {
    const SSECustomerKey = crypto
    .createHash("md5")
    .update("foo")
    .digest("hex");
    const SSECustomerKeyMD5 = crypto
  .createHash("md5")
  .update(SSECustomerKey)
  .digest("hex");

    const response = await client.send(new PutObjectCommand({
      Bucket: 'testbucket',
       Body: 'This is a test',      
       Key: "foo",    
       SSECustomerKey: SSECustomerKey,
      SSECustomerAlgorithm: 'AES256',
      SSECustomerKeyMD5: SSECustomerKeyMD5
    }))
    console.log(response.$metadata.httpStatusCode);
  } catch (error) {
    console.error(error);
  }
}

// 200
```
### Additional context

### Checklist

---

### DIFF
--- a/packages/middleware-ssec/src/index.ts
+++ b/packages/middleware-ssec/src/index.ts
@@ -39,8 +39,7 @@ export function ssecMiddleware(options: PreviouslyResolved): InitializeMiddlewar
         if (value) {
           let valueForHash: Uint8Array;
           if (typeof value === "string") {
-            const isBase64Encoded = /^(?:[A-Za-z0-9+/]{4})*([A-Za-z0-9+/]{2}==|[A-Za-z0-9+/]{3}=)?$/.test(value);
-            if (isBase64Encoded) {
+            if (isValidBase64(value)) {
               valueForHash = options.base64Decoder(value);
             } else {
               valueForHash = options.utf8Decoder(value);
@@ -78,3 +77,19 @@ export const getSsecPlugin = (config: PreviouslyResolved): Pluggable<any, any> =
     clientStack.add(ssecMiddleware(config), ssecMiddlewareOptions);
   },
 });
+
+function isValidBase64(str: string) {
+  const base64Regex = /^(?:[A-Za-z0-9+/]{4})*([A-Za-z0-9+/]{2}==|[A-Za-z0-9+/]{3}=)?$/;
+
+  if (!base64Regex.test(str)) return false;
+
+  try {
+    const decodedBytes = Buffer.from(str, "base64");
+    if (decodedBytes.length !== 32) return false;
+
+    const reEncoded = decodedBytes.toString("base64");
+    return str === reEncoded;
+  } catch {
+    return false;
+  }
+}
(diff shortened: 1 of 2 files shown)


packages/middleware-ssec/src/index.ts
```diff
@@ -78,3 +77,19 @@ export const getSsecPlugin = (config: PreviouslyResolved): Pluggable<any, any> =
...
   },
 });
+
+function isValidBase64(str: string) {
+  const base64Regex = /^(?:[A-Za-z0-9+/]{4})*([A-Za-z0-9+/]{2}==|[A-Za-z0-9+/]{3}=)?$/;
+
+  if (!base64Regex.test(str)) return false;
+
+  try {
+    const decodedBytes = Buffer.from(str, "base64");
```
kuhe, line 87: encoding/decoding must use the SerdeContext's encoder/decoders, i.e. `options` above. 

Buffer is not available in browsers.

packages/middleware-ssec/src/index.ts
```diff
@@ -78,3 +77,19 @@ export const getSsecPlugin = (config: PreviouslyResolved): Pluggable<any, any> =
     clientStack.add(ssecMiddleware(config), ssecMiddlewareOptions);
   },
 });
+
+function isValidBase64(str: string) {
```
kuhe, line 81: This function should be named after its logical behavior. It is not checking base64 validity.

It is checking whether the input is a base64 encoded SSE customer key.

packages/middleware-ssec/src/index.ts
```diff
@@ -78,3 +77,16 @@ export const getSsecPlugin = (config: PreviouslyResolved): Pluggable<any, any> =
     clientStack.add(ssecMiddleware(config), ssecMiddlewareOptions);
   },
 });
+
+function isValidBase64EncodedSSECustomerKey(str: string, options: PreviouslyResolved) {
```
kuhe, line 81: please add two unit or integration tests, one with a passing example and one with a failing example

kuhe (approved):

PR #5893, opened 2024-03-14
fix(ec2): handle unauthorized exception

DoctorFTB:
### Issue

### Description

Set FAILURE state if we doesn't have access for operation

### Testing
How was this change tested?

- Create access key for user which doesn't have any access
- Run waitUntilInstanceRunning function
- Got timeout, because always retry status

### Additional context

### Checklist

---


kuhe:
the waiter code generation is from https://github.com/smithy-lang/smithy-typescript and cannot be modified here directly

this applies for any code file having `// smithy-typescript generated code` at the top.

PR #5895, opened 2024-03-14
fix(middleware-signing): attempt secondary authscheme selection during request signing

kuhe:
### Issue
n/a

### Description
In awsAuthMiddleware, select the second authScheme if:
- second auth scheme exists
- first auth scheme is sigv4a
- sigv4a is not available

### Testing
existing unit test

- [x] e2e
- [x] e2e:legacy

### DIFF
--- a/packages/middleware-signing/src/awsAuthConfiguration.ts
+++ b/packages/middleware-signing/src/awsAuthConfiguration.ts
@@ -209,7 +209,7 @@ export const resolveAwsAuthConfig = <T>(
         input.signingRegion = originalInputSigningRegion || signingRegion;
         input.signingName = originalInputSigningName || signingService || input.serviceId;
       } else {
-        // update client's singing region and signing service config if they are resolved.
+        // update client's signing region and signing service config if they are resolved.
         // signing region resolving order: user supplied signingRegion -> endpoints.json inferred region -> client region
         input.signingRegion = input.signingRegion || signingRegion;
         // signing name resolving order:
--- a/packages/middleware-signing/src/awsAuthConfiguration.ts
+++ b/packages/middleware-signing/src/awsAuthConfiguration.ts
@@ -182,7 +182,15 @@ export const resolveAwsAuthConfig = <T>(
     // This branch is for endpoints V2.
     // Handle endpoints v2 that resolved per-command
     // TODO: need total refactor for reference auth architecture.
-    signer = async (authScheme?: AuthScheme) => {
+    const originalInputSigningRegion = input.signingRegion;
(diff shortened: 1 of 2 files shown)


packages/middleware-signing/src/awsAuthConfiguration.ts
```diff
@@ -196,12 +204,18 @@ export const resolveAwsAuthConfig = <T>(
...
-      input.signingRegion = input.signingRegion || signingRegion;
-      // signing name resolving order:
-      // user supplied signingName -> endpoints.json inferred (credential scope -> model arnNamespace) -> model service id
-      input.signingName = input.signingName || signingService || input.serviceId;
+
+      if (overwrite) {
+        input.signingRegion = originalInputSigningRegion || signingRegion;
+        input.signingName = originalInputSigningName || signingService || input.serviceId;
+      } else {
+        // update client's singing region and signing service config if they are resolved.
```
syall, line 212:
```suggestion
        // update client's signing region and signing service config if they are resolved.
```

packages/middleware-signing/src/awsAuthConfiguration.ts
```diff
@@ -196,12 +204,18 @@ export const resolveAwsAuthConfig = <T>(
...
       const signingRegion = authScheme.signingRegion;
       const signingService = authScheme.signingName;
-      // update client's singing region and signing service config if they are resolved.
-      // signing region resolving order: user supplied signingRegion -> endpoints.json inferred region -> client region
-      input.signingRegion = input.signingRegion || signingRegion;
-      // signing name resolving order:
-      // user supplied signingName -> endpoints.json inferred (credential scope -> model arnNamespace) -> model service id
-      input.signingName = input.signingName || signingService || input.serviceId;
+
+      if (overwrite) {
```
syall, line 208: nit: `overwrite` reads like the `input.*` parameters would be overwritten, but the code path here is using the original `input.*` parameters.

kuhe: unfortunately this part of the code mutates the config during signing and not during the constructor config resolution

syall (approved):

trivikr (approved):

packages/middleware-signing/src/awsAuthConfiguration.ts
kuhe: I changed it to only mutate if authscheme is not sigv4a. Sigv4a doesn't contain a meaningful region so far, so we can skip writing that to the input and leave it to any further resolutions containing a singular region.

Issue #5878, opened 2024-03-10
MIGRATION ISSUE: SDK v2 Credentials uses `expireTime` and not `expiration`


matan-mitiga:
### Pre-Migration Checklist

- [X] I've read the [Migration Guide](https://docs.aws.amazon.com/sdk-for-javascript/v3/developer-guide/migrating-to-v3.html).
- [X] I've reviewed the upgrading notes and major version differences mentioned in [`UPGRADING.md`](https://github.com/aws/aws-sdk-js-v3/blob/main/UPGRADING.md).
- [X] I've checked [AWS Forums](https://forums.aws.amazon.com) and [StackOverflow](https://stackoverflow.com/questions/tagged/aws-sdk-js) for similar migration issues.

### Which JavaScript Runtime is this issue in?

Node.js (includes AWS Lambda)

### AWS Lambda Usage

- [ ] Yes, my application is running on AWS Lambda.
- [x] No, my application is not running on AWS Lambda.

### Describe the Migration Issue

Our app has mixed v2 and v3 usages. Our internal authentication library uses v2 to create a `Credentials` object. Sometimes we receive `TokenExpired` errors. When I looked through the V3 code in I noticed that it's expecting that the `Credentials` object will have an `expiration` attribute:

`/packages/middleware-signing/src/awsAuthConfiguration.ts`
line 265
```
const normalizeCredentialProvider = (
  credentials: AwsCredentialIdentity | Provider<AwsCredentialIdentity>
): MemoizedProvider<AwsCredentialIdentity> => {
  if (typeof credentials === "function") {
    return memoize(
      credentials,
      (credentials) =>
        credentials.expiration !== undefined &&
        credentials.expiration.getTime() - Date.now() < CREDENTIAL_EXPIRE_WINDOW,
      (credentials) => credentials.expiration !== undefined
    );
  }
  return normalizeProvider(credentials);
};
```

The V2 Credentials object has a different attribute called `expireTime`. Hence, to my understanding, the token refresh fails.

Will appreciate your advice.

Thank you!

### Observed Differences/Errors

V2 `Credentials` uses `expireTime: Date`
V3 `Identity` uses `expiration: Date`

matan-mitiga:
I noticed another issue was opened in the past to rename and had been closed:
https://github.com/aws/aws-sdk-js-v3/issues/1037

RanVaknin:
HI @matan-mitiga ,

[...]

PR #5900, opened 2024-03-14
docs: adding upgrading notes pertaining to the cred object

RanVaknin:
### Issue
#5878 

### Description
The structure of the expected credential object has changed between v2 and v3. We need to document it.

---


UPGRADING.md
```diff
@@ -35,9 +35,43 @@ This list is indexed by [v2 config parameters](https://docs.aws.amazon.com/AWSJa
...
+      secretAccessKey: string;
+      expiration?: Date;
+      sessionToken?: string;
+      credentialScope?: string;
+      */
+    },
+  });
+  ```
+
+  It can also be an async function that returnsb the credentials object.
```
trivikr, line 71:
```suggestion
  It can also be an async function that returns the credentials object.
```

RanVaknin: good catch

Issue #5416, opened 2023-10-25
May 1, 2024: end of support for Node.js 14.x in AWS SDK for JavaScript (v3) 


kuhe:
We are planning to end support for Node.js 14.x on May 1, 2024. Please see this blog post for the official announcement: https://aws.amazon.com/blogs/developer/announcing-the-end-of-support-for-node-js-14-x-in-the-aws-sdk-for-javascript-v3/

kuhe:
Due to limitations described in https://github.com/aws/aws-sdk-js-v3/issues/5821, the end of support for Node.js 14.x within the AWS SDK for JavaScript (v3) will be released as a minor version change.

PR #5923, opened 2024-03-19
chore(external): add new package for external dependencies

kuhe:
### Issue
similar to https://github.com/smithy-lang/smithy-typescript/pull/1214

### Description
Creates an exporter for external packages that can conditionally bundle them for Node.js while retaining no-op alias behavior for browser bundlers.

### Testing
- [ ] e2e testing
- [ ] Node.js 14 testing if merging before end of support
- [ ] bundler testing

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -5,8 +5,8 @@
     "esModuleInterop": true,
     "incremental": true,
     "lib": ["es2015", "dom"],
-    "module": "commonjs",
-    "moduleResolution": "node",
+    "module": "Node16",
+    "moduleResolution": "Node16",
     "noFallthroughCasesInSwitch": true,
     "paths": {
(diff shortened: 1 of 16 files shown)


tsconfig.json
```diff
@@ -5,8 +5,8 @@
     "esModuleInterop": true,
     "incremental": true,
     "lib": ["es2015", "dom"],
-    "module": "commonjs",
-    "moduleResolution": "node",
+    "module": "Node16",
+    "moduleResolution": "Node16",
```
kuhe, line 9: This does not yet signal dropping Node.js 14, which is scheduled in https://github.com/aws/aws-sdk-js-v3/issues/5416.

This is needed for the TypeScript compiler to understand the `exports` block of the `@aws-sdk/external` package. In Node.js this was added in v12 already.

I will test to confirm this behavior.

PR #5926, opened 2024-03-21
feat(cloudfront-signer): allow `url` to be optional when using `policy`

NamesMT:
### Issue
Resolves #5836
Related #4516

### Description
`getSignedUrl()` now automatically gets the `url` from the `policy`'s Resource, allowing the `url` to be optionally typed when we're using `InputWithPolicy`.
Types are also updated.

### Testing
I copied the package locally into a workspace project to test.

### Additional context
The issue #5836 talks about Yenfry's answer in #4516, in which, functionally, his example answer with `getSignedCookies()` works, but reporting types error about `url` parameter missing.
This PR fixes the issue and also improving `getSignedUrl()` to have the `url` optionally inputted (when using with `policy`).

---
By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #5932, opened 2024-03-21
fix(cloudfront-signer): in getSignedCookies url and policy shouldn't be passed together

nayabatir1:
### Issue

### Description

### Testing
How was this change tested?

### Additional context

### Checklist

---

### DIFF
--- a/packages/cloudfront-signer/README.md
+++ b/packages/cloudfront-signer/README.md
@@ -47,3 +47,73 @@ const cookies = getSignedCookies({
   privateKey,
 });
 ```
+
+### Sign a url with Policy
+
+```javascript
+import { getSignedUrl } from "@aws-sdk/cloudfront-signer"; // ESM
+// const { getSignedUrl } = require("@aws-sdk/cloudfront-signer"); // CJS
+
+const cloudfrontDistributionDomain = "https://d111111abcdef8.cloudfront.net";
+const s3ObjectKey = "private-content/private.jpeg";
+const url = `${cloudfrontDistributionDomain}/${s3ObjectKey}`;
+const privateKey = "CONTENTS-OF-PRIVATE-KEY";
+const keyPairId = "PUBLIC-KEY-ID-OF-CLOUDFRONT-KEY-PAIR";
+const dateLessThan = "2022-01-01";
+
+const policy = {
+  Statement: [
+    {
+      Resource: url,
+      Condition: {
+        DateLessThan: {
+          "AWS:EpochTime": new Date(dateLessThan).getTime() / 1000, // time in seconds
+        },
+      },
+    },
+  ],
+};
+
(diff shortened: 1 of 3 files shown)


NamesMT:
IMO the `url` shouldn't be `never` and should not always be overwritten `if (policy)`, in some current complex use cases (my company internal included), the signedUrl could contain a policy that allows a flexible (wildcard) resource target, and we still could pass in the `url` param to get what we want as an initial URL,
i.e ignoring `url` when used with `policy` would be a breaking change.

nayabatir1:
> IMO the `url` shouldn't be `never` and should not always be overwritten `if (policy)`, in some current complex use cases (my company internal included), the signedUrl could contain a policy that allows a flexible (wildcard) resource target, and we still could pass in the `url` param to get what we want as an initial URL,
> i.e ignoring `url` when used with `policy` would be a breaking change.

Policy also has url, I'm extracting url from policy if it exists. You don't have to provide url as an additional param.

NamesMT:
I know the policy has url (Resource), but there is cases where the extracted Resource would introduce a breaking change, assuming the policy's Resource is: `https://example.org/subFolder/*`, currently we're using the `url` to set the initial URL to be `https://example.org/subFolder/mainFile.js`, by migrating this PR that would broke our app.
I do like your implementation. and the initial URL can very easily be changed by the user's requirement but breaking change to the SDK isn't good.

There's also another issue that would throw an error internally at `new URL()`, with wildcard protocol Resource: `*://example.com/file.ext`

nayabatir1:
As per AWS [doc](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/private-content-setting-signed-cookie-custom-policy.html#:~:text=Note%20the%20following%3A-,Protocol,-%E2%80%93%20The%20value%20must) protocal must be `http://` or `http://`. 
It also mentions that `url`, `DateLessThan`, `DateGreaterThan`, `IpAddress` can be provided in policy. It no where mentions to pass url along with policy.

Incase this is merged existing version won't be modified. Same version won't be breaking for you.

Let see what maintainers have to say about this.

NamesMT:
That's fun, the AWS doc for [signed URL](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/private-content-creating-signed-url-custom-policy.html#:~:text=Note%20the%20following%3A-,Protocol,-%E2%80%93%20The%20value%20must) do allow `*://`, and one of our service is using this protocol, however the service is using AWS-SDK v2: 
image

>Incase this is merged existing version won't be modified. Same version won't be breaking for you.

Would it not be modified?, I thought the `aws-sdk` version on Lambda would be up-to-date, still, if the SDK version on Lambda do follows the lockfile, a not thoroughly tested dependencies update would then break things.

BTW I added you as co-author on new some updates in #5926

PR #5963, opened 2024-04-04
chore(codegen): move s3 transform to Java code

kuhe:
### Issue
This converts the s3 expires/expiresString custom model transform from JavaScript to Java.

This makes it easier for consumers of `smithy-typescript-aws-codegen` to use the transform.

### DIFF
--- a/codegen/smithy-aws-typescript-codegen/src/main/java/software/amazon/smithy/aws/typescript/codegen/AddS3Config.java
+++ b/codegen/smithy-aws-typescript-codegen/src/main/java/software/amazon/smithy/aws/typescript/codegen/AddS3Config.java
@@ -26,6 +26,8 @@ import java.util.Map;
 import java.util.Set;
 import java.util.function.Consumer;
 import java.util.logging.Logger;
+import java.util.stream.Stream;
+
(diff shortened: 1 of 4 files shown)


codegen/smithy-aws-typescript-codegen/src/main/java/software/amazon/smithy/aws/typescript/codegen/AddS3Config.java
```diff
@@ -106,6 +113,89 @@ public Model preprocessModel(Model model, TypeScriptSettings settings) {
...
+                                    m.addTraits(memberShape.getAllTraits().values());
+                                });
+                        }
+                    }
+                    modelBuilder
+                        .addShape(structureShapeBuilder.build());
+                }
+            }
+        }
+
```
kuhe, line 198: This code should initially have been in Java anyway. However, at the time I couldn't figure out how to mutate the model, and wrote it in JS.

It turns out the `.addX()` methods on the various builders generally do a replace on a map collection, and it was easier to figure out this time.

trivikr (approved):

PR #5994, opened 2024-04-11
chore(codegen): fix codegen release

milesziemer:
Jreleaser has to work in a git repository, and uses the jgit java library to interact with git. When jgit tries to open the git repo, it tries to open `codegen/config` as a file, which blows up. I don't know of any way to configure what jgit is doing here, so I renamed the directory and it solves the issue. Also had to set `gitRootSearch` so jreleaser walks up the parent directories to find the git root.

---


syall (approved):

kuhe (approved):
for the record, this PR represents like 4 software developer hours of debugging

PR #6071, opened 2024-05-07
test: fix type checks in unit tests

trivikr:
### Issue
Unblocks https://github.com/aws/aws-sdk-js-v3/pull/6038

### Description
Fixes type checks in unit tests

### Testing
All tests except `@ts-expect-error` for ES2020 are successful in https://github.com/aws/aws-sdk-js-v3/pull/6052
The `@ts-expect-error` comments will be removed on bumping TSConfig

---

### DIFF
--- a/packages/credential-provider-process/src/getValidatedProcessCredentials.spec.ts
+++ b/packages/credential-provider-process/src/getValidatedProcessCredentials.spec.ts
@@ -18,7 +18,7 @@ describe(getValidatedProcessCredentials.name, () => {
     Expiration: mockExpiration,
   });
 
-  it.each([undefined, 2])("throws Error when Version is %s", (Version) => {
+  it.each([2])("throws Error when Version is %s", (Version) => {
     expect(() => {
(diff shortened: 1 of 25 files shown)


packages/credential-provider-process/src/getValidatedProcessCredentials.spec.ts
```diff
@@ -18,7 +18,7 @@ describe(getValidatedProcessCredentials.name, () => {
     Expiration: mockExpiration,
   });
 
-  it.each([undefined, 2])("throws Error when Version is %s", (Version) => {
+  it.each([2])("throws Error when Version is %s", (Version) => {
```
kuhe, line 21: is this lost case not useful?

kuhe (approved):

packages/credential-provider-process/src/getValidatedProcessCredentials.spec.ts
trivikr: The case is not supported as per TypeScript type definitions.

Issue #6042, opened 2024-05-01
MetadataService ignoring timeout config


morwoen:
### Checkboxes for prior research

- [X] I've gone through [Developer Guide](https://docs.aws.amazon.com/sdk-for-javascript/v3/developer-guide) and [API reference](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest)
- [X] I've checked [AWS Forums](https://forums.aws.amazon.com) and [StackOverflow](https://stackoverflow.com/questions/tagged/aws-sdk-js).
- [X] I've searched for [previous similar issues](https://github.com/aws/aws-sdk-js-v3/issues) and didn't find any solution.

### Describe the bug

The MetadataService from @aws-sdk/ec2-metadata-service package ignores timeout config. The config is [initialised in the constructor](https://github.com/aws/aws-sdk-js-v3/blob/main/packages/ec2-metadata-service/src/MetadataService.ts#L25) but is never retrieved from the object [here](https://github.com/aws/aws-sdk-js-v3/blob/main/packages/ec2-metadata-service/src/MetadataService.ts#L35) or [here](https://github.com/aws/aws-sdk-js-v3/blob/main/packages/ec2-metadata-service/src/MetadataService.ts#L85).

### SDK version number

@aws-sdk/ec2-metadata-service@3.565.0

### Which JavaScript Runtime is this issue in?

Node.js

### Details of the browser/Node.js/ReactNative version

v18.18.2

### Reproduction Steps

```
const metaData = new MetadataService({
    httpOptions: {
      timeout: 2000
    },
  });

return metaData.request('/latest/meta-data/placement/region', {})
```

The above has to be run locally where there is no metadata service access.

### Observed Behavior

Code hangs for over 30 seconds

### Expected Behavior

After 2 seconds the code should return a rejected promise

### Possible Solution

The config needs to be passed down to the HttpRequest.

RanVaknin:
Hi @morwoen ,

Thanks for letting us know. This seems like an oversight. @siddsriv is taking a look.

All the best,
Ran~

siddsriv:
thanks for reporting this, timeout config option been added for requests. closing this issue, feel free to create a new one if you encounter any other issues.

morwoen:
Thanks @siddsriv , could you confirm which version the fix is in?

morwoen:
Found it :) https://github.com/aws/aws-sdk-js-v3/releases/tag/v3.574.0

PR #6072, opened 2024-05-07
fix(ec2-metadata-service): set timeout for requests

siddsriv:
### Issue
#6042 

### Description
set the timeout for requests to ec2-mds.

### Testing
E2E tests for the package pass. 
```console

 PASS  src/MetadataService.e2e.spec.ts
  MetadataService E2E Tests
    ✓ should fetch metadata token successfully (6 ms)
    ✓ should fetch metadata successfully with token (4 ms)
    ✓ should fetch metadata successfully (without token -- disableFetchToken set to true) (2 ms)
    ✓ should handle TimeoutError by falling back to IMDSv1 (3 ms)
    ✓ should handle specific error codes by falling back to IMDSv1 (6 ms)
    ✓ should timeout as expected when a request exceeds the specified duration (4 ms)

Test Suites: 1 passed, 1 total
Tests:       6 passed, 6 total
Snapshots:   0 total
Time:        3.358 s, estimated 4 s
Ran all test suites.
Done in 4.34s.
```

### Checklist

---

### DIFF
--- a/packages/ec2-metadata-service/src/MetadataService.ts
+++ b/packages/ec2-metadata-service/src/MetadataService.ts
@@ -32,8 +32,10 @@ export class MetadataService {
   }
 
   async request(path: string, options: { method?: string; headers?: Record<string, string> }): Promise<string> {
-    const { endpoint, ec2MetadataV1Disabled } = await this.config;
-    const handler = new NodeHttpHandler();
+    const { endpoint, ec2MetadataV1Disabled, httpOptions } = await this.config;
+    const handler = new NodeHttpHandler({
+      requestTimeout: httpOptions?.timeout,
+    });
     const endpointUrl = new URL(endpoint!);
     const headers = options.headers || {};
     /**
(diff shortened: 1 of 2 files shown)


packages/ec2-metadata-service/src/MetadataService.ts
```diff
@@ -32,8 +32,10 @@ export class MetadataService {
   }
 
   async request(path: string, options: { method?: string; headers?: Record<string, string> }): Promise<string> {
-    const { endpoint, ec2MetadataV1Disabled } = await this.config;
-    const handler = new NodeHttpHandler();
+    const { endpoint, ec2MetadataV1Disabled, httpOptions } = await this.config;
+    const handler = new NodeHttpHandler({
+      requestTimeout: httpOptions?.timeout,
```
kuhe, line 37: should this timeout also be passed to connectionTimeout?

siddsriv: functionally, the timeout error should be thrown with just `requestTimeout`, but we can add `connectionTimeout` as well that'll pick up the same value from options. it's better to catch the two different timeouts.

kuhe (approved):

Issue #5964, opened 2024-04-04
Multipart upload doesn't throw or reject when max upload parts exceeded


piurafunk:
### Checkboxes for prior research

- [X] I've gone through [Developer Guide](https://docs.aws.amazon.com/sdk-for-javascript/v3/developer-guide) and [API reference](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest)
- [X] I've checked [AWS Forums](https://forums.aws.amazon.com) and [StackOverflow](https://stackoverflow.com/questions/tagged/aws-sdk-js).
- [X] I've searched for [previous similar issues](https://github.com/aws/aws-sdk-js-v3/issues) and didn't find any solution.

### Describe the bug

Exceeding `Upload.MAX_PARTS` doesn't throw an exception completely out

### SDK version number

@aws-sdk/client-s3@3.549.0

### Which JavaScript Runtime is this issue in?

Node.js

### Details of the browser/Node.js/ReactNative version

v20.12.0

### Reproduction Steps

```node
import { S3 } from '@aws-sdk/client-s3'
import { Upload } from '@aws-sdk/lib-storage'

const s3 = new S3()
const bucket = 'bucket' // Update this to your bucket name

// The upload must be more than 5242880000 bytes (5 MiB x 10000 parts)
// To pass this via stdin, you can use the following command:
// $ dd if=/dev/urandom bs=5242880 count=11000 | node repro.mjs

// If you have pipeviewer installed, you can use the following command to see the progress:
// $ dd if=/dev/urandom bs=5242880 count=11000 | pv -s $((5242880 * 11000)) | node repro.mjs
const upload = new Upload({
    params: {
        Bucket: bucket,
        Key: 'bigfile.txt',
        Body: process.stdin,
    },
    client: s3,
    queueSize: 50,
})

await upload.done()
```

### Observed Behavior

```
aws --profile sso-staging s3api get-object-attributes --bucket <bucket> --key bigfile.txt --object-attributes ObjectSize
{
    "LastModified": "2024-04-04T20:46:13+00:00",
    "VersionId": "lAYUcGXinSavYWPxPLsbfTFRWoD_UT5s",
    "ObjectSize": 52428800000
}
```

The object size is actually exactly 5 MiB x 10,000 parts: 52428800000 bytes. Notably, it also correctly completes the multi-part upload, so it really does look like it completes.

[...]

PR #6112, opened 2024-05-20
fix(lib-storage): call AbortMultipartUpload when failing to CompleteMultipartUpload

kuhe:
### Issue
related to https://github.com/aws/aws-sdk-js-v3/pull/4414
- [x] include fix for https://github.com/aws/aws-sdk-js-v3/issues/5964 as well

### Description
This PR modifies the behavior of `@aws-sdk/lib-storage`'s `Upload` class' `leavePartsOnError: boolean` to do what it sounds like it does instead of something meaningless.

Previous to this PR:
- when true, any error encountered during MPU would be thrown
- when false, any error encountered during MPU would be ignored
- default was false

As of this PR:
- when true, AbortMultipartUpload is not called in the event of a failed MPU
- when false, AbortMultipartUpload is called in the event of a failed MPU to mark the parts for cleanup as recommended by S3
- default is still false

### Testing
new e2e test

### Additional context
Portion of docs from AWS SDK JSv2:
```
 * ## Handling Multipart Cleanup
 *
 * By default, this class will automatically clean up any multipart uploads
 * when an individual part upload fails. This behavior can be disabled in order
 * to manually handle failures by setting the `leavePartsOnError` configuration
 * option to `true` when initializing the upload object.
```
behavior (JSv2 source code):
```js
    if (self.service.config.params.UploadId && !self.leavePartsOnError) {
      self.service.abortMultipartUpload().send();
    } else if (self.leavePartsOnError) {
      self.isDoneChunking = false;
    }
```

This behavior appears to have been lost in https://github.com/aws/aws-sdk-js-v3/pull/2039

### Checklist

### DIFF
--- a/lib/lib-storage/src/Upload.ts
+++ b/lib/lib-storage/src/Upload.ts
@@ -37,9 +37,9 @@ const MIN_PART_SIZE = 1024 * 1024 * 5;
 
 export class Upload extends EventEmitter {
   /**
-   * S3 multipart upload does not allow more than 10000 parts.
+   * S3 multipart upload does not allow more than 10,000 parts.
    */
-  private MAX_PARTS = 10000;
+  private MAX_PARTS = 10_000;
 
   // Defaults.
   private readonly queueSize: number = 4;
@@ -61,6 +61,7 @@ export class Upload extends EventEmitter {
   private abortMultipartUploadCommand: AbortMultipartUploadCommand | null = null;
 
   private uploadedParts: CompletedPart[] = [];
+  private uploadEnqueuedPartsCount = 0;
   /**
    * Last UploadId if the upload was done with MultipartUpload and not PutObject.
    */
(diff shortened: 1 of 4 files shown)


lib/lib-storage/src/Upload.ts
```diff
@@ -207,17 +208,17 @@ export class Upload extends EventEmitter {
 
   private async __doConcurrentUpload(dataFeeder: AsyncGenerator<RawDataPart, void, undefined>): Promise<void> {
     for await (const dataPart of dataFeeder) {
-      if (this.uploadedParts.length > this.MAX_PARTS) {
+      if (this.uploadEnqueuedPartsCount > this.MAX_PARTS) {
```
kuhe, line 211: reason for this change: `this.uploadedParts.length` is updated after an async operation (UploadPart), which causes a delay. 

Due to request concurrency, `this.uploadedParts.length` can be far behind the actual number of parts that have been iterated and submitted.

trivikr (approved):

Issue #4767, opened 2023-05-28
Request hangs for node-style Readable streams when ContentLength present as an Upload option


ferbs:
### Checkboxes for prior research

- [X] I've gone through [Developer Guide](https://docs.aws.amazon.com/sdk-for-javascript/v3/developer-guide) and [API reference](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest)
- [X] I've checked [AWS Forums](https://forums.aws.amazon.com) and [StackOverflow](https://stackoverflow.com/questions/tagged/aws-sdk-js).
- [X] I've searched for [previous similar issues](https://github.com/aws/aws-sdk-js-v3/issues) and didn't find any solution.

### Describe the bug

I'm streaming an http request body to S3 through lib-storage's Upload class. The proxied upload fails when the actual size exceeds the partSize (5mb by default) AND a Content-Length header is included as an upload option. When this happens, the upload silently hangs until socket timeout or abort signal. (Ironically, it doesn't occur when original request uses Transfer-Encoding chunked since that won't have a content length.)

The problem is easy to work around once understood but impact is severe and it's not easy to identify. 

### SDK version number

@aws-sdk/lib-storage@3.331.0

### Which JavaScript Runtime is this issue in?

Node.js

### Details of the browser/Node.js/ReactNative version

Node v18

### Reproduction Steps

Set Body to a Readable and include a ContentLength value for an upload larger than 5mb / partSize. Eg:

```javascript
const uploader = new Upload({
    client,
    params: {
      Bucket: 'some-bucket',
      Key: 'some-key,
      ContentType: 'some-mime',
      ContentLength: koaCtx.request.length,
      Body: koaCtx.req,
    },
  });
```

### Observed Behavior

Request hangs until socket timeout 

When logging the request headers in sdk middleware it looks like the full/total object size is applied to each part's Content-Length request header

### Expected Behavior

Upload success

### Possible Solution

Adding a validation check would be a decent enough solution, as would removing ContentLength as an Upload option, or best yet, correctly setting the request header for each part.

yenfryherrerafeliz:
Hi @ferbs, sorry for the late response. I agree that it hangs when using the Upload class and providing the ContentLength parameter. Also the following comment is correct:
> When logging the request headers in sdk middleware it looks like the full/total object size is applied to each part's Content-Length request header

This behavior comes from nodejs http lib which basically hangs when the content-length header value is greater than the length of the provided body.

[...]

PR #6131, opened 2024-05-23
Specify ContentLength for part uploads

f8k8:
### Issue
#4767 

### Description
The `UploadPartCommand` is passed the original parameters, which may include `ContentLength`. If the `ContentLength` is included in the original parameters, and the upload is larger than a single chunk, `UploadPartCommand` has an incorrect value for `ContentLength`. This change passes the actual size of the chunk as `ContentLength`, or makes sure it is undefined if the part size is not calculated.

### Testing
After upgrading from SDK v2 to v3, our uploads were hanging and eventually timing out. Making this change in our local copy of the SDK fixed the uploads.

### Checklist

---

### DIFF
--- a/lib/lib-storage/src/Upload.ts
+++ b/lib/lib-storage/src/Upload.ts
@@ -276,7 +276,9 @@ export class Upload extends EventEmitter {
       const partResult = await this.client.send(
         new UploadPartCommand({
           ...this.params,
-          ContentLength: partSize || undefined,
+          // dataPart.data is chunked into a non-streaming buffer
+          // so the ContentLength from the input should not be used for MPU.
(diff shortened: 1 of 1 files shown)


lib/lib-storage/src/Upload.ts
```diff
@@ -276,6 +276,7 @@ export class Upload extends EventEmitter {
       const partResult = await this.client.send(
         new UploadPartCommand({
           ...this.params,
+          ContentLength: partSize || undefined,
```
kuhe, line 279: any reason not to always set this to `undefined`?

the body, `dataPart.data` might always be a buffered value with known length

kuhe (approved):

PR #6132, opened 2024-05-23
fix(credentials-provider-ini): include general http provider when sourcing EcsContainer credentials

kuhe:
### Issue
investigating https://github.com/smithy-lang/smithy-typescript/pull/1288

this PR requires that https://github.com/smithy-lang/smithy-typescript/pull/1290 be published. It is using new APIs from that PR.

### Description

Fixes:
- When resolving EcsContainer as source in the ini provider, include also fromHttp next to fromContainerMetadata.
- When resolving credentials in ini-provider using an assume role "source_profile", do **not** require the source_profile to declare a `role_arn`. The `role_arn` should be declared by the origin profile.
- improve logging information for the credential chain process 

Explanation from [smithy-ts#1288](https://github.com/smithy-lang/smithy-typescript/pull/1288):

- `CredentialsProviderError: 169.254.170.23 is not a valid container metadata service hostname` this is the superficial error, which makes it look like we need to add more hosts to the hardcoded allowlist within  [packages/credential-provider-imds/src/fromContainerMetadata.ts](https://github.com/smithy-lang/smithy-typescript/pull/1288/files#diff-129b80319031d405dd4f9be72591a8fa065f6e14cad6768c9673ab832abcc449)
  - however, this is misleading, since the default credential provider chain in AWS SDKs for Node.js has many steps, and only reports the final terminal exception. There are actually many thrown & caught exceptions along the way as each credential provider within the chain is attempted.
  - as a fix, we will improve the logging of the AWS SDK's node default credential chain, https://github.com/smithy-lang/smithy-typescript/pull/1290 & downstream in [aws-sdk-js-v3 PR 6132](https://github.com/aws/aws-sdk-js-v3/pull/6132).
 
- the real root cause has two parts
  1. the `@aws-sdk/credential-provider-ini` pkg, which is part of the default chain and is responsible for file-configured assumeRole credential resolution, does not route to the newer `fromHttp` provider, only the older `fromContainerMetadata` provider. The fix PR will add this.
  2. the same package, when merging a `source_profile`, expects the `source_profile` to also have a `role_arn`. This is a separate bug, and I believe this isn't consistent with our docs at https://docs.aws.amazon.com/sdkref/latest/guide/feature-assume-role-credentials.html. It is not the source profile that should have a `role_arn`, it is the root profile.

### Testing
- [x] add to credential chain integration test

### DIFF
--- a/packages/credential-provider-ini/src/resolveAssumeRoleCredentials.ts
+++ b/packages/credential-provider-ini/src/resolveAssumeRoleCredentials.ts
@@ -170,7 +170,7 @@ export const resolveAssumeRoleCredentials = async (
     if (!options.mfaCodeProvider) {
       throw new CredentialsProviderError(
         `Profile ${profileName} requires multi-factor authentication, but no MFA code callback was provided.`,
-        { logger: options.logger }
+        { logger: options.logger, tryNextLink: false }
       );
     }
     params.SerialNumber = mfa_serial;
--- a/packages/credential-provider-ini/src/resolveAssumeRoleCredentials.ts
+++ b/packages/credential-provider-ini/src/resolveAssumeRoleCredentials.ts
@@ -1,6 +1,6 @@
 import { CredentialsProviderError } from "@smithy/property-provider";
 import { getProfileName } from "@smithy/shared-ini-file-loader";
-import { AwsCredentialIdentity, ParsedIniData, Profile } from "@smithy/types";
+import { AwsCredentialIdentity, Logger, ParsedIniData, Profile } from "@smithy/types";
 
 import { FromIniInit } from "./fromIni";
 import { resolveCredentialSource } from "./resolveCredentialSource";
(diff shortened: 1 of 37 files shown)


trivikr (approved):

packages/credential-provider-ini/src/resolveAssumeRoleCredentials.ts
```diff
@@ -105,16 +127,36 @@ export const resolveAssumeRoleCredentials = async (
...
+    ? resolveProfileData(
+        source_profile,
+        {
+          ...profiles,
+          [source_profile]: {
+            ...profiles[source_profile],
+            // This assigns the role_arn of the "root" profile
+            // to the credential_source profile so this recursive call knows
+            // what role to assume.
+            role_arn: data.role_arn ?? profiles[source_profile].role_arn,
```
jgrigg, line 150: 👏

packages/credential-provider-ini/src/resolveCredentialSource.ts
```diff
@@ -13,22 +13,34 @@ import { AwsCredentialIdentityProvider } from "@smithy/types";
...
-      import("@smithy/credential-provider-imds").then(({ fromContainerMetadata }) => fromContainerMetadata(options)),
-    Ec2InstanceMetadata: (options?: CredentialProviderOptions) =>
-      import("@smithy/credential-provider-imds").then(({ fromInstanceMetadata }) => fromInstanceMetadata(options)),
-    Environment: (options?: CredentialProviderOptions) =>
-      import("@aws-sdk/credential-provider-env").then(({ fromEnv }) => fromEnv(options)),
+    EcsContainer: async (options?: CredentialProviderOptions) => {
+      const { fromHttp } = await import("@aws-sdk/credential-provider-http");
+      const { fromContainerMetadata } = await import("@smithy/credential-provider-imds");
+      logger?.debug("@aws-sdk/credential-provider-ini - credential_source is EcsContainer");
+      return chain(fromHttp(options ?? {}), fromContainerMetadata(options));
```
jgrigg, line 24: Just an observation. Its does feel a little bit jank that `EcsContainer` is being overloaded here i.e it was previously just IMDS(container meta) but now being used as IMDS OR HTTP. It would feel cleaner to define a new credential source like `FromHTTP`...

kuhe: yes, it would be out of scope to declare an entirely new ini credential source in only this SDK. In this context fromHttp is only used for its ability to connect to EKS/ECS

PR #6227, opened 2024-06-26
fix(evenstream-handler-node): add system clock offset to event signing streams

kuhe:
- **chore: update devDependency webpack**
- **fix(eventstream-handler-node): add system clock offset to event signing streams**

### Issue

### Description

### Testing
How was this change tested?

### Additional context

### Checklist

---


packages/eventstream-handler-node/src/EventStreamPayloadHandler.ts
```diff
@@ -37,10 +38,12 @@ export interface EventStreamPayloadHandlerOptions {
 export class EventStreamPayloadHandler implements IEventStreamPayloadHandler {
   private readonly messageSigner: Provider<MessageSigner>;
   private readonly eventStreamCodec: EventStreamCodec;
+  private readonly systemClockOffsetProvider: Provider<number>;
 
   constructor(options: EventStreamPayloadHandlerOptions) {
     this.messageSigner = options.messageSigner;
     this.eventStreamCodec = new EventStreamCodec(options.utf8Encoder, options.utf8Decoder);
+    this.systemClockOffsetProvider = async () => options.systemClockOffset ?? 0;
```
kuhe, line 46: the offset is converted to a provider because the offset must be read at the time of signing and not at the time of initialization.

the offset may have been corrected at some point in the client lifecycle, necessitating a live check of the number value.

packages/eventstream-handler-node/src/EventSigningStream.ts
```diff
@@ -32,11 +34,12 @@ export class EventSigningStream extends Transform {
     this.priorSignature = options.priorSignature;
     this.eventStreamCodec = options.eventStreamCodec;
     this.messageSigner = options.messageSigner;
+    this.systemClockOffsetProvider = options.systemClockOffsetProvider;
   }
 
   async _transform(chunk: Uint8Array, encoding: string, callback: TransformCallback): Promise<void> {
     try {
-      const now = new Date();
+      const now = new Date(new Date().getTime() + (await this.systemClockOffsetProvider()));
```
kuhe, line 42: this uses `new Date().getTime()` instead of `Date.now()` due to a complex mock of the `Date` global in the unit tests.

siddsriv (approved):

trivikr (approved):

packages/eventstream-handler-node/src/EventStreamPayloadHandler.ts
kuhe: this won't work immediately because during config resolution we keep copying the object for whatever reason

PR #6232, opened 2024-06-27
chore(middleware-sdk-s3): update s3 status 200 error handling

kuhe:
### Issue
internal

Depends on https://github.com/aws/aws-sdk-js-v3/pull/6279 being merged

### Description
Updates the s3 status 200 error behavior. For JSv3 middleware, the change is to ignore the `<Error>` tag within a 200 response if it is not the top level XML element.

For Command criteria, it is no longer an allowlist, rather any operation that does not have a response payload of type blob with streaming trait.

### Testing

Added unit test.

- [x] e2e
- [x] integ


siddsriv (approved):

isaiahshall (approved):

trivikr (approved):

kuhe:
updated to require https://github.com/smithy-lang/smithy-typescript/pull/1336

uses stream splitting instead of turning the stream into a bytearray, which is incompatible for event streams

kuhe:
- [x] tested in browser app

Issue #2549, opened 2021-07-07
Setting AWS_PROFILE env var disables the env config profile loader


gambrose:
## Describe the issue with documentation
The @aws-sdk/credential-provider-node documentation says
> It will attempt to find credentials from the following sources
> * Environment variables exposed via process.env
> * etc...

It makes no mention that the @aws-sdk/credential-provider-env is only included if a profile is not set in the options (which is also picked up from AWS_PROFILE env var).

This feature to disable env loading is not very intuitive and should be more clearly documented.

## Steps to reproduce
Set env vars that would be resolved ie. `AWS_ACCESS_KEY_ID`, `AWS_SECRET_ACCESS_KEY`, etc..
Set `AWS_PROFILE`  env var to truthy value.

#### Observed behavior
The config in the env vars is not resolved.

#### Expected behaviour
The config in the env vars  ( `AWS_ACCESS_KEY_ID`, `AWS_SECRET_ACCESS_KEY`, etc..) should be resolved.

## Additional context
This is the line that optionally links in the env config resolver.
https://github.com/aws/aws-sdk-js-v3/blob/main/packages/credential-provider-node/src/index.ts#L68

pgrzesik:
I'm observing the exact same behavior, which is different than the behavior of `SDK v2`, as well as `AWS CLI` credentials resolution chain.

Is that the intended behavior? If yes, what is the reason behind it?

iocentos:
We are also hitting this behavior here. Something that was not an issue with V2.
Is this something that is intended and we should take it into account, or is it a bug and we should wait for resolution?

Zordrak:
I raised a bug on this against the wrong respository. I'm closing it there, but for the sake of complete detail - it's all there:
https://github.com/aws/aws-sdk-js/issues/4406

This definitely feels like a bug not a feature, and quite a potentially harmful one as people will (and have been) executing commands with the wrong credentials and potentially against the wrong accounts.


### Simple reproduction steps

```javascript
const AWS = require('aws-sdk');
​
const { SecurityHubClient } = require('@aws-sdk/client-securityhub');
​
const sh2 = new AWS.SecurityHub();
const sh3 = new SecurityHubClient();
​
const keyType = (creds) => {
  return creds.accessKeyId.substring(0,4);
}
​
(async () => {
  creds2 = sh2.config.credentials;
  creds3 = await sh3.config.credentials();
​
  console.log('AWS SDK v2: ' + keyType(creds2));
[...]

PR #6277, opened 2024-07-15
chore(credential-provider-node): emit warning when AWS_PROFILE is set alongside ENV credentials

kuhe:
### Issue
https://github.com/aws/aws-sdk-js-v3/issues/2549

### Description
Emit a warning when the environment sets AWS_PROFILE in addition to the pair AWS_ACCESS_KEY_ID/AWS_SECRET_ACCESS_KEY.

We should be resolving the credential pair based on AWS SDK JSv2, AWS CLI, and documentation, but the current implementation prefers AWS_PROFILE.
Because this has been released for a very long time, we should set a warning instead of directly switching over to preferring the credential pair. 

We could determine a specific date later on which to change the behavior over to preferring the credential pair.

### Testing
manual, integ, unit

### DIFF
--- a/packages/credential-provider-node/src/defaultProvider.ts
+++ b/packages/credential-provider-node/src/defaultProvider.ts
@@ -1,4 +1,4 @@
-import { fromEnv } from "@aws-sdk/credential-provider-env";
+import { ENV_KEY, ENV_SECRET, fromEnv } from "@aws-sdk/credential-provider-env";
 import type { FromHttpOptions } from "@aws-sdk/credential-provider-http";
 import type { FromIniInit } from "@aws-sdk/credential-provider-ini";
 import type { FromProcessInit } from "@aws-sdk/credential-provider-process";
@@ -20,6 +20,11 @@ export type DefaultProviderInit = FromIniInit &
   FromProcessInit &
   (FromSSOInit & Partial<SsoCredentialsParameters>) &
   FromTokenFileInit;
+
+/**
+ * @internal
+ */
+let multipleCredentialSourceWarningEmitted = false;
 
 /**
  * Creates a credential provider that will attempt to find credentials from the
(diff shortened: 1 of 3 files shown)


trivikr (approved):

packages/credential-provider-node/src/defaultProvider.ts
```diff
@@ -58,14 +63,36 @@ export type DefaultProviderInit = FromIniInit &
...
-          ]),
+      async () => {
+        const profile = init.profile ?? process.env[ENV_PROFILE];
+        if (profile) {
+          const envStaticCredentialsAreSet = process.env[ENV_KEY] && process.env[ENV_SECRET];
+          if (envStaticCredentialsAreSet) {
+            if (!multipleCredentialSourceWarningEmitted) {
+              const warnFn =
+                init.logger?.warn && init.logger?.constructor?.name !== "NoOpLogger" ? init.logger.warn : console.warn;
+              warnFn(
```
kuhe, line 74: This warning is emitted here in the default provider chain rather than in the `fromEnv` provider function.

The reason for this is because when using the `fromEnv` provider in isolation, it doesn't matter if the `AWS_PROFILE` is set or not.

trivikr:
Example warning

```console
@aws-sdk/credential-provider-node - defaultProvider::fromEnv WARNING:
    Multiple credential sources detected: 
    Both AWS_PROFILE and the pair AWS_ACCESS_KEY_ID/AWS_SECRET_ACCESS_KEY static credentials are set.
    This SDK will proceed with the AWS_PROFILE value.
    
    However, a future version may change this behavior to prefer the ENV static credentials.
    Please ensure that your environment only sets either the AWS_PROFILE or the
    AWS_ACCESS_KEY_ID/AWS_SECRET_ACCESS_KEY pair.
```

siddsriv (approved):

P0rth0s:
@kuhe I believe this has introduced a regression.

`
TypeError: _d.debug is not a function
    at _a (/usr/src/app/node_modules/@aws-sdk/credential-provider-node/dist-cjs/index.js:96:48)
    at /usr/src/app/node_modules/@smithy/property-provider/dist-cjs/index.js:97:33
    at coalesceProvider (/usr/src/app/node_modules/@smithy/property-provider/dist-cjs/index.js:121:17)
    at /usr/src/app/node_modules/@smithy/property-provider/dist-cjs/index.js:142:24
    at /usr/src/app/node_modules/@smithy/core/dist-cjs/index.js:82:23
    at processTicksAndRejections (node:internal/process/task_queues:95:5)
    at /usr/src/app/node_modules/@aws-sdk/middleware-logger/dist-cjs/index.js:34:22
`

I am currently working on a script to reproduce for you

Issue #6276, opened 2024-07-14
marshallInput and unmarshallOutput are missing from the CJS distribution of lib-dynamodb


sawilde:
### Checkboxes for prior research

- [X] I've gone through [Developer Guide](https://docs.aws.amazon.com/sdk-for-javascript/v3/developer-guide) and [API reference](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest)
- [X] I've checked [AWS Forums](https://forums.aws.amazon.com) and [StackOverflow](https://stackoverflow.com/questions/tagged/aws-sdk-js).
- [X] I've searched for [previous similar issues](https://github.com/aws/aws-sdk-js-v3/issues) and didn't find any solution.

### Describe the bug

`marshallInput` and `unmarshallOutput` are missing from the `CJS` distribution of `lib-dynamodb`

These functions are described in `types` and are exported in the `ES` distribution but are not exported in the `CJS` distribution.

Walking the packages this seems to have happened starting with build `3.495.0` when the generated utils.js started to only export from its local `index.js` file but the above functions were not included in that file.

I number of Open Source packages depend on this function e.g. [electrodb](https://github.com/tywalch/electrodb)

### SDK version number

@aws-sdk/lib-dynamodb@3.395.0, ...

### Which JavaScript Runtime is this issue in?

Node.js

### Details of the browser/Node.js/ReactNative version

v18.18.2 and others

### Reproduction Steps

1. install @aws-sdk/lib-dynamodb@3.394.0 (see exports in `dist-cjs/commands/utils.js` compare with `dist-es\...`)
2. install @aws-sdk/lib-dynamodb@3.395.0 (see exports in `dist-cjs/commands/utils.js` compare with `dist-es\...`)

### Observed Behavior

Cannot import `marshallInput` or `unmarshallOutput`

### Expected Behavior

Should be able to import `marshallInput` or `unmarshallOutput`

### Possible Solution

Export the two functions in the CJS distribution

kuhe:
This is a private API of the AWS SDK.

Based on the usage observed in electrodb, it can be replaced like this:

```ts
// unmarshallOutput is internal/private

if (reason.Item) {
  return unmarshallOutput(reason, [{ key: "Item" }]);
}
```

to:

```ts
import { unmarshall } from "@aws-sdk/util-dynamodb";

if (reason.Item) {
  return unmarshall(reason.Item);
}
```

sawilde:
[...]

PR #6281, opened 2024-07-15
feat(lib-dynamodb): export marshallInput and unmarshallOutput 

RanVaknin:
### Issue
https://github.com/aws/aws-sdk-js-v3/issues/6276

### Description
Exporting lib-dynamo's `marshallInput` and `unmarshallOutput` in the package index for CJS distribution. They were already exported for ES distribution.

### DIFF
--- a/lib/lib-dynamodb/src/commands/index.ts
+++ b/lib/lib-dynamodb/src/commands/index.ts
@@ -12,3 +12,4 @@ export * from "./ScanCommand";
 export * from "./TransactGetCommand";
 export * from "./TransactWriteCommand";
 export * from "./UpdateCommand";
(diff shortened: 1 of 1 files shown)


lib/lib-dynamodb/src/commands/index.ts
```diff
@@ -12,3 +12,4 @@ export * from "./ScanCommand";
 export * from "./TransactGetCommand";
 export * from "./TransactWriteCommand";
 export * from "./UpdateCommand";
+export { marshallInput, unmarshallOutput } from "./utils";
```
kuhe, line 15:
```suggestion
/**
 * @internal
 */
export { marshallInput, unmarshallOutput } from "./utils";
```

PR #6322, opened 2024-07-29
Explicit http credential source

jgrigg:
### Description

The `EcsContainer` provider implies IMDS. Having the `fromHttp` provider tacked on the front of the chain is a bit misleading. This adds a new `credential_source` called `Http` which does what it says on the tin.

### Testing

Unit

### Additional context

This is obviously a breaking change but i suspect not many businesses are using it anyway. I also understand that this diverges from the spec and all the other SDKs would need to follow suit. Mainly raising this for discussion purposes.

### Checklist

---

### DIFF
--- a/packages/credential-provider-ini/src/resolveCredentialSource.ts
+++ b/packages/credential-provider-ini/src/resolveCredentialSource.ts
@@ -32,7 +32,7 @@ export const resolveCredentialSource = (
       const { fromEnv } = await import("@aws-sdk/credential-provider-env");
       return fromEnv(options);
     },
-    HTTP: async (options?: CredentialProviderOptions) => {
+    Http: async (options?: CredentialProviderOptions) => {
       logger?.debug("@aws-sdk/credential-provider-ini - credential_source is Environment");
       const { fromHttp } = await import("@aws-sdk/credential-provider-http");
       return fromHttp(options ?? {});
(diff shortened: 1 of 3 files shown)


packages/credential-provider-ini/src/resolveCredentialSource.ts
```diff
@@ -18,10 +18,9 @@ export const resolveCredentialSource = (
 ): ((options?: CredentialProviderOptions) => Promise<AwsCredentialIdentityProvider>) => {
   const sourceProvidersMap = {
     EcsContainer: async (options?: CredentialProviderOptions) => {
-      const { fromHttp } = await import("@aws-sdk/credential-provider-http");
       const { fromContainerMetadata } = await import("@smithy/credential-provider-imds");
       logger?.debug("@aws-sdk/credential-provider-ini - credential_source is EcsContainer");
-      return chain(fromHttp(options ?? {}), fromContainerMetadata(options));
+      return fromContainerMetadata(options);
```
kuhe, line 23: we cannot accept this change, since the `fromHttp` provider implements unique functionality expected by users with a EcsContainer credential_source.

jgrigg: Yeah hence my comment around the breaking change (which could be made non-breaking). It was me that asked you for that functionality about two months back (thank you!) so i suspect I'm the only one using it.

Issue #6272, opened 2024-07-11
Allow customizable credential provider chain


mxxk:
### Describe the feature

Provide a construct to allow users of the AWS SDK for JavaScript v3 to express a chain of credential providers. An example of this is the [`@aws-sdk/credential-provider-node`](https://github.com/aws/aws-sdk-js-v3/tree/v3.614.0/packages/credential-provider-node) provider, but its chain of credentials is not customizable. This was possible in v2 using [`CredentialProviderChain`](https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/CredentialProviderChain.html), but there is no equivalent v3.

### Use Case

Here are some example use cases which are not currently supported due to lack of custom composition of credential providers:

- **Credential provider exclusion.** There is no way to remove specific credential providers from the chain for security purposes. For example, if credentials should only be sourced from ECS container metadata or a file, but not from environment variables, there is no way to chain only those two credential providers.
- **Custom credential providers.** If there is a custom credential providers, there is no way to compose it into a chain of other credential providers.

### Proposed Solution

Since `@aws-sdk/credential-provider-node` relies on `chain` to compose multiple credential providers,

https://github.com/aws/aws-sdk-js-v3/blob/794a37e9795f390d15c9530209e465d099716c86/packages/credential-provider-node/src/defaultProvider.ts#L58-L60

one possible way forward might be to expose [`chain`](https://github.com/smithy-lang/smithy-typescript/blob/%40smithy/property-provider%403.1.3/packages/property-provider/src/chain.ts) (and possibly [`memoize`](https://github.com/smithy-lang/smithy-typescript/blob/%40smithy/property-provider%403.1.3/packages/property-provider/src/memoize.ts)) for public consumption. Both currently belong to `@smithy/property-provider`, an internal package.

### Acknowledgements

- [X] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### SDK version used

3.614.0

### Environment details (OS name and version, etc.)

macOS 14.5

kuhe:
You can use promises to create a credential chain.

```ts
import { fromEnv, fromSSO, fromIni, fromHttp } from "@aws-sdk/credential-providers";

const myCredentialProvider = async () => ({ secretAccessKey, ... })

new AWSSDKClient({
  credentials: () => 
    fromEnv()()
      .catch(fromSSO())
      .catch(fromHttp())
      .catch(fromIni())
      .catch(myCredentialProvider)
});
[...]

PR #6332, opened 2024-07-30
Warning on retry

jgrigg:
### Description

The http provider fails silently after 3 retries and is very difficult to debug.

### Testing

N/A

### Additional context

In our case it turned out the pod identity agent wasn't running on the node and our istio gateway was throwing a 503. this was caught and silently failed then the default provider continued on it's merry way and failed loudly at another point.

### Checklist

---

### DIFF
--- a/packages/credential-provider-http/src/fromHttp/retry-wrapper.ts
+++ b/packages/credential-provider-http/src/fromHttp/retry-wrapper.ts
@@ -1,3 +1,5 @@
+import { Logger } from "@smithy/types";
+
 /**
  * @internal
  */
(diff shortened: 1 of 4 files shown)


packages/credential-provider-http/src/fromHttp/retry-wrapper.ts
```diff
@@ -10,14 +12,19 @@ export interface RetryableProvider<T> {
...
+  const { maxRetries, delayMs, logger } = options;
   return async () => {
     for (let i = 0; i < maxRetries; ++i) {
       try {
         return await toRetry();
       } catch (e) {
+        if (i === maxRetries - 1) {
+          logger?.warn?.(
+            `@aws-sdk/credential-provider-http - retry limit ${maxRetries} reached. error: ${JSON.stringify(e)}`
+          );
```
kuhe, line 26: This is not the right place to emit a warning because there is one more attempt made on line 31. For maxRetries=2, this would be attempt 2 of 3 total.

Also, this wouldn't be a scenario for which to emit a warning. The default credential chain can fail at many steps and be intentional. 

To ensure that a specific credential source throws a highly visible Error, create a chain or single-step function in which it is the final credential source.

```js
// example: use critical credential source as final step in a credential chain.
import { fromEnv, fromHttp } from "@aws-sdk/credential-providers";

new SDKClient({
  credentials: async () => fromEnv({})().catch(fromHttp({}))
});
```

If you want to continue using the default provider chain, use a log filter to surface the `CredentialProviderError`s:

```js
// example: log CredentialProviderError indicating progression in the 
//                 default credential chain from one source to the next.
new SDKClient({
  logger: {
     ...console,
    debug(...args) {
      // surface debug messages within the credential chain coming from CredentialProviderError
      if (args[0].startsWith("@smithy/property-provider")) {
        console.warn(...args);
      }
    },
  }
});
```

kuhe:
Going to reopen this feature request for easier creation of custom credential chains: https://github.com/aws/aws-sdk-js-v3/issues/6272.

PR #6349, opened 2024-08-02
fix(client-transcribe-streaming): add plugin only for operations with streaming trait

trivikr:
### Issue

Internal JS-5337

### Description

Adds plugin only for operations with streaming trait

### Testing
Internal

---

### DIFF
--- a/codegen/smithy-aws-typescript-codegen/src/main/java/software/amazon/smithy/aws/typescript/codegen/AddTranscribeStreamingDependency.java
+++ b/codegen/smithy-aws-typescript-codegen/src/main/java/software/amazon/smithy/aws/typescript/codegen/AddTranscribeStreamingDependency.java
@@ -22,6 +22,8 @@ import java.util.function.Consumer;
 import software.amazon.smithy.aws.traits.ServiceTrait;
 import software.amazon.smithy.codegen.core.SymbolProvider;
 import software.amazon.smithy.model.Model;
+import software.amazon.smithy.model.knowledge.EventStreamIndex;
+import software.amazon.smithy.model.shapes.OperationShape;
 import software.amazon.smithy.model.shapes.ServiceShape;
 import software.amazon.smithy.typescript.codegen.LanguageTarget;
 import software.amazon.smithy.typescript.codegen.TypeScriptSettings;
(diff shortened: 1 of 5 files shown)


trivikr:
This quick fix was provided to fix existing error with future commands which do not contain eventstream

```console
Error: eventStreamHeaderMiddleware is not found when adding websocketPortMiddleware middleware after eventStreamHeaderMiddleware
```

A better fix should should move the plugin relative to some other existing middleware present in all clients, so that `websocketPortMiddleware` is applicable for all commands when WebSocketHandler is used in browser runtime.

codegen/smithy-aws-typescript-codegen/src/main/java/software/amazon/smithy/aws/typescript/codegen/AddTranscribeStreamingDependency.java
```diff
@@ -92,6 +94,11 @@ private static boolean isTranscribeStreaming(ServiceShape service) {
         String serviceId = service.getTrait(ServiceTrait.class).map(ServiceTrait::getSdkId).orElse("");
         return serviceId.equals("Transcribe Streaming");
     }
+
+    private static boolean hasEventStreamInput(Model model, ServiceShape service, OperationShape operation) {
+        EventStreamIndex eventStreamIndex = EventStreamIndex.of(model);
+        return eventStreamIndex.getInputInfo(operation).isPresent();
+    }
```
kuhe, line 101: this is ok as a patch, but long term the conditional method should be exported for use by the websocket plugin, since it requires this plugin to be present.

kuhe (approved):

trivikr:
> A better fix should move the plugin relative to some other existing middleware present in all clients

An alternative fix was provided in https://github.com/aws/aws-sdk-js-v3/pull/6350

codegen/smithy-aws-typescript-codegen/src/main/java/software/amazon/smithy/aws/typescript/codegen/AddTranscribeStreamingDependency.java
trivikr: PR posted at https://github.com/aws/aws-sdk-js-v3/pull/6351

PR #6350, opened 2024-08-02
fix(middleware-sdk-transcribe-streaming): move middleware before signing

trivikr:
### Issue
Internal JS-5337

### Description
Move websocketPortMiddleware before httpSigningMiddleware

### Testing
ToDo

---


trivikr:
This alternative was rejected as:
* We are only using websockets on operations with event streams: https://github.com/aws/aws-sdk-js-v3/blob/9836a09e97bc7f260f75eb1c5ff831f671ce2ec8/codegen/smithy-aws-typescript-codegen/src/main/java/software/amazon/smithy/aws/typescript/codegen/AddWebsocketPlugin.java#L62-L65
* Changing the ordering is not required, as it might cause some other effects.

PR #6361, opened 2024-08-06
fix(turbo): improve error handling in build script & remove unnecessary filtering from build:ci

isaiahshall:
### Issue
Previously, Turborepo build errors due to package filtering or other optional arguments would result in a CI build failure. This code change adds a fallback for these errors, rebuilding the package without any optional arguments. It also removes unnecessary filtering from the build:ci script in package.json. 

---

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
     "bootstrap": "yarn",
     "bootstrap:ci": "yarn install --frozen-lockfile",
     "build:all": "yarn build:crypto-dependencies && node ./scripts/turbo build",
-    "build:ci": "node ./scripts/turbo build -F=[origin/main...HEAD]...",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -8,7 +8,7 @@
     "bootstrap": "yarn",
     "bootstrap:ci": "yarn install --frozen-lockfile",
     "build:all": "yarn build:crypto-dependencies && node ./scripts/turbo build",
-    "build:ci": "node ./scripts/turbo build -F=[origin/main...HEAD]...",
+    "build:ci": "node ./scripts/turbo build",
```
kuhe, line 11: is build crypto dependencies even necessary?

kuhe (approved):

PR #6379, opened 2024-08-12
test(endpoint): enable test for s3 empty arn type

trivikr:
### Issue
Internal JS-5234

### Description
Enables test for s3 empty arn type

### Testing
ToDo

---
By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #6380, opened 2024-08-12
fix(util-endpoints): check for entire resource-path being empty

trivikr:
### Issue
Refs: https://github.com/aws/aws-sdk-js-v3/pull/6379#issuecomment-2284943667

### Description
Check for entire resource-path being empty when parsing Arn

### Testing

Unit and Endpoint tests are successful
```console
$ yarn jest -c tests/endpoints-2.0/jest.config.js tests/endpoints-2.0/endpoints-integration.spec.ts
...
Test Suites: 1 passed, 1 total
Tests:       14672 passed, 14672 total
```

---

### DIFF
--- a/packages/util-endpoints/src/lib/aws/parseArn.ts
+++ b/packages/util-endpoints/src/lib/aws/parseArn.ts
@@ -1,24 +1,31 @@
 import { EndpointARN } from "@smithy/types";
 
+const ARN_DELIMITER = ":";
+const RESOURCE_DELIMITER = "/";
+
 /**
  * Evaluates a single string argument value, and returns an object containing
  * details about the parsed ARN.
  * If the input was not a valid ARN, the function returns null.
  */
 export const parseArn = (value: string): EndpointARN | null => {
-  const segments = value.split(":");
+  const segments = value.split(ARN_DELIMITER);
 
   if (segments.length < 6) return null;
 
-  const [arn, partition, service, region, accountId, ...resourceId] = segments;
+  const [arn, partition, service, region, accountId, ...resourcePath] = segments;
 
-  if (arn !== "arn" || partition === "" || service === "" || resourceId[0] === "") return null;
+  if (arn !== "arn" || partition === "" || service === "" || resourcePath.join(ARN_DELIMITER) === "") return null;
+
+  const resourceId = resourcePath[0].includes(RESOURCE_DELIMITER)
+    ? resourcePath[0].split(RESOURCE_DELIMITER)
+    : resourcePath;
 
   return {
     partition,
     service,
     region,
     accountId,
-    resourceId: resourceId[0].includes("/") ? resourceId[0].split("/") : resourceId,
+    resourceId,
   };
 };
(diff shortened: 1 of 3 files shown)


kuhe (approved):

packages/util-endpoints/src/lib/aws/parseArn.ts
```diff
@@ -1,24 +1,31 @@
...
 
-  const [arn, partition, service, region, accountId, ...resourceId] = segments;
+  const [arn, partition, service, region, accountId, ...resourcePath] = segments;
+
+  if (arn !== "arn" || partition === "" || service === "" || resourcePath.join(ARN_DELIMITER) === "") return null;
 
-  if (arn !== "arn" || partition === "" || service === "" || resourceId[0] === "") return null;
+  const resourceId = resourcePath[0].includes(RESOURCE_DELIMITER)
+    ? resourcePath[0].split(RESOURCE_DELIMITER)
+    : resourcePath;
```
gosar, line 22: This seems to assume that if `resourcePath[0]` contains `/`, then `resourcePath` is only of size 1. Is that a safe assumption? I don't know the scope of where this is going to get used, but if you see an arn like `arn:aws:s3:us-west-2:123456789012:access/point:my/endpoint` seems like you'd parse it as `resourceId: ["access", "point"]`. This doesn't seem expected behavior because it is losing data. I don't know if my made up example is not possible, but I'm assuming the expected parsing in this case might be `resourceId: ["access", "point:my", "endpoint"]`.

gosar: I think S3 allows `:` in file names, so the made up example seems possible.

trivikr: > This seems to assume that if resourcePath[0] contains /, then resourcePath is only of size 1. Is that a safe assumption?

Yes. The `resource` either contains `:` or `/` delimiter.
Official documentation on S3 ARN format https://docs.aws.amazon.com/IAM/latest/UserGuide/reference-arns.html

trivikr: Fixed in [`855f1fa` (#6380)](https://github.com/aws/aws-sdk-js-v3/pull/6380/commits/855f1fa613997ff2c2c0a08412e58c041911e834)

kuhe (approved):

packages/util-endpoints/src/lib/aws/parseArn.ts
trivikr: Reverted in [`3937a6a` (#6380)](https://github.com/aws/aws-sdk-js-v3/pull/6380/commits/3937a6a4783c320a08aa74ba7b1ea41cc8a2275e) as more discussion is needed. And it's not blocking the original fix.

gosar (approved):
Ok to ship this to unblock failing test, but follow up on expected behavior for case like `arn:aws:s3:us-west-2:123456789012:my:folder/my:file`

Issue #5958, opened 2024-04-04
AWS Transcribe Streaming HTTP2 error message with WebToken


MartinEmrich:
### Checkboxes for prior research

- [X] I've gone through [Developer Guide](https://docs.aws.amazon.com/sdk-for-javascript/v3/developer-guide) and [API reference](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest)
- [X] I've checked [AWS Forums](https://forums.aws.amazon.com) and [StackOverflow](https://stackoverflow.com/questions/tagged/aws-sdk-js).
- [X] I've searched for [previous similar issues](https://github.com/aws/aws-sdk-js-v3/issues) and didn't find any solution.

### Describe the bug

AWS Transcribe Streaming fails with errors like

```
{\"code\":\"ERR_HTTP2_ERROR\",\"errno\":-505,\"$metadata\":{\"attempts\":1,\"totalRetryDelay\":0}}
```

```
Error [ERR_HTTP2_ERROR]: Protocol error\n    at Http2Session.onSessionInternalError (node:internal/http2/core:801:26)
```

### SDK version number

@aws-sdk/client-transcribe-streaming 3.540.0

### Which JavaScript Runtime is this issue in?

Node.js

### Details of the browser/Node.js/ReactNative version

v20.5.1

### Reproduction Steps

Use Transcribe Streaming from a service within an AWS EKS Kubernetes cluster with an IAM service account.

````
const client = new transcribe.TranscribeStreamingClient();
// ...
const response = await client.send(new transcribe.StartStreamTranscriptionCommand(transcribeParams));
// ...
````

### Observed Behavior

Running the service (in fact the same container image) locally (AWS SSO auth) or on an EC2 instance (IAM instance role) works as expected.

Running it on AWS EKS with an IAM service account (Token file authentication), the STSClient call made by the Transcribe client fails:

````
[...]

PR #6389, opened 2024-08-15
fix(credential-providers): avoid sharing http2 requestHandler with inner STS

kuhe:
### Issue
https://github.com/aws/aws-sdk-js-v3/issues/5958

### Description
Avoid sharing the parent client requestHandler to the inner STS client if that handler is an HTTP2 handler.
All other types of known handlers: fetch, http1.1, websocket (because it defers to fetch for non-ws), xhr, are valid for STS.

### Testing
Made a TranscribeStreaming (HTTP2 default requesthandler) request and used the default provider chain such that it involved an STS credentials request. 

Verified this PR makes the difference in that not working -> working.

Sample code:

```js
import { TranscribeStreaming } from "@aws-sdk/client-transcribe-streaming";
import { fromIni } from "@aws-sdk/credential-providers";
import { createReadStream } from "node:fs";
import { join } from "node:path";

const audio = createReadStream(join("numbers.wav"));

process.env.AWS_PROFILE = "sts";

/**
 * When no credentials or requestHandler are specified, the http2
 * request handler default on the client is shared with the
 * inner STS client via config copying, but is invalid
 * for STS, which needs a regular http1 requestHandler.
 */

const ts = new TranscribeStreaming({
  region: "us-west-2",
  logger: console,
});

const LanguageCode = "en-GB";
const MediaEncoding = "pcm";
const MediaSampleRateHertz = 44100;
const result = await ts.startStreamTranscription({
  LanguageCode,
  MediaEncoding,
  MediaSampleRateHertz,
  AudioStream: (async function* () {
    for await (const chunk of audio) {
      yield { AudioEvent: { AudioChunk: chunk } };
    }
  })(),
});
const transcripts = [];
for await (const event of result.TranscriptResultStream) {
  transcripts.push(event);
}
console.log(transcripts);

```

### DIFF
--- a/clients/client-sts/src/defaultStsRoleAssumers.ts
+++ b/clients/client-sts/src/defaultStsRoleAssumers.ts
@@ -101,11 +101,13 @@ export const getDefaultRoleAssumer = (
         stsOptions?.parentClientConfig?.region,
         credentialProviderLogger
       );
+      const isCompatibleRequestHandler = !isH2(requestHandler);
+
       stsClient = new stsClientCtor({
         // A hack to make sts client uses the credential in current closure.
         credentialDefaultProvider: () => async () => closureSourceCreds,
         region: resolvedRegion,
-        requestHandler: requestHandler as any,
+        requestHandler: isCompatibleRequestHandler ? (requestHandler as any) : undefined,
         logger: logger as any,
       });
     }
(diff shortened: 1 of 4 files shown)


trivikr (approved):

kuhe:
```
npx cucumber-js -t @sts
...........
```

PR #6431, opened 2024-09-03
fix(codegen): use AWS SDK default credentials as sigv4 default credentials

kuhe:
### Issue
internal

### Description
For the scenario of a non-AWS service with the sigv4 trait, this PR modifies the ID&Auth SRA codegen behavior to provide the AWS SDK default credential chain as the default credentials object, for consistency with the legacyAuth behavior.

This was not present in the SRA implementation, presumably because of the desired separation between AWS and white label clients.

This only applies to codegen that uses both smithy-typescript and smithy-typescript-aws. If you only use smithy-typescript, it isn't currently aware of sigv4 at all. Sigv4 handling may need to be moved into smithy-typescript, but is out of scope of this PR.

### Testing
new integ test

### DIFF
--- a/codegen/smithy-aws-typescript-codegen/src/main/java/software/amazon/smithy/aws/typescript/codegen/auth/http/integration/AwsSdkCustomizeSigV4Auth.java
+++ b/codegen/smithy-aws-typescript-codegen/src/main/java/software/amazon/smithy/aws/typescript/codegen/auth/http/integration/AwsSdkCustomizeSigV4Auth.java
@@ -102,6 +102,11 @@ public class AwsSdkCustomizeSigV4Auth implements HttpAuthTypeScriptIntegration {
                         "credentialDefaultProvider", w ->
                             w.write("((_: unknown) => () => Promise.reject(new Error(\"Credential is missing\")))")
                     );
+                } else {
+                    return MapUtils.of(
+                        "credentials", w ->
+                            w.write("(() => () => Promise.reject(new Error(\"Credentials are missing\")))")
+                    );
                 }
             case NODE:
                 if (isAwsService(service)) {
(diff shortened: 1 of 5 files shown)


codegen/smithy-aws-typescript-codegen/src/main/java/software/amazon/smithy/aws/typescript/codegen/auth/http/integration/AwsSdkCustomizeSigV4Auth.java
```diff
@@ -133,6 +138,18 @@ public Map<String, Consumer<TypeScriptWriter>> getRuntimeConfigWriters(
                         );
                     }
                     return map;
+                } else {
+                    // isSigV4Service and !isAwsService are implied here.
+                    return MapUtils.of(
+                        "credentials", writer -> {
```
gosar, line 144: I'm not up to speed, but what's the difference between configuring `credentials` here v/s `credentialDefaultProvider`? It seems more consistent to configure `credentialDefaultProvider` here. And that's what it was [before](https://github.com/aws/aws-sdk-js-v3/blob/06f51ccdc1c825f4b4de6a865520ace78c098ff3/private/weather/src/runtimeConfig.ts#L34).

kuhe: `credentialDefaultProvider` was deprecated by ID&Auth SRA. It is a field that is only generated when the AwsSdkCustomizeSigV4Auth plugin is activated by the `isAwsService` and `isSigV4Service` conditionals in conjunction, and relies on data from the (AWS) service trait.

That plugin contains the config resolver that processes the credentialDefaultProvider field along with a few other AWS customizations. Rather than bringing all of that in, I'm using a default value for the credentials field, which is functionally very similar.

kuhe: `credentialDefaultProvider` seems to be a quirk from legacy dev. Even in legacyAuth, it was simply assigned onto the `credentials` field if the credentials field was empty during config resolution.

trivikr (approved):

gosar (approved):

PR #6448, opened 2024-09-09
chore: Modified bug issue template to add checkbox to report potential regression.

ashishdhingra:
### Issue

### Description
- Modified bug issue template to add checkbox to report potential regression.
- Added GitHub action to add/remove label `potential-regression` when issue is created/edited.

**NOTE:**
- The GitHub action code was ported from CDK repo.
- The label `potential-regression` would need to be created manually (we could use color `#FF6700` and description `Marking this issue as a potential regression to be checked by team member` to make it consistent across SDK repos)

Label `potential-regression` would make issue standout in the list and help engineers to handle high severity issues effectively.

### Testing
How was this change tested?

### Additional context

### Checklist

---

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug-report.yml
+++ b/.github/ISSUE_TEMPLATE/bug-report.yml
@@ -26,6 +26,14 @@ body:
       placeholder: A clear and concise description of the bug.
     validations:
       required: true
+  - type: checkboxes
+    id: regression
+    attributes:
+      label: Regression Issue
+      description: What is a regression? If it worked in a previous version but doesn't in the latest version, it's considered a regression. In this case, please provide specific version number in the report.
+      options:
+        - label: Select this option if this issue appears to be a regression.
+          required: false
   - type: input
     id: sdk-version
     attributes:
--- a/.github/workflows/issue-regression-labeler.yml
+++ b/.github/workflows/issue-regression-labeler.yml
@@ -1,0 +1,32 @@
+# Apply potential regression label on issues
+name: issue-regression-label
+on:
+  issues:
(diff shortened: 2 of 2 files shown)


.github/ISSUE_TEMPLATE/bug-report.yml
```diff
@@ -26,6 +26,14 @@ body:
       placeholder: A clear and concise description of the bug.
     validations:
       required: true
+  - type: checkboxes
+    id: regression
+    attributes:
+      label: Regression Issue
+      description: What is a regression? If it worked in a previous version but doesn't in the latest version, it's considered a regression. In this case, please provide specific version number in the report.
```
kuhe, line 33: this should probably say that differences from AWS SDK for JavaScript v2 are not regressions

ashishdhingra: > this should probably say that differences from AWS SDK for JavaScript v2 are not regressions

@kuhe Thanks for the feedback. However, the purpose of this checkbox is to identify potential regression in the current major version of SDK, not differences between previous and current major version. The change is implemented across all SDK(s), JS v3 one if them.

**Context:** There was a regression issue in CDK and team implemented this checkbox along with GitHub action that adds `potential-regression` label. It was helpful in triaging such issues quickly and has assisted in identifying some regression issues till date. There is another PR for JS v2. Even though that's getting deprecated, any future security issue fixes or service updates could also cause a regression. Hence I had reopened that PR; @aBurmeseDev will discuss it with JS team.

kuhe (approved):

PR #6455, opened 2024-09-10
fix(codegen): checkstyle issue in DocumentBareBonesClientGenerator

trivikr:
### Issue
This was missed in https://github.com/aws/aws-sdk-js-v3/pull/6449 as the codegen-ci was not run, likely to some issue with one of the paths configs https://github.com/aws/aws-sdk-js-v3/pull/6439

### Description
Fix the checkstyle issue to unblock https://github.com/aws/aws-sdk-js-v3/pull/6453
The investigation for why codegen-ci was not run will be done as a follow-up.

### Testing
CI

---
By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #6456, opened 2024-09-10
fix(codegen): checkstyle issue in DocumentBareBonesClientGenerator

trivikr:
Alternative to https://github.com/aws/aws-sdk-js-v3/pull/6455 to check if `codegen-ci` is run.


trivikr:
These findings are not reliable, since the commit ID is same as in https://github.com/aws/aws-sdk-js-v3/pull/6455 as GitHub just used the cache.

PR #6457, opened 2024-09-10
fix(codegen): different commit title to create new commit

trivikr:
Alternative to https://github.com/aws/aws-sdk-js-v3/pull/6455 to check if `codegen-ci` is run.


trivikr:
Verified that `codegen-ci` was run when PR was created as part of commit collection and the latest commit did not contain change in `codegen` folder.

PR #6465, opened 2024-09-12
chore(middleware-flexible-checksums): add RequestChecksumCalculation and ResponseChecksumValidation

trivikr:
### Issue
Internal JS-5396

### Description
Adds constants RequestChecksumCalculation and ResponseChecksumValidation.
They'll be set in clients in the following PR.

### Testing
N/A

---

### DIFF
--- a/packages/middleware-flexible-checksums/src/stringUnionSelector.ts
+++ b/packages/middleware-flexible-checksums/src/stringUnionSelector.ts
@@ -1,0 +1,27 @@
+export enum SelectorType {
+  ENV = "env",
+  CONFIG = "shared config entry",
+}
+
+/**
+ * Returns undefined, if obj[key] is not defined.
+ * Returns string value, if the string is defined in obj[key] and it's uppercase matches union value.
+ * Throws error for all other cases.
+ *
+ * @internal
+ */
+export const stringUnionSelector = (
+  obj: Record<string, string | undefined>,
+  key: string,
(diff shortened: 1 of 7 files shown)


packages/middleware-flexible-checksums/src/stringUnionSelector.ts
```diff
@@ -0,0 +1,27 @@
...
+) => {
+  if (!(key in obj)) return undefined;
+
+  const value = obj[key]!.toUpperCase();
+  if (!Object.values(union).includes(value)) {
+    throw new TypeError(`Cannot load ${type} '${key}'. Expected one of ${Object.values(union)}, got '${obj[key]}'.`);
+  }
+
+  return value;
+};
```
kuhe, line 27: it looks like this could/should be reused for other enum specifications like AccountIdEndpointMode

kuhe (approved):

packages/middleware-flexible-checksums/src/stringUnionSelector.ts
trivikr: Their validator does not check for casing, but flexible checksums one needs to.

https://github.com/aws/aws-sdk-js-v3/blob/6a539fee9650cba87b20dafc788488deb3f16878/packages/core/src/submodules/account-id-endpoint/AccountIdEndpointModeConstants.ts#L10-L12

PR #6470, opened 2024-09-13
chore(middleware-flexible-checksums): add config resolver

trivikr:
### Issue
Internal JS-5396

### Description
Add resolver for flexible checksums.
This PR also moves default values for configurations to constants.

### Testing
CI

---

### DIFF
--- a/packages/middleware-flexible-checksums/src/resolveFlexibleChecksumsConfig.ts
+++ b/packages/middleware-flexible-checksums/src/resolveFlexibleChecksumsConfig.ts
@@ -1,0 +1,33 @@
+import { Provider } from "@smithy/types";
+import { normalizeProvider } from "@smithy/util-middleware";
+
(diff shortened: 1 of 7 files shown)


packages/middleware-flexible-checksums/src/resolveFlexibleChecksumsConfig.ts
```diff
@@ -0,0 +1,33 @@
...
+
+export interface FlexibleChecksumsResolvedConfig {
+  requestChecksumCalculation: Provider<string>;
+  responseChecksumValidation: Provider<string>;
+}
+
+export const resolveFlexibleChecksumsConfig = <T>(
+  input: T & FlexibleChecksumsInputConfig
+): T & FlexibleChecksumsResolvedConfig => ({
+  ...input,
```
kuhe, line 26: Not something to address for this PR, but overall I think we should mutate directly instead of shallow-copy-and-write Object.assign.  

This is a config resolver and not in a hot path so it's fine here.

kuhe (approved):

PR #6487, opened 2024-09-17
fix(codegen): fix setting of default signing name

kuhe:
### Issue
n/a

### Description
Remove incorrect fallback attempt to the service arn namespace when setting the default signing name.

### Testing
CI


codegen/smithy-aws-typescript-codegen/src/main/java/software/amazon/smithy/aws/typescript/codegen/AddAwsAuthPlugin.java
```diff
@@ -123,15 +122,9 @@ public void addConfigInterfaceFields(
...
-            settings.setDefaultSigningName(
-                service.getTrait(SigV4Trait.class).map(SigV4Trait::getName)
-                    .orElse(serviceTrait.getArnNamespace())
-            );
-        } catch (Exception e) {
-            LOGGER.warning("Unable to set service default signing name. A SigV4 or Service trait is needed.");
-        }
+        settings.setDefaultSigningName(
+            service.expectTrait(SigV4Trait.class).getName()
+        );
```
kuhe, line 127: This should be allowed because if there is no sigv4 trait, the method already returned on line 110, and `name` is required in the trait.

kstich (approved):

Issue #6502, opened 2024-09-23
Dynamo DB client issue after 3.656.0 release


NikhilUmesh:
### Checkboxes for prior research

- [X] I've gone through [Developer Guide](https://docs.aws.amazon.com/sdk-for-javascript/v3/developer-guide) and [API reference](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest)
- [X] I've checked [AWS Forums](https://forums.aws.amazon.com) and [StackOverflow](https://stackoverflow.com/questions/tagged/aws-sdk-js).
- [X] I've searched for [previous similar issues](https://github.com/aws/aws-sdk-js-v3/issues) and didn't find any solution.

### Describe the bug

While converting the Typescript file to a javascript file, we encounter an issue. Module not found 

### Regression Issue

- [X] Select this option if this issue appears to be a regression.

### SDK version number

3.656.0

### Which JavaScript Runtime is this issue in?

Node.js

### Details of the browser/Node.js/ReactNative version

18

### Reproduction Steps

Converting the file from Typescript to java script for the files containing Dynamo Db dependency. 

### Observed Behavior

`node_modules/@aws-sdk/client-dynamodb/dist-types/DynamoDBClient.d.ts(1,39): error TS7016: Could not find a declaration file for module '@aws-sdk/core/account-id-endpoint'.  `

### Expected Behavior

Successfully create a dynamo DB client to establish the connection, while compiling typescript.

kennu:
I see the same problem after upgrading to latest AWS SDK. When running `tsc`, the error message says:

```this result could not be resolved under your current 'moduleResolution' setting. Consider updating to 'node16', 'nodenext', or 'bundler'```

But if I change that setting in tsconfig.json, I break a bunch of other existing stuff. My settings in tsconfig.json are "module": "commonjs" and no setting for "moduleResolution".

**Comment by kuhe** [comment]
I recommend updating your moduleResolution Node16 because you're running Node.js 18, but a one potential workaround is to add this to a `.d.ts` file within your project.

```ts
// file.d.ts
declare module "@aws-sdk/core/account-id-endpoint" {
  export * from "@aws-sdk/core/dist-types/submodules/account-id-endpoint/index.d";
}
```

We will explore adding this on our side as well.

kennu:
[...]

PR #6505, opened 2024-09-23
Fix aws-sdk-core module import

rohangupta6807:
### Issue
Issue number, if available, prefixed with "#6502"

### Description
While importing the account-id-endpoint from the aws-sdk/core, its not importing for the legacy typescript version.

### Testing
Module got imported successfully while compiling.

### Additional context

### Checklist

---


clients/client-dynamodb/src/DynamoDBClient.ts
```diff
@@ -1,5 +1,5 @@
 // smithy-typescript generated code
-import { AccountIdEndpointMode } from "@aws-sdk/core/account-id-endpoint";
+import { AccountIdEndpointMode } from "@aws-sdk/core";
```
kuhe, line 2: AccountIdEndpointMode is not exported from this package's root

PR #6541, opened 2024-10-03
chore(middleware-user-agent): add feature detection for account-id, s3-express, lib-dynamodb

kuhe:
adds feature detection for:
- s3 express bucket
- dynamodb mapper
- account id endpoints modes
- flexible checksum algorithm name

### DIFF
--- a/packages/middleware-user-agent/src/check-features.ts
+++ b/packages/middleware-user-agent/src/check-features.ts
@@ -1,0 +1,39 @@
+import { setFeature } from "@aws-sdk/core";
+import type { AccountIdEndpointMode } from "@aws-sdk/core/account-id-endpoint";
+import type { AwsHandlerExecutionContext } from "@aws-sdk/types";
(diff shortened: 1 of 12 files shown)


trivikr (approved):

packages/middleware-user-agent/src/check-features.ts
```diff
@@ -0,0 +1,39 @@
...
+ */
+export async function checkFeatures(
+  context: AwsHandlerExecutionContext,
+  config: PreviouslyResolved,
+  args: BuildHandlerArguments<any>
+): Promise<void> {
+  // eslint-disable-next-line
+  const request = args.request as IHttpRequest;
+  if (typeof config.accountIdEndpointMode === "function") {
+    switch (await config.accountIdEndpointMode?.()) {
```
kuhe, line 27: this is a function as of https://github.com/aws/aws-sdk-js-v3/pull/6544

PR #6554, opened 2024-10-09
Add missing generic parameter to ReadableStream type

LinusU:
This was added to the upstream dom typings in the Nov 2018 update.

ref:
https://github.com/microsoft/TypeScript/commit/b830ef82147e9448bd63df17006a086e937feddb 
https://github.com/microsoft/TypeScript/pull/28343
https://github.com/microsoft/TypeScript-DOM-lib-generator/pull/541

### Issue
n/a

### Description
This updates the `ReadableStream` type to match the upstream `dom.d.ts` from which it was mirrored in the first place.

This is needed because otherwise it can result in errors like these:

```text
node_modules/form-data-encoder/lib/index.d.ts:30:15 - error TS2315: Type 'ReadableStream' is not generic.

30     stream(): ReadableStream<Uint8Array> | AsyncIterable<Uint8Array>;
                 ~~~~~~~~~~~~~~~~~~~~~~~~~~


Found 1 error in node_modules/form-data-encoder/lib/index.d.ts:30
```

### Testing
I tested this by making the change in my `node_modules` folder and building and testing our project.

### Additional context

### Checklist

---

### DIFF
--- a/packages/types/src/serde.ts
+++ b/packages/types/src/serde.ts
@@ -26,7 +26,7 @@ declare global {
   /**
    * @public
    */
-  export interface ReadableStream {}
+  export interface ReadableStream<R = any> {}
   /**
    * @public
    */


kuhe (dismissed):

packages/types/src/serde.ts
```diff
@@ -26,7 +26,7 @@ declare global {
   /**
    * @public
    */
-  export interface ReadableStream {}
+  export interface ReadableStream<R = any> {}
```
kuhe, line 29: is this an issue if you have the dom or node types library installed?

LinusU: @kuhe sorry for the late reply

This is how it's defined both in the DOM typings:

https://github.com/microsoft/TypeScript/blob/cb44488fcec4348a448434afbf2ebcbf2b423c61/src/lib/dom.generated.d.ts#L19086

and in the Node.js typings:

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/3c2bac796de55bd7fe6c00bab670829efa6fbc1e/types/node/stream/web.d.ts#L173

I have had no problems using it in our project that has the node typings installed, and I just tested adding DOM to libs and it works as well 👍

kuhe:
this declaration hasn't otherwise been raised as a problem, so I'm closing this PR due to age

Issue #6567, opened 2024-10-16
The type of Status in the DescribeEnvironmentHealthResult interfac is incorrect.


jhpung:
### Checkboxes for prior research

- [X] I've gone through [Developer Guide](https://docs.aws.amazon.com/sdk-for-javascript/v3/developer-guide) and [API reference](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest)
- [X] I've checked [AWS Forums](https://forums.aws.amazon.com) and [StackOverflow](https://stackoverflow.com/questions/tagged/aws-sdk-js).
- [X] I've searched for [previous similar issues](https://github.com/aws/aws-sdk-js-v3/issues) and didn't find any solution.

### Describe the bug

Currently, the type is set to EnvironmentHealth, but it should be specified as EnvironmentStatus to match the response.

### Regression Issue

- [ ] Select this option if this issue appears to be a regression.

### SDK version number

@aws-sdk/client-elastic-beanstalk@3.670.0

### Which JavaScript Runtime is this issue in?

Node.js

### Details of the browser/Node.js/ReactNative version

20.18.0

### Reproduction Steps

1. Send a DescribeEnvironmentHealthCommand
2. The received output and the typeof output's status do not match with EnvironmentHealth type.

### Observed Behavior

DescribeEnvironmentHealthCommandOutput.Status is `Ready` but it's type is [EnvironmentHealth](https://github.com/jhpung/aws-sdk-js-v3/blob/7e7e222e644741e14dee72b09b085b6822e464cb/clients/client-elastic-beanstalk/src/models/models_0.ts#L931-L935) that doesn't matched

### Expected Behavior

DescribeEnvironmentHealthCommandOutput.Status's type should be [EnvironmentStatus](https://github.com/jhpung/aws-sdk-js-v3/blob/7e7e222e644741e14dee72b09b085b6822e464cb/clients/client-elastic-beanstalk/src/models/models_0.ts#L1016-L1025)

kuhe:
@aBurmeseDev please create a ticket with the service team to confirm the intended model structure. I have closed the related PR #6568 because it cannot be modified in this repository.

aBurmeseDev:
Hi @jhpung - thanks for reaching out. 

As my colleague mentioned above, The modifications related to service models need to be handled by the respective service teams upstream and we do not make direct changes to the models within SDK. 

[...]

PR #6568, opened 2024-10-16
fix(client-elastic-beanstalk): update DescribeEnvironmentHealthResult…

jhpung:
….Status with the correct type

### Issue
#6567

### Description
Update DescribeEnvironmentHealthResult.Status with the correct type.

### Testing
I don’t think testing is needed for this PR because it only changes the type of an interface property.

### Additional context
X

### Checklist

---

### DIFF
--- a/clients/client-elastic-beanstalk/src/models/models_0.ts
+++ b/clients/client-elastic-beanstalk/src/models/models_0.ts
@@ -2992,7 +2992,7 @@ export interface DescribeEnvironmentHealthResult {
    *         <code>Updating</code>, <code>Terminating</code>, or <code>Terminated</code>.</p>
    * @public
    */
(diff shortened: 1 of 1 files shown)


clients/client-elastic-beanstalk/src/models/models_0.ts
```diff
@@ -2992,7 +2992,7 @@ export interface DescribeEnvironmentHealthResult {
    *         <code>Updating</code>, <code>Terminating</code>, or <code>Terminated</code>.</p>
    * @public
    */
-  Status?: EnvironmentHealth;
+  Status?: EnvironmentStatus;
```
kuhe, line 2995: this file is generated from a model data source and not eligible for update from this repository

PR #6584, opened 2024-10-22
test: convert some packages unit and e2e tests to vitest

kuhe:
- converts `lib` unit and e2e to vitest.
- converts `package` unit and e2e to vitest for packages starting with letters A-E.

### Testing
running the tests


lib/lib-dynamodb/src/baseCommand/DynamoDBDocumentClientCommand.spec.ts
```diff
@@ -34,8 +35,7 @@ class AnyCommand extends DynamoDBDocumentClientCommand<{}, {}, {}, {}, {}> {
 }
 
 describe("DynamoDBDocumentClientCommand", () => {
-  // ToDo: Investigate why Jest29 throws TypeError: Class constructor Command cannot be invoked without 'new'
-  it.skip("should not allow usage of the default middlewareStack", () => {
+  it("should not allow usage of the default middlewareStack", () => {
```
kuhe, line 38: one benefit shown here, it works around certain Jest problems

vitest.config.ts
```diff
@@ -0,0 +1,29 @@
...
+      "packages/credential-provider-sso/**/*.spec.{ts,js}",
+      "packages/credential-provider-web-identity/**/*.spec.{ts,js}",
+      "packages/credential-providers/**/*.spec.{ts,js}",
+      "packages/ec2-metadata-service/**/*.spec.{ts,js}",
+      "packages/endpoint-cache/**/*.spec.{ts,js}",
+      "packages/eventstream-handler-node/**/*.spec.{ts,js}",
+    ],
+    environment: "node",
+  },
+});
```
kuhe, line 29: This config file explicitly lists out packages because not all .spec.ts files are vitest compatible.

In CI, `test:unit` is run by turbo, so this file is not used, and does not need to be maintained. When conversion is complete, the inclusion will be wild-card based instead of curated, and only used if a developer wishes to run `npx vite run` etc. from the root.

vitest.config.e2e.ts
```diff
@@ -0,0 +1,8 @@
+import { defineConfig } from "vitest/config";
+
+export default defineConfig({
+  test: {
+    include: ["lib/lib-dynamodb/**/*.e2e.spec.{ts,js}", "lib/lib-storage/**/*.e2e.spec.{ts,js}"],
+    environment: "node",
+  },
+});
```
kuhe, line 8: This is not part of CI. It is a temporary local-dev file for running vitest e2e tests from the application root.

trivikr (approved):

PR #6619, opened 2024-11-04
test(credential-provider-node): remove sts and sso client mocks

kuhe:
### Issue
Mocking of STS caused a bug in https://github.com/aws/aws-sdk-js-v3/releases/tag/v3.667.0 to not be caught.

### Description
Removes the mocking of the STS and SSO clients in the credential provider chain's integration test. The mocking is moved to the HTTP Handler level.

### Testing
CI / running these integration tests.

Verifying that this would've caught the issue in v3.667.0:

```
git checkout v3.667.0
make tpk # builds packages only
b cred prov node - integ # runs integration tests, hangs
```

```
git checkout v3.668.0 # patched version
make tpk # builds packages only
b cred prov node - integ # runs integration tests, succeeds
```

### DIFF
--- a/packages/credential-provider-node/src/credential-provider-node.integ.spec.ts
+++ b/packages/credential-provider-node/src/credential-provider-node.integ.spec.ts
@@ -3,7 +3,7 @@ import * as credentialProviderHttp from "@aws-sdk/credential-provider-http";
 import { HttpResponse } from "@smithy/protocol-http";
 import type { SourceProfileInit } from "@smithy/shared-ini-file-loader";
 import type { HttpRequest, NodeHttpHandlerOptions, ParsedIniData } from "@smithy/types";
-import { Readable } from "stream";
+import { PassThrough } from "stream";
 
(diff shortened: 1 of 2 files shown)


packages/credential-provider-node/src/credential-provider-node.integ.spec.ts
```diff
@@ -474,18 +430,13 @@ describe("credential-provider-node integration test", () => {
...
         $source: {
           CREDENTIALS_PROFILE_SOURCE_PROFILE: "o",
           CREDENTIALS_STS_ASSUME_ROLE: "i",
         },
       });
     });
 
-    xit("should prefer static credentials over role assumption metadata only if the profile is not the first one visited", async () => {
-      // TODO
-    });
```
kuhe, line 487: This test case isn't necessary, because other tests cover it.

I wrote this initially when I was going over the default credential provider chain and it looked like a distinct use-case, but isn't.

trivikr (approved):

PR #6625, opened 2024-11-05
test(client-ec2): convert cucumber tests to vitest for ec2

kuhe:
### Issue
n/a

### Description
Converts the cucumber tests for ec2 into Vitest and places them in the ec2 middleware package (the tests mainly cover what is in the middleware package, not the client itself).

Benefits:
- test is now in typescript (caught an old typo)
- cucumber tests run much faster without the EC2 test taking up 3 minutes.

### Testing
Ran the tests in the PR.

### Checklist


features/ec2/step_definitions/ec2.js
```diff
@@ -1,175 +0,0 @@
...
-          },
-        ],
-      });
-      dstSnapId = copySnapshot.SnapshotId;
-
-      // this waiter is not doing what I think it does for copied snapshot in destination region.
-      // await waitUntilSnapshotCompleted(
-      //   { client: dstEc2, maxWaitTime: 20 },
-      //   {
-      //     SnapshotId: copySnapshot.SnapshotId,
```
kuhe, line 128: this waiter call was wrong, which is why it didn't work. Revealed by TypeScript.

features/ec2/step_definitions/ec2.js
```diff
@@ -1,175 +0,0 @@
...
-          },
-        ],
-      });
-
-      srcSnapId = createSnapshot.SnapshotId;
-
-      await waitUntilSnapshotCompleted(
-        { client: srcEc2, maxWaitTime: 300 },
-        {
-          SnapshotId: createSnapshot.SnapshotId,
```
kuhe, line 105: this waiter also had the same typo, but for whatever reason passed. Must be a false positive, with `SnapshotIds[]` considered empty and passing.

trivikr (approved):

PR #6629, opened 2024-11-06
chore(middleware-flexible-checksums): delay checksum validation until stream read

trivikr:
### Issue
Internal JS-5537

### Description
Delays checksum validation until stream read

### Testing

```js
import assert from "assert";
// import { S3 } from "../aws-sdk-js-v3/clients/client-s3/dist-cjs/index.js";

const s3 = new S3({ region: "us-west-2" });
const Bucket = "test-flexible-checksums"; // Change to your bucket name
const Key = "test.txt";
const Body = "helloworld".repeat(10_000_000); // A string of size ~100MB

await s3.putObject({ Bucket, Key, Body, ChecksumAlgorithm: "CRC32" });

const apiCallStart = performance.now();
const response = await s3.getObject({ Bucket, Key, ChecksumMode: "ENABLED" });
console.log("time taken for request:", (performance.now() - apiCallStart).toFixed(2), "ms");

const transformStart = performance.now();
const body = await response.Body.transformToString();
console.log("time taken for transform:", (performance.now() - transformStart).toFixed(2), "ms");

assert(body === Body);
```

#### Before

Response checksum is validated when API call is made.
```console
time taken for request: 3141.02 ms
time taken for transform: 221.46 ms
```

#### After

Response checksum is validation when stream is consumed (i.e. when transform call is made)
```console
time taken for request: 1217.04 ms
time taken for transform: 2141.58 ms
```

---

### DIFF
--- a/packages/middleware-flexible-checksums/src/validateChecksumFromResponse.ts
+++ b/packages/middleware-flexible-checksums/src/validateChecksumFromResponse.ts
@@ -35,7 +35,7 @@ export const validateChecksumFromResponse = async (
       const { base64Encoder } = config;
 
       if (isStreaming(responseBody)) {
-        createChecksumStream({
+        response.body = createChecksumStream({
           expectedChecksum: checksumFromResponse,
           checksumSourceLocation: responseHeader,
           checksum: new checksumAlgorithmFn() as Checksum,
--- a/packages/middleware-flexible-checksums/src/validateChecksumFromResponse.ts
+++ b/packages/middleware-flexible-checksums/src/validateChecksumFromResponse.ts
@@ -1,10 +1,13 @@
 import { HttpResponse } from "@smithy/protocol-http";
+import { Checksum } from "@smithy/types";
+import { createChecksumStream } from "@smithy/util-stream";
 
 import { PreviouslyResolved } from "./configuration";
 import { ChecksumAlgorithm } from "./constants";
 import { getChecksum } from "./getChecksum";
 import { getChecksumAlgorithmListForResponse } from "./getChecksumAlgorithmListForResponse";
 import { getChecksumLocationName } from "./getChecksumLocationName";
+import { isStreaming } from "./isStreaming";
 import { selectChecksumAlgorithmFunction } from "./selectChecksumAlgorithmFunction";
 
 export interface ValidateChecksumFromResponseOptions {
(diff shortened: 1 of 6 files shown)


packages/middleware-flexible-checksums/src/validateChecksumFromResponse.ts
```diff
@@ -29,9 +32,20 @@ export const validateChecksumFromResponse = async (
     const checksumFromResponse = responseHeaders[responseHeader];
     if (checksumFromResponse) {
       const checksumAlgorithmFn = selectChecksumAlgorithmFunction(algorithm as ChecksumAlgorithm, config);
-      const { streamHasher, base64Encoder } = config;
-      const checksum = await getChecksum(responseBody, { streamHasher, checksumAlgorithmFn, base64Encoder });
+      const { base64Encoder } = config;
 
+      if (isStreaming(responseBody)) {
+        createChecksumStream({
```
kuhe, line 38: Isn't the return value of this needed?

trivikr: The assert in tests show that body is getting transformed, and timing indicates checksum validation is happening, although it's not verified. Is it because the Checksum Duplex Stream is created, but is not consumed?

I'll assign the return value to body, since that's how the Duplex stream is designed to be used.

trivikr: response.body was updated in [`e3d92aa` (#6629)](https://github.com/aws/aws-sdk-js-v3/pull/6629/commits/e3d92aada9238d90045ca0c9746509312334394c)

packages/middleware-flexible-checksums/src/validateChecksumFromResponse.ts
```diff
@@ -35,7 +35,7 @@ export const validateChecksumFromResponse = async (
       const { base64Encoder } = config;
 
       if (isStreaming(responseBody)) {
-        createChecksumStream({
+        response.body = createChecksumStream({
```
kuhe, line 38: now that the body stream has a checksum wrapper, the section below where we perform the checksum validation should be removed

```js
      const checksum = await getChecksum(responseBody, { checksumAlgorithmFn, base64Encoder });
      if (checksum === checksumFromResponse) {
        // The checksum for response payload is valid.
        break;
      }
      throw new Error(
        `Checksum mismatch: expected "${checksum}" but received "${checksumFromResponse}"` +
          ` in response header "${responseHeader}".`
      );
```

this function should be renamed `validateChecksumFromResponse` -> `applyChecksumStreamWrapper` or something like that to reflect its new functionality.

kuhe: the ChecksumStream wrapper will do the error-throwing when the user reads the stream to completion

trivikr: The wrapper is used only when response.body is streaming. That's why there's a return on line 45.
For non-streaming use cases, the stringHasher needs to validate the checksum.

kuhe (approved):

packages/middleware-flexible-checksums/src/validateChecksumFromResponse.ts
kuhe: Oh, I understand now

trivikr:
E2E test was added in [`d9d7534` (#6629)](https://github.com/aws/aws-sdk-js-v3/pull/6629/commits/d9d75346c3e5fca11bb09fa8fad59c367f0dd6d4)

PR #6640, opened 2024-11-07
chore: experiment with wasm-fmt/biome

kuhe:
clone this PR and then run `yarn generate:client` in clients/client-s3.


kuhe:
see https://github.com/aws/aws-sdk-js-v3/pull/7134

Issue #6687, opened 2024-11-22
SSO requests ignore proxy configuration


otaviomacedo:
### Checkboxes for prior research

- [X] I've gone through [Developer Guide](https://docs.aws.amazon.com/sdk-for-javascript/v3/developer-guide) and [API reference](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest)
- [X] I've checked [AWS Forums](https://forums.aws.amazon.com) and [StackOverflow](https://stackoverflow.com/questions/tagged/aws-sdk-js).
- [X] I've searched for [previous similar issues](https://github.com/aws/aws-sdk-js-v3/issues) and didn't find any solution.

### Describe the bug

The SDK gets credentials via SSO directly (ignoring proxy configuration) provided in the `fromIni()` method.

### Regression Issue

- [ ] Select this option if this issue appears to be a regression.

### SDK version number

@aws-sdk/credential-providers@3.696.0

### Which JavaScript Runtime is this issue in?

Node.js

### Details of the browser/Node.js/ReactNative version

node version: v22.9.0

### Reproduction Steps

0. Precondition: the SSO set-up has been done (creating user, `aws sso login` etc).
1. Start a proxy on port 8080. I'm using [mitmproxy](https://mitmproxy.org/).
2. [Install](https://docs.mitmproxy.org/stable/concepts-certificates/) the CA certificate.
3. Run the following code:

```ts
    const agent = new ProxyAgent({
      ca: fs.readFileSync('/path/to/pem', { encoding: 'utf-8' }),
      getProxyForUrl: () => Promise.resolve('http://localhost:8080'),
    });

    const provider = fromIni({
      profile: 'test',
      ignoreCache: true,
      clientConfig: {
        requestHandler: {
          connectionTimeout: 30000,
          requestTimeout: 10000,
          httpsAgent: agent,
          httpAgent: agent,
        },
      },
    });

    console.log(await provider());
```

4. Watch the proxy logs.

### Observed Behavior

The TypeScript program will print the AWS credentials, but the proxy won't show any activity. To further confirm, if I shut down the proxy, the credentials are still retrieved.

### Expected Behavior

Proxy show activity when up and request fail when proxy is down.

### Additional Information/Context

Related issue: https://github.com/aws/aws-cdk/issues/32208

kuhe:
a fix for this was released in https://github.com/aws/aws-sdk-js-v3/releases/tag/v3.699.0

PR #6688, opened 2024-11-22
fix(credential-provider-ini): pass clientConfig to sso and sso-oidc inner clients

kuhe:
### Issue
addresses https://github.com/aws/aws-sdk-js-v3/issues/6687

### Description
This change ensures `clientConfig`, a property shared by many credential providers, is used as the initialization config for all inner clients including SSO and SSOOIDC. Currently, it is only passed through for CognitoIdentity and STS clients. 

### Testing
manual SSO testing
- [x] add unit tests

### DIFF
--- a/packages/token-providers/src/getSsoOidcClient.ts
+++ b/packages/token-providers/src/getSsoOidcClient.ts
@@ -1,23 +1,18 @@
-const ssoOidcClientsHash: Record<string, any> = {};
+import { FromSsoInit } from "./fromSso";
 
 /**
- * Returns a SSOOIDC client for the given region. If the client has already been created,
- * it will be returned from the hash.
+ * Returns a SSOOIDC client for the given region.
  * @internal
  */
-export const getSsoOidcClient = async (ssoRegion: string) => {
+export const getSsoOidcClient = async (ssoRegion: string, init: FromSsoInit = {}) => {
   // @ts-ignore Cannot find module '@aws-sdk/client-sso-oidc'
   const { SSOOIDCClient } = await import("@aws-sdk/client-sso-oidc");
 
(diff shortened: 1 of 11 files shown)


packages/token-providers/src/getSsoOidcClient.ts
```diff
@@ -1,23 +1,18 @@
-const ssoOidcClientsHash: Record<string, any> = {};
```
kuhe, line 1: I removed this caching functionality. Credentials are cached already, there should be no need to cache the client as well. Caching the client conflicts with being able to configure the client, since this cache exposes no method to clear it.

packages/token-providers/src/getSsoOidcClient.ts
```diff
@@ -1,23 +1,18 @@
...
-  }
-
-  // Create new SSOOIDC client, and store is in hash.
-  // If we need to support configuration of SsoOidc client in future through code,
-  // the provision to pass region from client configuration needs to be added.
-  const ssoOidcClient = new SSOOIDCClient({ region: ssoRegion });
-  ssoOidcClientsHash[ssoRegion] = ssoOidcClient;
+  const ssoOidcClient = new SSOOIDCClient(
+    Object.assign({}, init.clientConfig ?? {}, {
+      region: ssoRegion ?? init.clientConfig.region,
```
siddsriv, line 13: just to double check, is there a case where `init.clientConfig` is undefined? should we use optional chaining here too?

kuhe: fixed, but I also noticed that ssoRegion will never need a fallback anyway

siddsriv (approved):

PR #6726, opened 2024-12-11
feat(credential-providers): make credential providers aware of caller client region

kuhe:
We [previously](https://github.com/aws/aws-sdk-js-v3/pull/5758) introduced the concept of a "parentClientConfig" for credential providers that create another (inner) SDK client such as STS. 

One problem was that if the provider was initialized outside of a client, this connection could not be made. This PR makes AWS credential providers accept a contextual client config, making them region aware at call time. The result is improved intuitiveness of default behavior, as shown below.

### currently

⚠️ when using a CODE level credential configuration, any potential spawning of an inner client such as STS for AssumeRole must be passed the region in a redundant manner. This is because `fromIni` is a standalone function that can resolve credentials without the context of an SDK client.

```js
import { DynamoDB } from "@aws-sdk/client-dynamodb";
import { fromIni } from "@aws-sdk/credential-providers";

const ddb = new DynamoDB({
  region: "eu-west-1",
  credentials: fromIni({
    clientConfig: { region: "eu-west-1" }, // required, but often forgotten
  }),
});
```

### in this PR

Improved intuitiveness of credentials region resolution.

```js
import { DynamoDB } from "@aws-sdk/client-dynamodb";
import { fromIni } from "@aws-sdk/credential-providers";

const ddb = new DynamoDB({
  region: "eu-west-1",
  // will default to the client's region instead of the commercial partition default (us-east-1)
  credentials: fromIni({}), 
});

// initialized outside a client
const credentialsProvider = fromIni();

// uses global default of us-east-1 as before, when invoked outside of a client.
const credentials = await credentialsProvider();

const ddb = new DynamoDB({
  region: "eu-west-1",
  credentials, // will default to the client's region if supplied to a client
});
```

### DIFF
--- a/packages/core/src/submodules/httpAuthSchemes/aws_sdk/resolveAwsSdkSigV4Config.ts
+++ b/packages/core/src/submodules/httpAuthSchemes/aws_sdk/resolveAwsSdkSigV4Config.ts
@@ -129,8 +129,8 @@ export const resolveAwsSdkSigV4Config = <T>(
     }
   }
 
-  const contextBoundCredentialsProvider = async () => {
-    return credentialsProvider!({ contextClientConfig: config });
+  const boundCredentialsProvider = async () => {
+    return credentialsProvider!({ callerClientConfig: config });
   };
 
   // Populate sigv4 arguments
@@ -174,7 +174,7 @@ export const resolveAwsSdkSigV4Config = <T>(
 
           const params: SignatureV4Init & SignatureV4CryptoInit = {
             ...config,
-            credentials: contextBoundCredentialsProvider,
+            credentials: boundCredentialsProvider,
             region: config.signingRegion,
             service: config.signingName,
             sha256,
@@ -210,7 +210,7 @@ export const resolveAwsSdkSigV4Config = <T>(
 
       const params: SignatureV4Init & SignatureV4CryptoInit = {
         ...config,
-        credentials: contextBoundCredentialsProvider,
+        credentials: boundCredentialsProvider,
         region: config.signingRegion,
         service: config.signingName,
         sha256,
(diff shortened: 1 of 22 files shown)


packages/core/src/submodules/httpAuthSchemes/aws_sdk/resolveAwsSdkSigV4Config.ts
```diff
@@ -106,29 +106,33 @@ export const resolveAwsSdkSigV4Config = <T>(
...
-      normalizedCreds = async () => {
+      credentialsProvider = async () => {
         throw new Error("`credentials` is missing");
       };
     }
   }
 
+  const boundCredentialsProvider = async () => {
+    return credentialsProvider!({ callerClientConfig: config });
+  };
```
trivikr, line 134: nit: brackets can be removed
```suggestion
  const boundCredentialsProvider = async () => credentialsProvider!({ callerClientConfig: config });
```

packages/credential-provider-ini/src/resolveAssumeRoleCredentials.ts
```diff
@@ -116,13 +117,18 @@ export const resolveAssumeRoleCredentials = async (
       {
         ...options.clientConfig,
         credentialProviderLogger: options.logger,
-        parentClientConfig: options?.parentClientConfig,
+        parentClientConfig: {
+          ...options?.parentClientConfig,
+          // The profile region is the last fallback, and only applies
+          // if the clientConfig.region is not defined by the user
+          // and no contextual outer client configuration region can be found.
+          region: options?.parentClientConfig?.region ?? region,
```
trivikr, line 125: Should region specification appear before overriding with parentClientConfig?

kuhe: Yes, I was 50/50 during the design phase, but since you pointed it out I believe it makes more sense with higher priority.

trivikr (approved):

PR #6729, opened 2024-12-13
fix(util-user-agent-node): appId config file name correction

siddsriv:
shared config file setting name correction for appId

---

### DIFF
--- a/packages/util-user-agent-node/src/nodeAppIdConfigOptions.ts
+++ b/packages/util-user-agent-node/src/nodeAppIdConfigOptions.ts
@@ -14,7 +14,7 @@ export const UA_APP_ID_INI_NAME = "sdk_ua_app_id";
 /**
  * @deprecated
  */
-export const UA_APP_ID_INI_NAME_DEPRECATED = "sdk-ua-app-id";
+const UA_APP_ID_INI_NAME_DEPRECATED = "sdk-ua-app-id";
 
 /**
  * @internal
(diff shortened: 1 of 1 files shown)


packages/util-user-agent-node/src/nodeAppIdConfigOptions.ts
```diff
@@ -10,13 +9,18 @@ export const UA_APP_ID_ENV_NAME = "AWS_SDK_UA_APP_ID";
 /**
  * @internal
  */
-export const UA_APP_ID_INI_NAME = "sdk-ua-app-id";
+export const UA_APP_ID_INI_NAME = "sdk_ua_app_id";
+
+/**
+ * @internal
```
kuhe, line 15: add `@deprecated` to this one

kuhe: also, don't export it

packages/util-user-agent-node/src/nodeAppIdConfigOptions.ts
```diff
@@ -10,13 +9,18 @@ export const UA_APP_ID_ENV_NAME = "AWS_SDK_UA_APP_ID";
 /**
  * @internal
  */
-export const UA_APP_ID_INI_NAME = "sdk-ua-app-id";
+export const UA_APP_ID_INI_NAME = "sdk_ua_app_id";
+
+/**
+ * @deprecated
+ */
+export const UA_APP_ID_INI_NAME_DEPRECATED = "sdk-ua-app-id";
```
kuhe, line 17:
```suggestion
const UA_APP_ID_INI_NAME_DEPRECATED = "sdk-ua-app-id";
```

PR #6492, opened 2024-09-18
chore(middleware-flexible-checksums): use RequestChecksumCalculation and ResponseChecksumValidation

trivikr:
### Issue
* Internal JS-5396
* Alternative to https://github.com/aws/aws-sdk-js-v3/pull/6484 with `flexibleChecksumsInterceptorMiddleware`

### Description
* Uses values in RequestChecksumCalculation and ResponseChecksumValidation to decide flexible checksums algorithm.
* Adds `flexibleChecksumsInterceptorMiddleware` to populate `input[requestValidationModeMember]`
* Deprecates MD5, and sets CRC32 as default.

### Testing
Unit testing in CI

Also, verified that checksums are computed by default
```js
import { S3 } from "../aws-sdk-js-v3/clients/client-s3/dist-cjs/index.js";
import { NodeHttpHandler } from "@smithy/node-http-handler";
import { equal } from "assert";

// Prints checksum headers for request and response.
class CustomHandler extends NodeHttpHandler {
  constructor() {
    super();
  }

  printChecksumHeaders(prefix, headers) {
    for (const [header, value] of Object.entries(headers)) {
      if (header.startsWith("x-amz-checksum-") || header.startsWith("x-amz-sdk-checksum-")) {
        console.log(`${prefix}['${header}']: '${value}'`);
      }
    }
  }

  async handle(request, options) {
    this.printChecksumHeaders("request", request.headers);
    const response = await super.handle(request, options);
    this.printChecksumHeaders("response", response.response.headers);
    return response;
  }
}

// WHEN_SUPPORTED is default.
const client = new S3({
  requestHandler: new CustomHandler(),
});
const Bucket = "test-flexible-checksums-v2"; // Replace with your test bucket name.
const Key = "hello-world.txt";
const Body = "Hello World"; // Replace with the content you want to test.

console.log("Put Object");
// ChecksumAlgorithm is not explicitly required to be specified.
await client.putObject({ Bucket, Key, Body });

console.log("\nGet Object");
// ChecksumMode is not explicitly required to be specified.
const response = await client.getObject({ Bucket, Key });

equal(Body, await response.Body.transformToString());
```

Output
```console
Put Object
request['x-amz-sdk-checksum-algorithm']: 'CRC32'
request['x-amz-checksum-crc32']: 'ShexVg=='
response['x-amz-checksum-crc32']: 'ShexVg=='
response['x-amz-checksum-type']: 'FULL_OBJECT'

Get Object
[...]

PR #6750, opened 2024-12-19
chore(middleware-flexible-checksums): perform checksum calculation and validation by default

trivikr:
### Issue
* Internal JS-5396
* Split from https://github.com/aws/aws-sdk-js-v3/pull/6492
* Should be rebased after https://github.com/aws/aws-sdk-js-v3/pull/6749 is merged

### Description
Performs checksum calculation and validation by default based on values in RequestChecksumCalculation and ResponseChecksumValidation

### Testing
Unit testing in CI

Verified that CRC32 checksums are computed by default
```js
import { S3 } from "../aws-sdk-js-v3/clients/client-s3/dist-cjs/index.js";
import { NodeHttpHandler } from "../aws-sdk-js-v3/node_modules/@smithy/node-http-handler/dist-cjs/index.js";
import { equal } from "assert";

// Prints checksum headers for request and response.
class CustomHandler extends NodeHttpHandler {
  constructor() {
    super();
  }

  printChecksumHeaders(prefix, headers) {
    for (const [header, value] of Object.entries(headers)) {
      if (
        header.startsWith("x-amz-checksum-") ||
        header.startsWith("x-amz-sdk-checksum-")
      ) {
        console.log(`${prefix}['${header}']: '${value}'`);
      }
    }
  }

  async handle(request, options) {
    this.printChecksumHeaders("request", request.headers);
    const response = await super.handle(request, options);
    this.printChecksumHeaders("response", response.response.headers);
    return response;
  }
}

// WHEN_SUPPORTED is default.
const client = new S3({
  requestHandler: new CustomHandler(),
});
const Bucket = "test-flexible-checksums-v2"; // Replace with your test bucket name.
const Key = "hello-world.txt";
const Body = "Hello World"; // Replace with the content you want to test.

console.log("Put Object");
// ChecksumAlgorithm is not explicitly required to be specified.
await client.putObject({ Bucket, Key, Body });

console.log("\nGet Object");
// ChecksumMode is not explicitly required to be specified.
const response = await client.getObject({ Bucket, Key });

equal(Body, await response.Body.transformToString());
```

Output
```js
Put Object
request['x-amz-sdk-checksum-algorithm']: 'CRC32'
request['x-amz-checksum-crc32']: 'ShexVg=='
response['x-amz-checksum-crc32']: 'ShexVg=='
response['x-amz-checksum-type']: 'FULL_OBJECT'

Get Object
request['x-amz-checksum-mode']: 'ENABLED'
response['x-amz-checksum-crc32']: 'ShexVg=='
response['x-amz-checksum-type']: 'FULL_OBJECT'
```

---

### DIFF
--- a/packages/middleware-flexible-checksums/src/getChecksumAlgorithmForRequest.spec.ts
+++ b/packages/middleware-flexible-checksums/src/getChecksumAlgorithmForRequest.spec.ts
@@ -65,7 +65,7 @@ describe(getChecksumAlgorithmForRequest.name, () => {
     };
     expect(() => {
       getChecksumAlgorithmForRequest(mockInput, mockOptions);
-    }).toThrow(
+    }).toThrowError(
       `The checksum algorithm "${unsupportedAlgo}" is not supported by the client.` +
         ` Select one of ${CLIENT_SUPPORTED_ALGORITHMS}.`
     );
--- a/packages/middleware-flexible-checksums/src/getChecksumAlgorithmForRequest.spec.ts
+++ b/packages/middleware-flexible-checksums/src/getChecksumAlgorithmForRequest.spec.ts
@@ -1,6 +1,6 @@
 import { describe, expect, test as it } from "vitest";
 
-import { ChecksumAlgorithm, DEFAULT_CHECKSUM_ALGORITHM } from "./constants";
+import { DEFAULT_CHECKSUM_ALGORITHM, RequestChecksumCalculation } from "./constants";
 import { getChecksumAlgorithmForRequest } from "./getChecksumAlgorithmForRequest";
 import { CLIENT_SUPPORTED_ALGORITHMS } from "./types";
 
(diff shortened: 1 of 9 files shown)


packages/middleware-flexible-checksums/src/getChecksumAlgorithmForRequest.spec.ts
```diff
@@ -1,43 +1,71 @@
...
-    const mockOptions = { requestChecksumRequired: true, requestAlgorithmMember: mockRequestAlgorithmMember };
+    const mockOptions = {
+      requestChecksumRequired: true,
+      requestAlgorithmMember: mockRequestAlgorithmMember,
+      requestChecksumCalculation: RequestChecksumCalculation.WHEN_REQUIRED,
+    };
     expect(() => {
       getChecksumAlgorithmForRequest(mockInput, mockOptions);
-    }).toThrowError(
+    }).toThrow(
```
kuhe, line 68: what's the difference between toThrowError and toThrow?

kuhe (approved):
Add more tests in middleware-flexible-checksums.integ.spec.ts?

packages/middleware-flexible-checksums/src/getChecksumAlgorithmForRequest.spec.ts
trivikr: They're same as per maintainer of vitest https://github.com/vitest-dev/vitest/discussions/2570#discussioncomment-4503284

trivikr: Changed to specific `toThrowError` in [`b49d70e` (#6750)](https://github.com/aws/aws-sdk-js-v3/pull/6750/commits/b49d70e8c754b20f170d7fe240f1f2773c3c16b6)

trivikr:
> Add more tests in middleware-flexible-checksums.integ.spec.ts?

The tests were added in `clients/client-s3/test/unit/flexibleChecksums.spec.ts` in prior diff.
However, the integration tests need to as close to the source code as possible - so generic tests were moved to middleware package in https://github.com/aws/aws-sdk-js-v3/pull/6752

And the tests were added to updated file in [`d6af53c` (#6750)](https://github.com/aws/aws-sdk-js-v3/pull/6750/commits/d6af53c6f56968e87e0d3f91bb63de283d4522ea)

PR #6791, opened 2025-01-09
fix(nested-clients): create nested clients for internal use

kuhe:
- **feat(nested-clients): create nested clients for internal sts usage**
- **chore(nested-clients): generate nested clients**

### Issue

### Description

### Testing
How was this change tested?

### Additional context

### Checklist

---


packages/nested-clients/src/index.ts
```diff
@@ -0,0 +1,25 @@
+/**
+ * @internal
+ */
+export { CreateTokenCommand, SSOOIDCClient } from "./nested-sso-oidc/index";
```
trivikr, line 4: Should we export sso under `@aws-sdk/nested-clients/sso`?
That way a conflict between commands can be avoided.

packages/nested-clients/src/nested-sso-oidc/SSOOIDC.ts
```diff
@@ -0,0 +1,67 @@
+// smithy-typescript generated code
+import { createAggregatedClient } from "@smithy/smithy-client";
```
trivikr, line 2: The module name already has `nested-` prefix.
I think it can be removed from name of the folder?

packages/nested-clients/src/index.ts
kuhe: sounds good, I can port the submodule system from core over here. I'll hold off on merging for now.

packages/nested-clients/src/nested-sso-oidc/SSOOIDC.ts
kuhe: will remove as part of introducing submodules in this package.

kuhe:
- [ ] create submodules using package.json imports similar to what was done in `@aws-sdk/core`.

trivikr (approved):

PR #6798, opened 2025-01-14
fix(crc64-nvme-crt): return checksum for empty string if called without data

trivikr:
### Issue
Internal JS-5674

### Description
Returns checksum for empty string if called without data

### Testing

Verified using internal test code that CRC64VME checksum if `"AAAAAAAAAAA="` when called for empty string.

---

### DIFF
--- a/packages/crc64-nvme-crt/src/CrtCrc64Nvme.ts
+++ b/packages/crc64-nvme-crt/src/CrtCrc64Nvme.ts
@@ -3,20 +3,17 @@ import { Checksum } from "@smithy/types";
 import { toUint8Array } from "@smithy/util-utf8";
 
 export class CrtCrc64Nvme implements Checksum {
-  private previous: DataView | undefined;
+  private checksum: DataView = new DataView(new ArrayBuffer(8));
 
-  update(chunk: Uint8Array) {
-    this.previous = checksums.crc64nvme(chunk, this.previous);
+  update(data: Uint8Array) {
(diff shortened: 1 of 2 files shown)


siddsriv (approved):

packages/crc64-nvme-crt/src/CrtCrc64Nvme.ts
```diff
@@ -3,20 +3,17 @@ import { Checksum } from "@smithy/types";
 import { toUint8Array } from "@smithy/util-utf8";
 
 export class CrtCrc64Nvme implements Checksum {
-  private previous: DataView | undefined;
+  private checksum: DataView = new DataView(new ArrayBuffer(8));
```
kuhe, line 6: very minor issue:

given that `CrtCrc64Nvme implements Checksum`, an instance of this may be called "checksum", so this private property would be `checksum.checksum`. 

perhaps it would more correctly be called "hash"?

kuhe (approved):

packages/crc64-nvme-crt/src/CrtCrc64Nvme.ts
trivikr: I'll update this in a separate PR, since we use `checksum` for the private member in `NodeCrc32` too.

https://github.com/aws/aws-sdk-js-v3/blob/2293f5a86626aba323de57cd65a6d9b4e46e9aed/packages/middleware-flexible-checksums/src/getCrc32ChecksumAlgorithmFunction.ts#L7

Issue #6565, opened 2024-10-15
[Feature Request] in fromTemporaryCredentials, get the region for the STSClient from the client that's using the temporary credentials


jedwards1211:
### Describe the feature

We were getting a "Region is missing" error when doing this:

```ts
const ec2 = new EC2Client({
  region: 'us-west-2',
  credentials: fromTemporaryCredentials({
    params: {
      RoleArn: '...',
      RoleSessionName: '...',
      DurationSeconds: 3600,
    },
    masterCredentials: { ... },
  }),
})
const { Images = [] } = await ec2.send(
  new DescribeImagesCommand({ ImageIds: [ImageId] })
)
```

This threw us for a loop, because `region: 'us-west-2'` is right there.

After a lot of debugging, turns out it's because the `STSClient` created by `fromTemporaryCredentials` doesn't get that `region` setting.

The workaround is to pass `clientConfig: { region: 'us-west-2' }` to `fromTemporaryCredentials`, but I don't think we should have to do this.  It's not obvious that we need to do this since `clientConfig` is optional, the "Region is missing" error message only added to the confusion, and it took a lot of debugging to figure out what was actually the problem.  Should be easier.

### Use Case

Doing virtually anything with temporary credentials

### Proposed Solution

`fromTemporaryCredentials` should somehow be able to get the `region` of the enclosing client it is passed as `credentials` to, so that we don't have to explicitly pass `clientConfig: { region: ... }`.  This would be more user-friendly, ergonomic behavior.

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### SDK version used

3.651.1

### Environment details (OS name and version, etc.)

macOS Sonoma 14.4.1

zshzbh:
@jedwards1211 ,

Thanks for the feedback!  I can confirm this issue. Will talk to the team regarding the next steps!

zshzbh:
Hey @jedwards1211 

I upgraded the package version and the error is gone. Please let me know if the issue persists.
Code I have: 
```js
const client = new EC2Client({
  region: "us-east-1",
  credentials: fromTemporaryCredentials({
    //region: "us-east-1",
    params: {
      RoleArn: "arn:aws:iam::XXXXXXX:role/ec2",
      RoleSessionName: "session1",
    },
  }),
});
const command = new DescribeInstancesCommand({});
const response = await client.send(command);
```

[...]

PR #6838, opened 2025-01-22
feat(credential-providers): pass caller client options to fromTemporaryCredentials inner STSClient

kuhe:
### Issue
https://github.com/aws/aws-sdk-js-v3/issues/6565

### Description
- caller client options for region, profile, requestHandler, logger, credentialDefaultProvider are passed to the `fromTemporaryCredentials` inner STS client unless overridden with `fromTemporaryCredentials({ clientConfig })` fields.

### Testing
new unit tests
existing credential integration tests

### DIFF
--- a/packages/nested-clients/src/submodules/sts/defaultStsRoleAssumers.ts
+++ b/packages/nested-clients/src/submodules/sts/defaultStsRoleAssumers.ts
@@ -84,7 +84,7 @@ const resolveRegion = async (
  */
 export const getDefaultRoleAssumer = (
   stsOptions: STSRoleAssumerOptions,
-  stsClientCtor: new (options: STSClientConfig) => STSClient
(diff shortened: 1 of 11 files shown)


packages/nested-clients/src/submodules/sts/defaultStsRoleAssumers.ts
```diff
@@ -84,7 +84,7 @@ const resolveRegion = async (
  */
 export const getDefaultRoleAssumer = (
   stsOptions: STSRoleAssumerOptions,
-  stsClientCtor: new (options: STSClientConfig) => STSClient
+  STSClient: new (options: STSClientConfig) => STSClient
```
kuhe, line 87: renamed this variable so searching `new STS` is more consistent

siddsriv (approved):

trivikr (approved):

PR #6846, opened 2025-01-24
fix(middleware-flexible-checksum): remove stream collection code

kuhe:
- removes code that buffers streaming bodies


packages/middleware-flexible-checksums/src/flexibleChecksumsResponseMiddleware.ts
```diff
@@ -84,22 +81,11 @@ export const flexibleChecksumsResponseMiddleware =
...
-      await validateChecksumFromResponse(result.response as HttpResponse, {
+      await validateChecksumFromResponse(response as HttpResponse, {
         config,
         responseAlgorithms,
         logger: context.logger,
       });
-
-      if (isStreamingBody && collectedStream) {
-        response.body = createReadStreamOnBuffer(collectedStream);
-      }
```
kuhe, line 102: at this point, if the original body is a stream A, the stream has been
1. buffered
2. stream B created from the buffer
3. stream wrapper C (from checksumstream) added over the B stream
4. buffered again
5. stream D created on the second buffer

trivikr (approved):

Issue #6859, opened 2025-01-31
S3Client with UploadPartCommand is not working after updating to the latest SDK version from 3.583.0


ssafayet:
### Checkboxes for prior research

- [x] I've gone through [Developer Guide](https://docs.aws.amazon.com/sdk-for-javascript/v3/developer-guide) and [API reference](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest)
- [x] I've checked [AWS Forums](https://forums.aws.amazon.com) and [StackOverflow](https://stackoverflow.com/questions/tagged/aws-sdk-js).
- [x] I've searched for [previous similar issues](https://github.com/aws/aws-sdk-js-v3/issues) and didn't find any solution.

### Describe the bug

I have updated my SDK js library to 3.738.0 from 3.583.0 and now it is throwing an error after sending it via S3Client using UploadPartCommand. 

Params look like this:

```
new UploadPartCommand({
    Bucket: <BUCKET>
    Key: <KEY>,
    UploadId: <UPLOAD_ID>
    PartNumber: 3,    
    ContentLength: 20992280,
    Body: <FileStream>
})
```

Error:
```
An error was encountered in a non-retryable streaming request.
S3ServiceException [InvalidChunkSizeError]: Only the last chunk is allowed to have a size less than 8192 bytes
```

But this is not the last chunk. It is the number 3 chunk of 11 chunks. This has also failed for subsequent chunks. 

### Regression Issue

- [ ] Select this option if this issue appears to be a regression.

### SDK version number

@aws-sdk/s3-client@3.738.0

### Which JavaScript Runtime is this issue in?

Node.js

### Details of the browser/Node.js/ReactNative version

20.13.1

### Reproduction Steps

Try to create a createMultipartUpload using s3 client and then try to upload FileStream using UploadPartCommand.

### Observed Behavior

This error should occur:

```
An error was encountered in a non-retryable streaming request.
S3ServiceException [InvalidChunkSizeError]: Only the last chunk is allowed to have a size less than 8192 bytes
```

### Expected Behavior

FileStream should be uploaded correctly.

aBurmeseDev:
Hi @ssafayet - thanks for reaching out. 

I'm not able to reproduce this error with reported version on my end. Could you share your code snippet that shows entire Multipart Upload which includes `CreateMultipartUploadCommand, UploadPartCommand, CompleteMultipartUploadCommand, AbortMultipartUploadCommand`? 

[...]

PR #6882, opened 2025-02-13
fix(middleware-flexible-checksums): buffer stream chunks to minimum required size

kuhe:
helps with https://github.com/aws/aws-sdk-js-v3/issues/6859

Attempts to buffer input stream chunks to a minimum required size as required by S3, as described in https://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-streaming.html.

This is for input streams e.g. S3 PUT Object or Part.

With checksumming on, user streams go through a filter called AWS chunked encoding. This process annotates every stream chunk with its size and thus the chunks are sent to S3 as-is in whatever size the user provides them, instead of Node.js being able to automatically buffer smaller chunks.

S3 has always required a chunk size of 8kb, preferably 64kb, but due to automatic buffering this was rarely an issue prior to the introduction of default checksums.

As a mitigation / convenience feature, the JS SDK can automatically buffer streams for the user before undergoing chunked encoding. This incurs a small performance penalty that can be avoided by the user if they provide streams with larger chunks.

- [x] needs codegen update from https://github.com/smithy-lang/smithy-typescript/pull/1523

### DIFF
--- a/packages/middleware-flexible-checksums/src/configuration.ts
+++ b/packages/middleware-flexible-checksums/src/configuration.ts
@@ -66,4 +66,9 @@ export interface PreviouslyResolved {
    * Collects streams into buffers.
    */
   streamCollector: StreamCollector;
+
+  /**
+   * Minimum bytes from a stream to buffer into a chunk before passing to chunked encoding.
+   */
+  requestStreamBufferSize: number;
 }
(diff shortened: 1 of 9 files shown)


kuhe:
- [ ] unit
- [ ] integ
- [ ] e2e
- [ ] cucumber (s3 tag especially)

PR #6929, opened 2025-03-11
fix(cloudfront-signer): parse date using Date constructor

smilkuri:
### Issue
#6403

### Description
This PR adds changes in the parseDate function to use date constructor instead of Date.parse() to support timestamps in milliseconds.

### Testing
Locally by testing it with milliseconds and date format string inputs.

 ``` RUN  v2.1.9 /local/home/smilkuri/aws-sdk-js-v3/packages/cloudfront-signer

 ✓ src/sign.spec.ts (25)
   ✓ getSignedUrl (11)
     ✓ should maintain query params after signing a URL
     ✓ should include url path in policy of signed URL
     ✓ should sign a URL with a canned policy
     ✓ should sign a URL with a custom policy containing a start date
     ✓ should sign a URL with a custom policy containing an ip address
     ✓ should sign a URL with a custom policy containing a start date and ip address
     ✓ should allow an ip address with and without a mask
     ✓ should throw an error when the ip address is invalid
     ✓ should sign a RTMP URL
     ✓ should sign a URL with a policy provided by the user
     ✓ should sign a URL automatically extracted from a policy provided by the user
   ✓ getSignedCookies (8)
     ✓ should allow an ip address with and without a mask
     ✓ should throw an error when the ip address is invalid
     ✓ should be able sign cookies that contain a URL with wildcards
     ✓ should sign cookies with a canned policy
     ✓ should sign cookies with a custom policy containing a start date
     ✓ should sign cookies with a custom policy containing an ip address
     ✓ should sign cookies with a custom policy containing a start date and ip address
     ✓ should sign cookies with a policy provided by the user without a url
   ✓ getSignedUrl- when signing a URL with a date range (6)
     ✓ allows string input compatible with Date constructor
     ✓ allows number input in milliseconds compatible with Date constructor
     ✓ allows Date object input
     ✓ allows string input for date range
     ✓ allows number input for date range
     ✓ allows Date object input for date range

 Test Files  1 passed (1)
      Tests  25 passed (25)
   Start at  14:58:42
   Duration  507ms (transform 133ms, setup 0ms, collect 135ms, tests 96ms, environment 0ms, prepare 87ms)

```

### DIFF
--- a/packages/cloudfront-signer/src/sign.spec.ts
+++ b/packages/cloudfront-signer/src/sign.spec.ts
@@ -607,8 +607,11 @@ describe("getSignedCookies", () => {
 
 describe("getSignedUrl- when signing a URL with a date range", () => {
   const dateString = "2024-05-17T12:30:45.000Z";
+  const dateGreaterThanString = "2024-05-16T12:30:45.000Z";
   const dateNumber = 1125674245900;
+  const dateGreaterThanNumber = 1716034245000;
   const dateObject = new Date(dateString);
+  const dateGreaterThanObject = new Date(dateGreaterThanString);
   it("allows string input compatible with Date constructor", () => {
     const epochDateLessThan = Math.round(new Date(dateString).getTime() / 1000);
     const resultUrl = getSignedUrl({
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: packages/cloudfront-signer/src/sign.spec.ts [hunk 1]
```diff
@@ -7,9 +7,18 @@ import { getSignedCookies, getSignedUrl } from "./index";
 const url = "https://d111111abcdef8.cloudfront.net/private-content/private.jpeg";
 const keyPairId = "APKAEIBAERJR2EXAMPLE";
 const dateLessThan = "2020-01-01";
-const epochDateLessThan = Math.round(new Date(dateLessThan).getTime() / 1000);
+const epochDateLessThan = Math.round(
+  (typeof dateLessThan === "string" && /^\d+$/.test(dateLessThan)
+    ? Number(dateLessThan)
+    : new Date(dateLessThan).getTime()) / 1000
+);
```
kuhe, line 14: this test code change does not test the runtime code change

packages/cloudfront-signer/src/sign.spec.ts
```diff
@@ -7,9 +7,18 @@ import { getSignedCookies, getSignedUrl } from "./index";
...
 const dateGreaterThan = "2019-12-01";
-const epochDateGreaterThan = Math.round(new Date(dateGreaterThan).getTime() / 1000);
+const epochDateGreaterThan = Math.round(
+  (typeof dateGreaterThan === "string" && /^\d+$/.test(dateGreaterThan)
+    ? Number(dateGreaterThan)
+    : new Date(dateGreaterThan).getTime()) / 1000
+);
```
kuhe, line 20: this test change also does not test the runtime change

packages/cloudfront-signer/src/sign.ts
```diff
@@ -363,8 +363,13 @@ class CloudfrontSignBuilder {
     if (!date) {
       return undefined;
     }
-    const parsedDate = Date.parse(date);
-    return isNaN(parsedDate) ? undefined : this.epochTime(new Date(parsedDate));
+    let parsedDate: Date;
```
kuhe, line 366: change the input signature to `string | number | Date`.

Indicate that the field will be passed to the native Date constructor.

emit a warning if the time resolved has a non-zero milliseconds.

kuhe (changes requested):

packages/cloudfront-signer/src/sign.ts
smilkuri: okay wasn't sure if changing type would be a breaking change.

packages/cloudfront-signer/src/sign.ts
```diff
@@ -363,8 +363,13 @@ class CloudfrontSignBuilder {
...
+    if (/^\d+$/.test(date)) {
```
kuhe, line 367: no regexes are needed. Once you update the input signature, add test cases for number and Date inputs, and pass the user value directly to the Date constructor.

packages/cloudfront-signer/src/sign.spec.ts
kuhe: `dateLessThan` is a constant string. The rounding, the ternary, and the Date constructor don't do anything.

 The new test cases should be:

```js
describe('when signing a URL with a date range')
  it('allows string input compatible with Date constructor')
  it('allows number input compatible with Date constructor')
  it('allows Date input')
```

in each case assert that the getSignedUrl and getSignedCookies outputs are as expected

packages/cloudfront-signer/src/sign.spec.ts
```diff
@@ -604,3 +604,49 @@ describe("getSignedCookies", () => {
     expect(verifySignature(denormalizeBase64(result["CloudFront-Signature"]), policy)).toBeTruthy();
   });
 });
+
+describe("getSignedUrl- when signing a URL with a date range", () => {
+  const dateString = "2024-05-17";
```
kuhe, line 609: Use a precise date down to the second in tests. Millisecond number or ISO date string with timezone.

This YYYY-MM-DD date string only works because of the specific JS runtime, and I don't believe is generally spec-compatible with the Date constructor.

packages/cloudfront-signer/src/sign.spec.ts
```diff
@@ -604,3 +604,49 @@ describe("getSignedCookies", () => {
...
+      url,
+      keyPairId,
+      dateLessThan: dateNumber as unknown as string,
+      privateKey,
+      passphrase,
+    });
+
+    expect(resultUrl).toContain(`Expires=${epochDateLessThan}`);
+    expect(resultCookies["CloudFront-Expires"]).toBe(epochDateLessThan);
+  });
```
kuhe, line 651: need another test case for `Date` type input.

packages/cloudfront-signer/src/sign.spec.ts
```diff
@@ -604,3 +604,49 @@ describe("getSignedCookies", () => {
...
+
+describe("getSignedUrl- when signing a URL with a date range", () => {
+  const dateString = "2024-05-17";
+  const dateNumber = 1125674245900;
+  it("allows string input compatible with Date constructor", () => {
+    const epochDateLessThan = Math.round(new Date(dateString).getTime() / 1000);
+    const resultUrl = getSignedUrl({
+      url,
+      keyPairId,
+      dateLessThan: dateString,
```
kuhe, line 616: there are two ends to the date range options, right? this needs to test the other end too.

kuhe (changes requested):

packages/cloudfront-signer/src/sign.ts
```diff
@@ -359,12 +359,12 @@ class CloudfrontSignBuilder {
     return Math.round(date.getTime() / 1000);
   }
 
-  private parseDate(date?: string): number | undefined {
+  private parseDate(date?: string | number | Date): number | undefined {
```
kuhe, line 362: change the public signature of of dateLessThan and dateGreaterThan

packages/cloudfront-signer/src/sign.ts
```diff
@@ -359,15 +359,15 @@ class CloudfrontSignBuilder {
...
-    return isNaN(parsedDate) ? undefined : this.epochTime(new Date(parsedDate));
+    const parsedDate = new Date(date);
+    return isNaN(parsedDate.getTime()) ? undefined : this.epochTime(parsedDate);
   }
 
-  private parseDateWindow(expiration: string, start?: string): PolicyDates {
+  private parseDateWindow(expiration: string | number | Date, start?: string | number | Date): PolicyDates {
     const dateLessThan = this.parseDate(expiration);
     if (!dateLessThan) {
       throw new Error("dateLessThan is invalid. Ensure the date string is compatible with the Date constructor.");
```
siddsriv, line 373: nit: should we change this message to "....Ensure the date value is compatible with the Date constructor." instead of the "date string"

siddsriv (approved):

Issue #6937, opened 2025-03-15
`middleware-recursion-detection` logic is not skipped when "X-Amzn-Trace-Id" header is set in lowercase


jj22ee:
### Checkboxes for prior research

- [x] I've gone through [Developer Guide](https://docs.aws.amazon.com/sdk-for-javascript/v3/developer-guide) and [API reference](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest)
- [x] I've checked [AWS Forums](https://forums.aws.amazon.com) and [StackOverflow](https://stackoverflow.com/questions/tagged/aws-sdk-js).
- [x] I've searched for [previous similar issues](https://github.com/aws/aws-sdk-js-v3/issues) and didn't find any solution.

### Describe the bug

[X-Ray SDK propagates](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/propagator-aws-xray-v1.26.2/propagators/propagator-aws-xray/src/AWSXRayPropagator.ts#L33) X-Ray Trace Header like: `x-amzn-trace-id`

[OTel SDK propagates](https://github.com/aws/aws-xray-sdk-node/blob/aws-xray-sdk-node%403.10.3/packages/core/lib/patchers/aws3_p.ts#L144) X-Ray Trace Header like: `X-Amzn-Trace-Id`

The [current logic in `middleware-recursion-detection`](https://github.com/aws/aws-sdk-js-v3/blob/v3.768.0/packages/middleware-recursion-detection/src/index.ts#L28-L48) intends to inject trace ID to request header to detect recursion invocation in Lambda.
- This [logic is skipped when](https://github.com/aws/aws-sdk-js-v3/blob/v3.768.0/packages/middleware-recursion-detection/src/index.ts#L33) `X-Amzn-Trace-Id` is set in the request headers.
- This logic is **not** skipped when `x-amzn-trace-id` is set in the request headers, but it should, given that header checks should be case-insensitive

### Regression Issue

- [ ] Select this option if this issue appears to be a regression.

### SDK version number

@aws-sdk/middleware-recursion-detection@3.734.0

### Which JavaScript Runtime is this issue in?

Node.js

### Details of the browser/Node.js/ReactNative version

v20.10.0

### Reproduction Steps

1. Have LambdaX and LambdaY, LambdaX invokes LambdaY via Lambda `invoke` from `@aws-sdk/client-lambda`, add permissions to LambdaX to use `invoke`.
[...]

PR #6938, opened 2025-03-15
fix(middleware-recursion-detection): make the x-ray trace header check case-insensitive

jj22ee:
### Issue
https://github.com/aws/aws-sdk-js-v3/issues/6937

### Description

`middleware-recursion-detection` performs a request header check that is case-sensitive.
This PR changes this check to be case-insensitive.

### Testing

Added unit tests.

Also:
1. Have LambdaX and LambdaY, LambdaX invokes LambdaY via Lambda invoke from @aws-sdk/client-lambda, add permissions to LambdaX to use invoke.
2. Modified `middleware-recursion-detection` in node_modules of the LambdaX to make the trace header check case-insensitive
3. Enable Lambda service traces in both LambdaX and LambdaY
4. Enable Application Signals in LambdaX both LambdaX and LambdaY
5. Invoke LambdaX function.
6. Observed that `middleware-recursion-detection` will correctly not inject the trace header when calling `lambdaClient's invoke`, because the `x-amzn-trace-id` header already exists due to step (4)

### Additional context

### Checklist

---

### DIFF
--- a/packages/middleware-recursion-detection/src/index.spec.ts
+++ b/packages/middleware-recursion-detection/src/index.spec.ts
@@ -70,6 +70,27 @@ describe(recursionDetectionMiddleware.name, () => {
     expect(request.headers[TRACE_ID_HEADER_NAME]).toBe("some-real-trace-id");
   });
 
+  it(`should NOT set ${TRACE_ID_HEADER_NAME} header when the header is already set in all lowercase`, async () => {
+    process.env = {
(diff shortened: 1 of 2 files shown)


jj22ee:
This solution introduces looping through all the keys in the Request Headers, the alternative solution would be to check for both `X-Amzn-Trace-Id` and `x-amzn-trace-id` headers.

PR #6941, opened 2025-03-17
docs: how to create a layer in AWS Lambda

kuhe:
add information about creating lambda layers to our AWS Lambda supplemental docs

### DIFF
--- a/supplemental-docs/AWS_LAMBDA.md
+++ b/supplemental-docs/AWS_LAMBDA.md
@@ -76,10 +76,7 @@ await lambda.publishLayerVersion({
     ZipFile: fs.readFileSync("./path/to/layer_content.zip"),
   },
   CompatibleRuntimes: ["nodejs18.x", "nodejs20.x", "nodejs22.x"],
-  CompatibleArchitectures: [
-    "x86_64",
-    "arm64",
-  ],
+  CompatibleArchitectures: ["x86_64", "arm64"],
 });
 ```
 
--- a/supplemental-docs/AWS_LAMBDA.md
+++ b/supplemental-docs/AWS_LAMBDA.md
@@ -75,6 +75,8 @@ await lambda.publishLayerVersion({
     // read the layer zip file as a buffer.
     ZipFile: fs.readFileSync("./path/to/layer_content.zip"),
   },
+  // Confirm from SDK package.json engines field that version supports each runtime.
+  // It usually does, if Lambda hasn't blocked function create for the runtime.
   CompatibleRuntimes: ["nodejs18.x", "nodejs20.x", "nodejs22.x"],
   CompatibleArchitectures: ["x86_64", "arm64"],
 });
--- a/supplemental-docs/AWS_LAMBDA.md
+++ b/supplemental-docs/AWS_LAMBDA.md
@@ -67,7 +67,7 @@ import fs from "node:fs";
 const lambda = new Lambda();
 
 await lambda.publishLayerVersion({
-  LayerName: "MyAwsSdkDependencies-layer",
+  LayerName: "AWS-SDK-JavaScript-v3.750.0-layer",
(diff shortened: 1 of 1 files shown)


supplemental-docs/AWS_LAMBDA.md
```diff
@@ -19,6 +20,112 @@ exports.handler = function (event) {
...
+  Description: "Contains AWS SDK dependencies to use in my AWS Lambda functions",
+  Content: {
+    // read the layer zip file as a buffer.
+    ZipFile: fs.readFileSync("./path/to/layer_content.zip"),
+  },
+  CompatibleRuntimes: ["nodejs18.x", "nodejs20.x", "nodejs22.x"],
+  CompatibleArchitectures: [
+    "x86_64",
+    "arm64",
+  ],
```
trivikr, line 82: nit: single line just as previous one
```suggestion
  CompatibleArchitectures: ["x86_64", "arm64"],
```

supplemental-docs/AWS_LAMBDA.md
```diff
@@ -19,6 +20,112 @@ exports.handler = function (event) {
...
+});
+
+await lambda.publishLayerVersion({
+  LayerName: "MyAwsSdkDependencies-layer",
+  Description: "Contains AWS SDK dependencies to use in my AWS Lambda functions",
+  Content: {
+    // read the layer zip file as a buffer.
+    ZipFile: fs.readFileSync("./path/to/layer_content.zip"),
+  },
+  CompatibleRuntimes: ["nodejs18.x", "nodejs20.x", "nodejs22.x"],
```
trivikr, line 78: nit: add comment about SDK version support in compatible runtime
```suggestion
  // Confirm from SDK package.json engines field that version supports each runtime.
  // It usually does, if Lambda hasn't blocked function create for the runtime.
  CompatibleRuntimes: ["nodejs18.x", "nodejs20.x", "nodejs22.x"],
```

supplemental-docs/AWS_LAMBDA.md
```diff
@@ -19,6 +20,112 @@ exports.handler = function (event) {
...
+import { Lambda } from "@aws-sdk/client-lambda";
+import fs from "node:fs";
+
+const lambda = new Lambda({
+  region: "us-west-2",
+});
+
+await lambda.publishLayerVersion({
+  LayerName: "MyAwsSdkDependencies-layer",
+  Description: "Contains AWS SDK dependencies to use in my AWS Lambda functions",
```
trivikr, line 73: Should we suggest putting SDK version number in the layer name, in addition to project?
```suggestion
  LayerName: "<project>-nodejs-sdk-v<version>-layer",
  Description: "Contains AWS SDK v<version> dependencies to use in <project>",
```
Lambda layers has it's own versioning, but it's independent of contents. It's better to have version in layer name for easy discoverability.

supplemental-docs/AWS_LAMBDA.md
```diff
@@ -19,6 +20,112 @@ exports.handler = function (event) {
...
+### Deploying the AWS Lambda layer to your account
+
+```js
+// example: deploying the AWS Lambda layer.
+import { Lambda } from "@aws-sdk/client-lambda";
+import fs from "node:fs";
+
+const lambda = new Lambda({
+  region: "us-west-2",
+});
```
trivikr, line 69: Should we remove the region from the examples altogether to keep the code simple?
```suggestion
const lambda = new Lambda();
```
Ditto on line 95.

supplemental-docs/AWS_LAMBDA.md
kuhe: Added version used above, which was 3.750.0. 

I think fewer variables would be easier to understand for the reader. The layer doesn't really need to be specific to a project.

trivikr (approved):

PR #6959, opened 2025-03-21
fix(packages): maintain singleton config object

kuhe:
Related to https://github.com/smithy-lang/smithy-typescript/pull/1550

This change makes all config resolvers mutate the input config instead of creating a shallow copy. The shallow copies, if preserved by references, can result in reading inconsistent config data.

### DIFF
--- a/packages/middleware-endpoint-discovery/src/resolveEndpointDiscoveryConfig.ts
+++ b/packages/middleware-endpoint-discovery/src/resolveEndpointDiscoveryConfig.ts
@@ -68,13 +68,16 @@ export interface EndpointDiscoveryConfigOptions {
 export const resolveEndpointDiscoveryConfig = <T>(
   input: T & PreviouslyResolved & EndpointDiscoveryInputConfig,
   { endpointDiscoveryCommandCtor }: EndpointDiscoveryConfigOptions
-): T & EndpointDiscoveryResolvedConfig => ({
-  ...input,
-  endpointDiscoveryCommandCtor,
-  endpointCache: new EndpointCache(input.endpointCacheSize ?? 1000),
-  endpointDiscoveryEnabled:
-    input.endpointDiscoveryEnabled !== undefined
-      ? () => Promise.resolve(input.endpointDiscoveryEnabled)
-      : input.endpointDiscoveryEnabledProvider,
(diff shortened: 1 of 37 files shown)


packages/middleware-endpoint-discovery/src/resolveEndpointDiscoveryConfig.ts
```diff
@@ -62,13 +62,16 @@ export interface EndpointDiscoveryConfigOptions {
...
+  const { endpointCacheSize, endpointDiscoveryEnabled, endpointDiscoveryEnabledProvider } = input;
+
+  return Object.assign(input, {
+    endpointDiscoveryCommandCtor,
+    endpointCache: new EndpointCache(endpointCacheSize ?? 1000),
+    endpointDiscoveryEnabled:
+      endpointDiscoveryEnabled !== undefined
+        ? () => Promise.resolve(endpointDiscoveryEnabled)
+        : endpointDiscoveryEnabledProvider,
+    isClientEndpointDiscoveryEnabled: endpointDiscoveryEnabled !== undefined,
```
kuhe, line 75: this boolean seems wrong... what if `endpointDiscoveryEnabled === false`? 

not related to this PR though

packages/middleware-token/src/resolveTokenConfig.ts
```diff
@@ -5,7 +5,9 @@ import { tokenDefaultProvider } from "./tokenDefaultProvider";
...
  * @internal
  */
-export const resolveTokenConfig = <T>(input: T & TokenInputConfig): T & TokenResolvedConfig => ({
-  ...input,
-  token: input.token ? normalizeTokenProvider(input.token) : tokenDefaultProvider(input as any),
-});
+export const resolveTokenConfig = <T>(input: T & TokenInputConfig): T & TokenResolvedConfig => {
+  const { token } = input;
+  return Object.assign(input, {
+    token: token ? normalizeTokenProvider(token) : tokenDefaultProvider(input as any),
```
kuhe, line 11: this fallback makes no sense to me since `input` has no overlap with `FromSsoInit`, the type of the falsy branch.

trivikr (approved):

Issue #6960, opened 2025-03-22
Regression in credential provider force refresh starting in 3.667, now affects lambda runners in us-east-1


RossWilliams:
### Checkboxes for prior research

- [x] I've gone through [Developer Guide](https://docs.aws.amazon.com/sdk-for-javascript/v3/developer-guide) and [API reference](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest)
- [x] I've checked [AWS Forums](https://forums.aws.amazon.com) and [StackOverflow](https://stackoverflow.com/questions/tagged/aws-sdk-js).
- [x] I've searched for [previous similar issues](https://github.com/aws/aws-sdk-js-v3/issues) and didn't find any solution.

### Describe the bug

A change in 3.667, PR #6546 , accidentally changed behaviour of the credential provider. This change is now in the lambda runner in US-East-1, causing production issues for users not bundling the SDK, or for users that bundle version 3.667 or later.

[This line](https://github.com/aws/aws-sdk-js-v3/commit/089f1a42e58c7a85c0720c83ae5354e75a253695#diff-d1233147c05cab2966b899bc90ffb9dbb277713b5180b50aebbd84b892faa27bR227) invokes the customer-supplied credential provider, but does not pass through the credential arguments, causing any arguments supplied by the user to be dropped. 

A fix for the issue is to pass through the parameters. 

A specific parameter that gets dropped in my case is the `forceRefresh` parameter, which is used in a more complex scenario to provide tenant-scoped credentials.

### Regression Issue

- [x] Select this option if this issue appears to be a regression.

### SDK version number

@aws-sdk/core@3.667

### Which JavaScript Runtime is this issue in?

Node.js

### Details of the browser/Node.js/ReactNative version

v22.13.0

### Reproduction Steps

Minimal reproduction which passes in 3.666 and fails in 3.667 (and later)

```javascript
import { DynamoDBClient, ListTablesCommand } from "@aws-sdk/client-dynamodb";
import { assert } from "console";

let client = null;
async function getClient() {
  if (!client) {
    client = new DynamoDBClient({
      credentials: credentialsProvider(),
      region: "us-east-1",
    });
  }
  await client.config.credentials({ forceRefresh: true });
  return client;
}

let counter = 0;
function credentialsProvider() {
  return async function () {
    counter += 1;
    console.log("Getting credentials");
    return {
      accessKeyId: process.env.AWS_ACCESS_KEY_ID,
[...]

PR #6961, opened 2025-03-22
fix: pass through credentials parameters when wrapping the provider

RossWilliams:
### Issue
#6960

### Description

Passes through credentials parameter object when wrapping the credential provider

### Testing
Minimal reproduction example in linked issue

### Additional context
Fixes inadvertent behaviour change. Issue recently rolled out into bundled Lambda runtime.

### Checklist

---


RossWilliams:
The types in [Smithy-lang/smithy-typescript here](https://github.com/smithy-lang/smithy-typescript/blob/234b491f667101e08fab524b7d88c67bfed64c46/packages/types/src/util.ts#L103) could also be tightened to require the first parameter so that it does not get accidentally invoked without passing it in. This is probably important in the aws-sdk-js-v3 codebase that does a lot of layering.

kuhe (approved):
This looks right as a fix.

trivikr (approved):

PR #6964, opened 2025-03-24
docs(middleware-flexible-checksums): add api annotations

kuhe:
adds missing api extractor configs and api annotation to the flexible checksums package

### DIFF
--- a/packages/middleware-flexible-checksums/api-extractor.json
+++ b/packages/middleware-flexible-checksums/api-extractor.json
@@ -1,0 +1,4 @@
+{
+  "extends": "../../api-extractor.packages.json",
+  "mainEntryPointFilePath": "./dist-types/index.d.ts"
+}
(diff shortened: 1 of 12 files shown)


packages/middleware-host-header/package.json
```diff
@@ -9,6 +9,7 @@
     "build:types": "tsc -p tsconfig.types.json",
     "build:types:downlevel": "downlevel-dts dist-types dist-types/ts3.4",
     "clean": "rimraf ./dist-* && rimraf *.tsbuildinfo",
+    "extract:docs": "api-extractor run --local",
```
trivikr, line 12: Should this diff be part of a different PR?
```suggestion
```

trivikr: Ditto for middleware-endpoint-discovery

trivikr (approved):

PR #6970, opened 2025-03-25
test(aws-client-api-test): assert client config custody in integ test

kuhe:
asserts config object custody for a selection of AWS SDK clients that represent the majority if not entirety of our config resolvers


private/aws-client-api-test/src/client-interface-tests/config/client-config-custody.spec.ts
```diff
@@ -0,0 +1,44 @@
+import { describe, test as it, expect } from "vitest";
```
kuhe, line 1: the test is run in the unit test group but is an integration test

trivikr (approved):

trivikr (approved):

PR #7038, opened 2025-04-28
chore(clients): populate default values for auth scheme preference

trivikr:
### Issue
Internal JS-5811

### Description
Populates default values for auth scheme preference

### Testing
CI

---

### DIFF
--- a/clients/client-accessanalyzer/src/runtimeConfig.ts
+++ b/clients/client-accessanalyzer/src/runtimeConfig.ts
@@ -2,7 +2,7 @@
 // @ts-ignore: package.json will be imported from dist folders
 import packageInfo from "../package.json"; // eslint-disable-line
 
-import { emitWarningIfUnsupportedVersion as awsCheckVersion } from "@aws-sdk/core";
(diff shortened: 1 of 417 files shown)


clients/client-accessanalyzer/src/runtimeConfig.ts
```diff
@@ -2,7 +2,7 @@
 // @ts-ignore: package.json will be imported from dist folders
 import packageInfo from "../package.json"; // eslint-disable-line
 
-import { emitWarningIfUnsupportedVersion as awsCheckVersion } from "@aws-sdk/core";
+import { NODE_AUTH_SCHEME_PREFERENCE_OPTIONS, emitWarningIfUnsupportedVersion as awsCheckVersion } from "@aws-sdk/core";
```
kuhe, line 5: NODE_AUTH_SCHEME_PREFERENCE_OPTIONS should be imported from `@aws-sdk/core/httpAuthSchemes` to eventually move away from importing anything from the root index of core.

PR #7056, opened 2025-05-06
chore(verdaccio): enable offline publishing for packages

smilkuri:
### Issue
Internal JS-5879

### Description
This PR adds the `offline_publish: true` setting to the configuration in the Verdaccio config.yaml file. This change allows publishing the packages even when the registry is offline or when the npmjs uplink is not available.

### Testing
How was this change tested?
CI

---


kuhe:
- [ ] local benchmark script run (uses verdaccio)
- [ ] perform dry run

kuhe (approved):

PR #7067, opened 2025-05-08
chore(core): return httpBearerAuth in authSchemePreference if token is set in env

trivikr:
### Issue
Internal JS-5860

### Description
Returns httpBearerAuth in authSchemePreference if token is set in env

### Testing
Unit tests

---

### DIFF
--- a/packages/core/src/submodules/httpAuthSchemes/aws_sdk/NODE_AUTH_SCHEME_PREFERENCE_OPTIONS.ts
+++ b/packages/core/src/submodules/httpAuthSchemes/aws_sdk/NODE_AUTH_SCHEME_PREFERENCE_OPTIONS.ts
@@ -1,6 +1,4 @@
-// ToDo: replace with EnvOptions once exported from node-config-provider
-import { LoadedConfigSelectors } from "@smithy/node-config-provider";
-import { EnvOptions } from "@smithy/node-config-provider/dist-types/fromEnv";
+import { EnvOptions, LoadedConfigSelectors } from "@smithy/node-config-provider";
 
 import { getArrayForCommaSeparatedString } from "../utils/getArrayForCommaSeparatedString";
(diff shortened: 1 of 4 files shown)


packages/core/src/submodules/httpAuthSchemes/aws_sdk/NODE_AUTH_SCHEME_PREFERENCE_OPTIONS.ts
```diff
@@ -14,7 +17,11 @@ export const NODE_AUTH_SCHEME_PREFERENCE_OPTIONS: LoadedConfigSelectors<string[]
    * @param env - Node process environment object
    * @returns Array of auth scheme strings if preference is set, undefined otherwise
    */
-  environmentVariableSelector: (env: NodeJS.ProcessEnv) => {
+  environmentVariableSelector: (env: NodeJS.ProcessEnv, options?: EnvOptions) => {
+    if (options?.signingName) {
+      const bearerTokenKey = getBearerTokenEnvKey(options.signingName);
+      if (bearerTokenKey in env) return ["httpBearerAuth"];
```
kuhe, line 23: should this only activate if `!(NODE_AUTH_SCHEME_PREFERENCE_ENV_KEY in env)` is true?

kuhe (commented):
lgtm

packages/core/src/submodules/httpAuthSchemes/aws_sdk/NODE_AUTH_SCHEME_PREFERENCE_OPTIONS.ts
trivikr: No. Confirmed internally that token is more specific

kuhe (approved):

PR #7085, opened 2025-05-15
fix(core/protocols): read error code case-insensitively

kuhe:
Our current awsQueryCompatibility implementations reads the `x-amzn-query-error` response header and places this value in the error object's `Code` field.

However, our error parsers read from the `code` field, missing that data.

For the end users, this can cause the thrown error to have a generic "ServiceException" type rather than a more specific type, but it will contain the same data fields. This affects clients for services running in awsQueryCompatibility mode (currently only SQS).

related: https://github.com/smithy-lang/smithy-typescript/pull/1597

### DIFF
--- a/codegen/smithy-aws-typescript-codegen/src/main/java/software/amazon/smithy/aws/typescript/codegen/AwsSmithyRpcV2Cbor.java
+++ b/codegen/smithy-aws-typescript-codegen/src/main/java/software/amazon/smithy/aws/typescript/codegen/AwsSmithyRpcV2Cbor.java
@@ -6,6 +6,7 @@
 package software.amazon.smithy.aws.typescript.codegen;
 
 import software.amazon.smithy.aws.traits.protocols.AwsQueryCompatibleTrait;
+import software.amazon.smithy.typescript.codegen.SmithyCoreSubmodules;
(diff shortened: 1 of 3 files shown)


codegen/smithy-aws-typescript-codegen/src/main/java/software/amazon/smithy/aws/typescript/codegen/AwsSmithyRpcV2Cbor.java
```diff
@@ -55,15 +56,18 @@ protected void writeSharedRequestHeaders(ProtocolGenerator.GenerationContext con
 
     @Override
     protected void writeErrorCodeParser(GenerationContext generationContext) {
-        super.writeErrorCodeParser(generationContext);
```
kuhe, line 58: caller super here put the lines in the wrong order previously

kuhe: there's no codegen diff for this since the protocol test for cbor does not run with awsQueryCompat on

kuhe:
moved Java CI issues to https://github.com/aws/aws-sdk-js-v3/pull/7086

siddsriv (approved):

PR #7092, opened 2025-05-19
chore(client-privatenetworks): remove privatenetworks service

siddsriv:
### Issue
Internal JS-5850 

### Description
removes `privatenetworks` model and client

---


trivikr (changes requested):
Lockfile needs to be updated

trivikr (approved):

PR #7097, opened 2025-05-21
chore(token-providers): add provider to retrieve token from env var

trivikr:
### Issue
Internal JS-5860

### Description
Adds provider to retrieve token from env var

### Testing
Unit testing

### Checklist

---

### DIFF
--- a/packages/token-providers/src/fromEnvBearerToken.ts
+++ b/packages/token-providers/src/fromEnvBearerToken.ts
@@ -1,0 +1,34 @@
+import { getBearerTokenEnvKey } from "@aws-sdk/core";
+import type { CredentialProviderOptions, TokenIdentityProvider } from "@aws-sdk/types";
+import { TokenProviderError } from "@smithy/property-provider";
+
+export interface FromEnvBearerTokenInit extends CredentialProviderOptions {
+  signingName?: string;
+}
+
+/**
+ * Creates a TokenIdentityProvider that retrieves bearer token from environment variable
+ *
+ * @param options - Configuration options for the token provider
+ * @param options.logger - Optional logger for debug messages
+ * @param options.signingName - Service signing name used to determine environment variable key
+ * @returns TokenIdentityProvider that provides bearer token from environment variable
+ *
+ * @public
+ */
+export const fromEnvBearerToken =
+  ({ logger, signingName }: FromEnvBearerTokenInit = {}): TokenIdentityProvider =>
+  async () => {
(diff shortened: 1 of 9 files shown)


packages/token-providers/src/fromEnvBearerToken.ts
```diff
@@ -0,0 +1,34 @@
+import { getBearerTokenEnvKey } from "@aws-sdk/core";
+import type { CredentialProviderOptions, TokenIdentityProvider } from "@aws-sdk/types";
+import { TokenProviderError } from "@smithy/property-provider";
+
+export interface FromEnvBearerTokenInit extends CredentialProviderOptions {
```
kuhe, line 5:
```suggestion
/**
 * @public
 */
export interface FromEnvBearerTokenInit extends CredentialProviderOptions {
```

packages/token-providers/src/fromEnvBearerToken.ts
```diff
@@ -0,0 +1,34 @@
...
+ * @param options.logger - Optional logger for debug messages
+ * @param options.signingName - Service signing name used to determine environment variable key
+ * @returns TokenIdentityProvider that provides bearer token from environment variable
+ *
+ * @public
+ */
+export const fromEnvBearerToken =
+  ({ logger, signingName }: FromEnvBearerTokenInit = {}): TokenIdentityProvider =>
+  async () => {
+    logger?.debug("@aws-sdk/token-providers - fromEnvBearerToken");
```
kuhe, line 22:
```suggestion
    logger?.debug?.("@aws-sdk/token-providers - fromEnvBearerToken");
```

kuhe: This came up before, where JS users would not provide a full logger implementation and run into errors.

packages/token-providers/src/fromEnvBearerToken.ts
```diff
@@ -0,0 +1,34 @@
...
+ * Creates a TokenIdentityProvider that retrieves bearer token from environment variable
+ *
+ * @param options - Configuration options for the token provider
+ * @param options.logger - Optional logger for debug messages
+ * @param options.signingName - Service signing name used to determine environment variable key
+ * @returns TokenIdentityProvider that provides bearer token from environment variable
+ *
+ * @public
+ */
+export const fromEnvBearerToken =
```
kuhe, line 19: should this appear in a readme either in `token-providers` or `credential-providers` package? 

should credential-providers re-export it?

kuhe (approved):

PR #7099, opened 2025-05-23
chore(codegen): use block for token provider call

trivikr:
### Issue
Avoid conflicts with JS-5860

### Description
Uses block for token provider call

### Testing
N/A, as it's formatting change

---

### DIFF
--- a/codegen/smithy-aws-typescript-codegen/src/main/java/software/amazon/smithy/aws/typescript/codegen/auth/http/integration/AwsSdkCustomizeHttpBearerTokenAuth.java
+++ b/codegen/smithy-aws-typescript-codegen/src/main/java/software/amazon/smithy/aws/typescript/codegen/auth/http/integration/AwsSdkCustomizeHttpBearerTokenAuth.java
@@ -55,7 +55,9 @@ public final class AwsSdkCustomizeHttpBearerTokenAuth implements HttpAuthTypeScr
                     .addDependency(AwsDependency.TOKEN_PROVIDERS)
                     .addImport("nodeProvider", null, AwsDependency.TOKEN_PROVIDERS)
                     .addImport("FromSsoInit", null, AwsDependency.TOKEN_PROVIDERS)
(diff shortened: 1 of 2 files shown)


siddsriv (approved):

codegen/smithy-aws-typescript-codegen/src/main/java/software/amazon/smithy/aws/typescript/codegen/auth/http/integration/AwsSdkCustomizeHttpBearerTokenAuth.java
```diff
@@ -55,7 +55,9 @@ public void customizeSupportedHttpAuthSchemes(
                     .addDependency(AwsDependency.TOKEN_PROVIDERS)
                     .addImport("nodeProvider", null, AwsDependency.TOKEN_PROVIDERS)
                     .addImport("FromSsoInit", null, AwsDependency.TOKEN_PROVIDERS)
-                    .write("async (idProps) => await nodeProvider(idProps as FromSsoInit)(idProps)"))
+                    .openBlock("async (idProps) => {", "}", () -> {
+                        w.write("return await nodeProvider(idProps as FromSsoInit)(idProps);");
```
kuhe, line 59: async function can return without await

kuhe (approved):

codegen/smithy-aws-typescript-codegen/src/main/java/software/amazon/smithy/aws/typescript/codegen/auth/http/integration/AwsSdkCustomizeHttpBearerTokenAuth.java
trivikr: It can, but we want errors to be thrown inside the function block, for changes coming up in future.
Not specifically for this line, but when we'll add a try block in future.

If it returns without await the error will be thrown where it's awaited by callee.

PR #7126, opened 2025-06-11
chore(middleware-http-debug-log): add middleware timer

kuhe:
This adds a plugin for measuring the time spent in individual middleware functions.

The package is private and not released to NPM.

### DIFF
--- a/packages/middleware-http-debug-log/src/getMiddlewareTimerPlugin.ts
+++ b/packages/middleware-http-debug-log/src/getMiddlewareTimerPlugin.ts
@@ -1,0 +1,112 @@
+import {
+  HandlerExecutionContext,
+  InitializeHandler,
+  InitializeHandlerArguments,
+  MiddlewareStack,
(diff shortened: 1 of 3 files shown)


packages/middleware-http-debug-log/src/getMiddlewareTimerPlugin.ts
```diff
@@ -0,0 +1,112 @@
...
+  SerializeHandler,
+  SerializeHandlerArguments,
+} from "@smithy/types";
+
+/**
+ * FOR DEBUG USE ONLY.
+ * This plugin logs or records the timings of the middleware named in options.middlewareNames.
+ *
+ * @param options.middlewareNames - names of the middleware to measure.
+ * @param options.storage - a container for the data. If non is provided, metrics will be logged to console.
```
kuhe:
```suggestion
 * @param options.storage - a container for the data. If none is provided, metrics will be logged to console.
```

PR #7139, opened 2025-06-19
chore: install biomejs cli

kuhe:
### Issue

### Description

### Testing
How was this change tested?

### Additional context

### Checklist

---


biome.json
```diff
@@ -0,0 +1,79 @@
+{
+  "$schema": "https://biomejs.dev/schemas/2.0.0/schema.json",
+  "files": {
+    "maxSize": 16000000,
+    "includes": ["{packages,lib,scripts,tests}/**/*.{mjs,js,ts,json}", "!**/dist-{cjs,es,types}/**", "!**/ruleset.ts"]
+  },
+  "formatter": {
```
kuhe, line 7: I've turned many rules off initially because we need to establish that biome is runnable in our build systems before reintroducing rules that would lead to code changes.

clients/biome.json
```diff
@@ -0,0 +1,26 @@
...
+        "noInvalidBuiltinInstantiation": "error",
+        "noSwitchDeclarations": "off",
+        "noUnusedFunctionParameters": "off"
+      },
+      "style": "off",
+      "nursery": "off",
+      "suspicious": "off"
+    }
+  }
+}
```
kuhe, line 26: because biome requires that subconfiguration must also be named `biome.json` rather than allowing a pointer to a custom config file, and that it must sit in the folder it has control over, I had to edit several scripts that scan the clients folder to exclude this file.

package.json
```diff
@@ -62,6 +62,9 @@
   },
   "license": "UNLICENSED",
   "devDependencies": {
+    "@biomejs/biome": "^2.0.0",
+    "@biomejs/cli-darwin-arm64": "^2.0.0",
+    "@biomejs/cli-linux-x64-musl": "^2.0.0",
```
trivikr: Biome recommends pinning the dependencies
```suggestion
    "@biomejs/biome": "2.0.0",
    "@biomejs/cli-darwin-arm64": "2.0.0",
    "@biomejs/cli-linux-x64-musl": "2.0.0",
```

trivikr (changes requested):

package.json
trivikr: https://biomejs.dev/guides/getting-started/

PR #7144, opened 2025-06-23
chore: install typescript native preview

kuhe:
- use the TS native preview to run the test source code type-check, which is non-critical

```
time npx tsc -p tsconfig.test.json
       12.85 real        18.63 user         1.85 sys
time npx tsgo -p tsconfig.test.json
        5.50 real         7.35 user         1.84 sys
```


kuhe:
we'll use it when it's ready

PR #7160, opened 2025-06-27
chore(core/protocols): fix for XML shape deser

kuhe:
fix for schema XML deser


packages/core/src/submodules/protocols/xml/XmlShapeDeserializer.ts
```diff
@@ -59,11 +59,10 @@ export class XmlShapeDeserializer extends SerdeContextConfig implements ShapeDes
   public readSchema(_schema: Schema, value: any): any {
     const ns = NormalizedSchema.of(_schema);
     const traits = ns.getMergedTraits();
-    const schema = ns.getSchema();
 
     if (ns.isListSchema() && !Array.isArray(value)) {
       // single item in what should have been a list.
-      return this.readSchema(schema, [value]);
+      return this.readSchema(ns, [value]);
```
kuhe, line 65: this is the change: pass NormalizedSchema instead of a raw schema object to the recursive call. 

Passing a raw Schema loses the member context.

trivikr (approved):

PR #7209, opened 2025-07-16
test(util-endpoints): add AWS_ENDPOINT_URL integ test

kuhe:
Adds integ test for the AWS_ENDPOINT_URL feature.

### DIFF
--- a/packages/util-endpoints/src/env-endpoint.integ.spec.ts
+++ b/packages/util-endpoints/src/env-endpoint.integ.spec.ts
@@ -1,0 +1,90 @@
+import { DynamoDB } from "@aws-sdk/client-dynamodb";
+import { S3 } from "@aws-sdk/client-s3";
+import { afterAll, beforeEach, describe, test as it } from "vitest";
+
+import { requireRequestsFrom } from "../../../private/aws-util-test/src";
+
+/**
+ * This is an AWS-specific integration test for a piece of functionality that is contained
+ * within `@smithy/middleware-endpoint` (another repository).
+ *
+ * It augments unit tests in that package.
+ */
+describe("environment variable custom endpoint", () => {
+  beforeEach(async () => {
+    delete process.env.AWS_ENDPOINT_URL;
+    delete process.env.AWS_ENDPOINT_URL_DYNAMODB;
+  });
+
+  afterAll(async () => {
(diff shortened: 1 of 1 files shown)


packages/util-endpoints/src/env-endpoint.integ.spec.ts
```diff
@@ -0,0 +1,90 @@
...
+    const s3 = new S3({
+      region: "us-west-2",
+    });
+    const ddb = new DynamoDB({
+      region: "us-west-2",
+    });
+
+    requireRequestsFrom(s3).toMatch({
+      hostname: /s3\.us-west-2\.amazonaws\.com/,
+    });
```
trivikr, line 81: Do `not.toMatch`, as we don't want this test to fail if computation for default endpoint changes in future.
```suggestion
    requireRequestsFrom(s3).not.toMatch({
      hostname: /localhost/,
      path: /\/1/,
    });
```

packages/util-endpoints/src/env-endpoint.integ.spec.ts
```diff
@@ -0,0 +1,90 @@
...
+describe("environment variable custom endpoint", () => {
+  beforeEach(async () => {
+    delete process.env.AWS_ENDPOINT_URL;
+    delete process.env.AWS_ENDPOINT_URL_DYNAMODB;
+  });
+
+  afterAll(async () => {
+    delete process.env.AWS_ENDPOINT_URL;
+    delete process.env.AWS_ENDPOINT_URL_DYNAMODB;
+  });
```
trivikr, line 22: afterAll is not needed
```suggestion
```

packages/util-endpoints/src/env-endpoint.integ.spec.ts
trivikr: There's no negative matcher

https://github.com/aws/aws-sdk-js-v3/blob/main/private/aws-util-test/src/requests/test-http-handler.ts

packages/util-endpoints/src/env-endpoint.integ.spec.ts
trivikr: Verified that afterAll is needed to remove values set by some tests

trivikr (approved):

PR #7211, opened 2025-07-17
feat(lib-storage): skip uploading parts if they are already listed in uploadedParts.

joy-yu:
Skip uploading parts if they are already listed in uploadedParts.

### Description

In our biz scenario, we record `uploadId` `uploadedParts` `bytesUploadedSoFar` params to preserve information about uploaded parts. When an unexpected interruption occurs and the upload is resumed, we overwrite the previously recorded params. However, during the traversal of parts, the already recorded parts are not currently being skipped.

Therefore, adding this check will have a positive impact on the resume upload functionality.

### Checklist

---

### DIFF
--- a/lib/lib-storage/src/Upload.ts
+++ b/lib/lib-storage/src/Upload.ts
@@ -247,10 +247,8 @@ export class Upload extends EventEmitter {
         }
       }
 
-      // If the part is already uploaded, skip it.
-      if (this.uploadedParts.some((uploadedPart) => uploadedPart.PartNumber === dataPart.partNumber)) {
-        continue;
-      }
+      // ✅ Suggestion: Use early continue with concise condition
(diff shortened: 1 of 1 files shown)


lib/lib-storage/src/Upload.ts
```diff
@@ -247,6 +247,11 @@ export class Upload extends EventEmitter {
         }
       }
 
+      // If the part is already uploaded, skip it.
+      if (this.uploadedParts.some((uploadedPart) => uploadedPart.PartNumber === dataPart.partNumber)) {
```
dr-bharada:
```typescript
// ✅ Suggestion: Use early continue with concise condition
if (this.uploadedParts.some(p => p.PartNumber === dataPart.partNumber)) continue;
```

PR #7229, opened 2025-07-28
test(core): new integration tests

kuhe:
### Issue
internal JS-6084

### Description
adds new integration tests for
- account id endpoints
- pagination
- request compression
- lazy json string

PR #7230, opened 2025-07-28
test: add assorted integ/e2e tests

kuhe:
### Issue
Follows https://github.com/aws/aws-sdk-js-v3/pull/7229

### Testing
- s3 special characters in keys
- service id (client metadata)
- paginators
- waiters
- s3 object lambda
- rds signer
- dsql signer
- polly signer


clients/client-s3/test/e2e/s3-object-features.e2e.spec.ts
```diff
@@ -91,6 +91,67 @@ describe("@aws-sdk/client-s3", () => {
...
+      afterAll(async () => {
+        for (const key of keys) {
+          await client
+            .deleteObject({
+              Bucket,
+              Key: key,
+            })
+            .catch(() => {});
+        }
+      });
```
trivikr, line 128: nit: afterAll is not needed since the test deletes all the objects containing special characters?
```suggestion
```

kuhe: the test is specifically for the `deleteObjects` (plural) operation. If it failed, the backup is to run single `deleteObject` operations for the keys to return to a neutral state. 

The presence of the objects in the bucket prior to the test isn't a big deal either way, though.

trivikr: You're right. The objects will need to be deleted if test fails, so `afterAll` is needed.

PR #7250, opened 2025-08-08
chore(client-s3): bucket contextParam customization for schema-serde mode

kuhe:
### Issue
https://github.com/smithy-lang/smithy-typescript/issues/1600

### Description
For schema-serde mode:
- applies model transform removing the `/{Bucket}` endpoint uri prefix IFF input members group has `Bucket` and the member is a `contextParam` (trait).
- error handlers for schema-serde protocols have more consistent error metadata assignment

### Testing
S3 e2e tests, integration (CI)

### Checklist


packages/core/src/submodules/protocols/xml/AwsRestXmlProtocol.ts
```diff
@@ -61,16 +61,6 @@ export class AwsRestXmlProtocol extends HttpBindingProtocol {
...
 
-    request.path =
-      String(request.path)
-        .split("/")
-        .filter((segment) => {
-          // for legacy reasons,
-          // Bucket is in the http trait but is handled by endpoints ruleset.
-          return segment !== "{Bucket}";
-        })
-        .join("/") || "/";
```
kuhe, line 72: removed this hack, replaced by model preprocessing transform

trivikr (approved):

PR #7256, opened 2025-08-18
test(middleware-flexible-checksums): retry reuses the computed checksum

trivikr:
### Issue
Internal JS-5905

### Description
Adds an integration test which confirms checksum is not recomputed on retries 

### Testing
CI

---

### DIFF
--- a/packages/middleware-flexible-checksums/src/middleware-flexible-checksums.integ.spec.ts
+++ b/packages/middleware-flexible-checksums/src/middleware-flexible-checksums.integ.spec.ts
@@ -2,7 +2,6 @@
 import "@aws-sdk/crc64-nvme-crt";
 
 import { ChecksumAlgorithm, S3 } from "@aws-sdk/client-s3";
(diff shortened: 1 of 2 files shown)


trivikr:
Although the test is successful locally, it fails in CI

```console

 FAIL  packages/middleware-flexible-checksums/src/middleware-flexible-checksums.integ.spec.ts > middleware-flexible-checksums > S3 > putObject > retry doesn't recompute the checksum
AssertionError: expected "spy" to be called 1 times, but got 0 times
 ❯ packages/middleware-flexible-checksums/src/middleware-flexible-checksums.integ.spec.ts:153:39
    151| 
    152|         // Validate that flexibleChecksumsMiddleware is called once.
    153|         expect(mockFlexChecksCallsFn).toHaveBeenCalledTimes(1);
       |                                       ^
    154|         // Validate that retryMiddleware is called maxAttempts times.

```

smilkuri (approved):

PR #7293, opened 2025-08-28
chore(middleware-recursion-detection): read TraceID from Lambda Invoke Store

trivikr:
### Issue
Internal JS-6050

### Description
Reads TraceID from Lambda Invoke Store

### Testing
Local unit testing

---

### DIFF
--- a/packages/middleware-recursion-detection/src/recursionDetectionMiddleware.ts
+++ b/packages/middleware-recursion-detection/src/recursionDetectionMiddleware.ts
@@ -1,3 +1,4 @@
+import { InvokeStore } from "@aws/lambda-invoke-store";
 import { HttpRequest } from "@smithy/protocol-http";
 import {
   BuildHandler,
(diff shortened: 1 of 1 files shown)


packages/middleware-recursion-detection/src/index.ts
```diff
@@ -1,3 +1,4 @@
+import { InvokeStore } from "@aws/lambda-invoke-store";
```
kuhe: there should be a browser version created for this package (middleware-recursion-detection) that is an empty middleware, because the import of lambda invoke store will try to import async_hooks

trivikr:
Converted to draft. 

This PR can be rebased and made ready for review once https://github.com/aws/aws-sdk-js-v3/pull/7326 is merged.

packages/middleware-recursion-detection/src/index.ts
trivikr: Done in https://github.com/aws/aws-sdk-js-v3/pull/7326

Issue #7297, opened 2025-09-01
`Cannot read properties of undefined` if the logger is an object that references other members


rix0rrr:
### Checkboxes for prior research

- [x] I've gone through [Developer Guide](https://docs.aws.amazon.com/sdk-for-javascript/v3/developer-guide) and [API reference](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest)
- [x] I've checked [AWS Forums](https://forums.aws.amazon.com) and [StackOverflow](https://stackoverflow.com/questions/tagged/aws-sdk-js).
- [x] I've searched for [previous similar issues](https://github.com/aws/aws-sdk-js-v3/issues) and didn't find any solution.

### Describe the bug

In the `fromHttp` credential provider, the function `options.logger.warn` is copied into a local variable, and that function is subsequently called:

https://github.com/aws/aws-sdk-js-v3/blob/6681c8008308764078b7c44df99262f4e52ff0c2/packages/credential-provider-http/src/fromHttp/fromHttp.ts#L30-L39

If `options.logger` happens to point to a class, that might be implemented with a reference to `this`, then the reference to `this` is lost and an error will be thrown.

```ts
class SomeLogger {
  public warn(...args: string[]) {
     this.helper(...args);     // 💥 'this' is not bound to anything because the function was not called on an object
     // Cannot read properties of undefined
  }

  private helper(...args:string[]) {
    // whatever
  }
}
```

The warn function should retain a reference to the original object by using `bind()`, otherwise it is not possible to use classes as loggers which is very unexpected behavior.

```ts
const warn: (warning: string) => void =
    options.logger?.constructor?.name === "NoOpLogger" || !options.logger ? console.warn : options.logger.warn.bind(options.logger);
```

This pattern might occur in more places than just the place we ran into it.

### Regression Issue

- [ ] Select this option if this issue appears to be a regression.

### SDK version number

Latest

### Which JavaScript Runtime is this issue in?

Node.js

### Details of the browser/Node.js/ReactNative version

All nodes

### Reproduction Steps

See above.

### Observed Behavior

See above.

### Expected Behavior

See above.

smilkuri:
The fix has been merged and is available in **@aws-sdk/credential-provider-node** version **3.882.0**. https://www.npmjs.com/package/@aws-sdk/credential-provider-node/v/3.882.0

PR #7302, opened 2025-09-02
fix(credential-provider-http): fix logger binding issue

smilkuri:
### Issue
#7297 

### Description

- `FromHttpOptions` extends `CredentialProviderOptions` which has an optional `Logger` interface. When users provide class-based logger without binding breaks the `this` context resulting in `Cannot read properties of undefined` errors, if the logger is an object that references other members.
- Added .bind(options.logger) to preserve method context.

### Testing
Locally

```
yarn test:all
Tasks:    516 successful, 516 total
Cached:    516 cached, 516 total
  Time:    25.9s >>> FULL TURBO

 PASS  packages/signature-v4-crt/src/CrtSignerV4.spec.ts
 PASS  packages/signature-v4-crt/src/crtSuite.spec.ts

Test Suites: 2 passed, 2 total
Tests:       25 passed, 25 total
Snapshots:   0 total
Time:        4.837 s, estimated 5 s
Ran all test suites.
lerna notice cli v5.5.2
lerna notice filter including "@aws-sdk/{fetch-http-handler,hash-blob-browser}"
lerna info filter [ '@aws-sdk/{fetch-http-handler,hash-blob-browser}' ]
lerna ERR! EFILTER No packages remain after filtering [ '@aws-sdk/{fetch-http-handler,hash-blob-browser}' ]
```

### Checklist
- [n/a ] If you wrote E2E tests, are they resilient to concurrent I/O?

---


packages/credential-provider-http/src/fromHttp/fromHttp.ts
```diff
@@ -28,7 +28,9 @@ export const fromHttp = (options: FromHttpOptions = {}): AwsCredentialIdentityPr
   const tokenFile = options.awsContainerAuthorizationTokenFile ?? process.env[AWS_CONTAINER_AUTHORIZATION_TOKEN_FILE];
 
   const warn: (warning: string) => void =
-    options.logger?.constructor?.name === "NoOpLogger" || !options.logger ? console.warn : options.logger.warn;
+    options.logger?.constructor?.name === "NoOpLogger" || !options.logger
```
kuhe:
```suggestion
    options.logger?.constructor?.name === "NoOpLogger" || !options.logger?.warn
```

our types assume when logger exists, logger.warn exists, but let's be safe here. Some users have passed incomplete loggers in the past.

PR #7314, opened 2025-09-04
chore(core/protocols): add fallback for schema short name

kuhe:
### Issue
https://github.com/smithy-lang/smithy-typescript/issues/1600 (epic)

https://github.com/smithy-lang/smithy-typescript/pull/1681 - schema refactor includes adding a namespace string to all unwrapped schema objects, such that the `name` field contains only the short name.

### Description
Makes the SDK compatible with both pre and post merge of the above smithy-ts PR.

### Testing
`make test-protocols` both with current code and changes from [1681](https://github.com/smithy-lang/smithy-typescript/pull/1681)

### Checklist

### DIFF
--- a/packages/core/src/submodules/protocols/ProtocolLib.ts
+++ b/packages/core/src/submodules/protocols/ProtocolLib.ts
@@ -67,7 +67,8 @@ export class ProtocolLib {
     } else if (!inputSchema.isUnitSchema()) {
       const hasBody = Object.values(members).find((m) => {
         const { httpQuery, httpQueryParams, httpHeader, httpLabel, httpPrefixHeaders } = m.getMergedTraits();
-        return !httpQuery && !httpQueryParams && !httpHeader && !httpLabel && httpPrefixHeaders === void 0;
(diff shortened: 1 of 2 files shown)


packages/core/src/submodules/protocols/ProtocolLib.ts
```diff
@@ -67,7 +67,8 @@ export class ProtocolLib {
     } else if (!inputSchema.isUnitSchema()) {
       const hasBody = Object.values(members).find((m) => {
         const { httpQuery, httpQueryParams, httpHeader, httpLabel, httpPrefixHeaders } = m.getMergedTraits();
-        return !httpQuery && !httpQueryParams && !httpHeader && !httpLabel && httpPrefixHeaders === void 0;
+        const noPrefixHeaders = httpPrefixHeaders === void 0;
```
kuhe, line 70: added this variable because the previous expression was potentially confusing as to the order of operations of `&&` and `===`.

smilkuri (approved):

PR #7322, opened 2025-09-08
fix(lib-storage): improve concurrency and abortion management

matthieusieben:
### Issue
https://github.com/aws/aws-sdk-js-v3/issues/6426

### Description

This change improves the following aspects of the lib:

- Options use an `abortSignal` instead of an `abortController` in order to be more consistent with other APIs
- The code now actually waits that *all* pending uploads are finished before calling `AbortMultipartUpload`
- The code no longer returns control to the caller if operations are still pending in the background. Before, a `Promise.race` would allow multiple requests to still be made before returning control.
- Uploads will no longer throw if they are aborted during the `PutObjectTaggingCommand` phase, and actually succeeded.
- Uploads of individual chunks will be cancelled as soon as the upload is aborted, instead of waiting for the chunk upload to complete before cancelling the whole process.

### Testing
Not tested.

### Additional context
We encountered upload "hangs" in our code, similar to those described in https://github.com/aws/aws-sdk-js-v3/issues/6426 . While investigating, I encountered several opportunities for improvement, as noted in the Description above.

### Checklist

---


lib/lib-storage/src/Upload.ts
```diff
@@ -111,7 +112,12 @@ export class Upload extends EventEmitter {
       );
     }
     this.sent = true;
-    return await Promise.race([this.__doMultipartUpload(), this.__abortTimeout(this.abortController.signal)]);
```
matthieusieben, line 132: The race here could cause the returned promise to be rejected when `__doMultipartUpload` actually succeeds (and is still working on updating tags).

matthieusieben:
Hey @kuhe @trivikr , sorry to bother you. Any chance you'd might be giving this a look ?

lib/lib-storage/src/Upload.ts
```diff
@@ -353,28 +366,27 @@ export class Upload extends EventEmitter {
...
   private async __doMultipartUpload(): Promise<CompleteMultipartUploadCommandOutput> {
     const dataFeeder = getChunk(this.params.Body, this.partSize);
-    const concurrentUploaderFailures: Error[] = [];
+    const concurrentUploads: Promise<void>[] = [];
 
     for (let index = 0; index < this.queueSize; index++) {
-      const currentUpload = this.__doConcurrentUpload(dataFeeder).catch((err) => {
-        concurrentUploaderFailures.push(err);
-      });
-      this.concurrentUploaders.push(currentUpload);
```
matthieusieben, line 362: There was no point in keeping a reference on the instance as it was not used anywhere else.

lib/lib-storage/src/types.ts
```diff
@@ -51,8 +54,15 @@ export interface Configuration {
...
+   *
+   * @deprecated use `abortSignal` instead.
+   */
+  abortController?: IAbortController;
+
+  /**
+   * Optional abort signal for controlling this upload's abort signal externally.
    */
-  abortController?: AbortController;
+  abortSignal?: globalThis.AbortSignal;
```
matthieusieben, line 65: The code technically supports `IAbortSignal` but since those are kinda deprecated, and this is a new feature, there is no point in adding more legacy here.

PR #7324, opened 2025-09-08
docs: add effective-practices supplemental documentation

kuhe:
### Issue
Internal JS-6099

### Description
Adds a documentation page in supplemental docs that presents a set of effective practices.

Although written directly for SDK users by SDK maintainers, the contents can also be loaded into an LLM context to use for evaluating code.

### DIFF
--- a/supplemental-docs/EFFECTIVE_PRACTICES.md
+++ b/supplemental-docs/EFFECTIVE_PRACTICES.md
@@ -197,7 +197,8 @@ console.log(getObjectResponse.$metadata.httpStatusCode);
 // ⚠️ byte stream is unhandled, leaving a socket in use.
 ```
 
-Although the API call is performed and you have access to response, the connection will remain open until the byte stream, or payload, is read or discarded.
+Although the API call is performed, and you have access to response, the connection will remain open until the byte
+stream, or payload, is read or discarded.
 Not doing so will leave the connection open, and in Node.js this can lead to a condition we call socket exhaustion. In
 the worst cases this can cause your application to slow, leak memory, and/or deadlock.
 
@@ -231,6 +232,21 @@ if (case1) {
 }
 ```
 
+To identify _which_ operations contain byte stream response payloads, refer to our API documentation. In the
+"Example Syntax" section of each operation's API reference page, the field that constitutes a byte stream will be marked
+as such:
+
+```ts
+// { // GetObjectOutput
+//   Body: "<SdkStream>", // see \@smithy/types -> StreamingBlobPayloadOutputTypes
+//   ... other fields ...
+// };
+```
+
+in the same way
+as [GetObjectCommand](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/client/s3/command/GetObjectCommand/).
+The byte stream field will always be a top-level property of the response object.
+
 ### (4) Allow more time to establish connections when making requests cross-region
 
 This is outside the AWS SDK interfaces but an important consideration when making cross-region requests in AWS when
@@ -246,7 +262,8 @@ The default value of 250ms may be too low for some cross-region pairs within AWS
 opposite sides of the world, or simply in conditions of low network speed. This may manifest as an `AggregateError` with
 code `ETIMEDOUT` in Node.js.
 
-To increase this value within your application, use a `node` launch parameter such as `--network-family-autoselection-attempt-timeout=500` or
+To increase this value within your application, use a `node` launch parameter such as
+`--network-family-autoselection-attempt-timeout=500` or
 the `node:net` API:
 
 ```ts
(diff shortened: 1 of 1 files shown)


supplemental-docs/EFFECTIVE_PRACTICES.md
```diff
@@ -0,0 +1,231 @@
...
+  - [Incompatible write](#incompatible-write)
+  - [Incompatible read](#incompatible-read)
+  - [Recommended alternatives](#recommended-alternatives)
+- [(3) Always read streaming responses to completion or discard them](#3-always-read-streaming-responses-to-completion-or-discard-them)
+
+<!-- TOC end -->
+
+### (1) Minimize creating new copies of AWS SDK Clients for multiple operation calls
+
+The following example creates a client prior to making a request.
```
trivikr: A client needs to be created before making request. Rephrase this?

Example
```suggestion
The following example creates a new client in every iteration of the loop.
```

supplemental-docs/EFFECTIVE_PRACTICES.md
```diff
@@ -0,0 +1,231 @@
...
+  const client = new S3Client({
+    region,
+    credentials,
+  });
+  await client.send(new PutObjectCommand(item));
+}
+```
+
+In cases where the operations being called are to the same client configuration, i.e. the same region
+and with the same credentials, creating a new client is unnecessary. It adds additional work for the calling program,
```
trivikr: nit:
* un to not
* adds additional to increases
```suggestion
and with the same credentials, creating a new client is not necessary. It increases work for the calling program,
```

supplemental-docs/EFFECTIVE_PRACTICES.md
```diff
@@ -0,0 +1,231 @@
...
+await client.send(new CreateBucketCommand({ Bucket }));
+
+for (const item of items) {
+  await client.send(new PutObjectCommand(item));
+}
+
+const objects = await client.send(new ListObjectsV2Command({ Bucket }));
+```
+
+If you find a need to create a new client because of capacity issues with a single client,
```
trivikr: nit: remove redundant words
```suggestion
If you need to create a new client because of capacity issues with a single client,
```

supplemental-docs/EFFECTIVE_PRACTICES.md
```diff
@@ -0,0 +1,231 @@
...
+// ⚠️ incompatible reading, will throw an error: Uncaught TypeError: client.config.endpoint is not a function
+const endpoint = await client.config.endpoint();
+```
+
+This may seem initially reasonable, since each regional AWS service typically has a set of endpoints of the pattern
+`{service}.{region}.amazonaws.com`. However, AWS services can configure endpoints that differ based on many factors,
+including down to the distinct operation being called and its inputs. Therefore, the canonical endpoint cannot be
+accurately given before the operation and operation inputs are known.
+
+For example, the AWS SDK's S3 client uses the bucket name in the hostname, an operation level parameter, and the
```
trivikr, line 135: Should we document the correct way to get endpoints in this section?

```ts
import { S3Client, ListTablesCommand } from "@aws-sdk/client-s3";
import { getEndpointFromInstructions } from "@smithy/middleware-endpoint";

const client = new S3Client();
const command = new ListTablesCommand();

const endpoint = await getEndpointFromInstructions(command.input, CreateAccessPointCommand, client.config);
console.log(endpoint.url.href);
```

supplemental-docs/EFFECTIVE_PRACTICES.md
```diff
@@ -0,0 +1,231 @@
...
+### (3) Always read streaming responses to completion or discard them
+
+Some operations, the most common of which
+is [GetObjectCommand](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/client/s3/command/GetObjectCommand/),
+return a byte stream.
+
+Although awaiting such a request will return an HTTP status code and response headers,
+
+```ts
+const getObjectResponse = client.send(GetObjectCommand({ Bucket, Key }));
```
trivikr: missing await
```suggestion
const getObjectResponse = await client.send(GetObjectCommand({ Bucket, Key }));
```

supplemental-docs/EFFECTIVE_PRACTICES.md
```diff
@@ -0,0 +1,231 @@
...
+
+console.log(getObjectResponse.$metadata.httpStatusCode);
+// ⚠️ byte stream is unhandled, leaving a socket in use.
+```
+
+the request is incomplete. The connection will remain open until the byte stream, or payload, is read or discarded.
```
trivikr: It's arguable if the request is incomplete from code perspective. Maybe rephrase it?

Example
```suggestion
Although the API call is performed and you have access to response, the connection will remain open until the byte stream, or payload, is read or discarded.
```

End it by stating this only specific to operations which return byte streams, and how customer can find it out.

supplemental-docs/EFFECTIVE_PRACTICES.md
trivikr: It was on line 163.

PR #7326, opened 2025-09-09
chore(middleware-recursion-detection): add no-op middleware outside of Node.js

trivikr:
### Issue
Noticed in https://github.com/aws/aws-sdk-js-v3/pull/7293

### Description
The recursion detection middleware should be specific to Node.js, and it currently checks for `options.runtime` at runtime. This is okay in it's current state, as it only checks for environment variables.

We're adding a new dependency in https://github.com/aws/aws-sdk-js-v3/pull/7293 which uses Node.js specific `async_hooks`. Since this is a import level, there should be no-op in browser and react-native

### Testing
CI

---


packages/middleware-recursion-detection/package.json
```diff
@@ -53,5 +53,12 @@
     "downlevel-dts": "0.10.1",
     "rimraf": "3.0.2",
     "typescript": "~5.8.3"
+  },
+  "browser": {
+    "./dist-es/recursionDetectionMiddleware": "./dist-es/recursionDetectionMiddleware.no-op"
+  },
+  "react-native": {
+    "./dist-es/recursionDetectionMiddleware": "./dist-es/recursionDetectionMiddleware.no-op",
+    "./dist-cjs/recursionDetectionMiddleware": "./dist-cjs/recursionDetectionMiddleware.no-op"
```
kuhe: want to rename this to `.browser` and `.native`? some bundlers replace files when there is a same file with the `.browser` or `.native` extension infix. As a convention if the package metadata doesn't work.

kuhe: notably, that's why our clients' `runtiemConfig.[browser/native].js` files get replaced even though there is no replacement metadata in the package.json files

trivikr: Done in [`4e27682` (#7326)](https://github.com/aws/aws-sdk-js-v3/pull/7326/commits/4e27682a577e5a5e2f805fa8728fae7728c06dd6)

PR #7346, opened 2025-09-15
feat(xml-builder): use DOMParser for browser XML parsing

kuhe:
### Issue
https://github.com/aws/aws-sdk-js-v3/issues/3979

### Description
Uses window.DOMParser in browsers instead of fast-xml-parser

### Testing
new unit tests, also ran all of our existing S3 tests with the browser bundle, confirming that it is using the replacement.

### Checklist
  - XML parsing is only used on the response, at which point unit tests do the same job


kuhe:
moved to aws origin

PR #7347, opened 2025-09-15
feat(xml-builder): use DOMParser for browser XML parsing

kuhe:
### Issue
https://github.com/aws/aws-sdk-js-v3/issues/3979

### Description
Uses window.DOMParser in browsers instead of fast-xml-parser

### Testing
new unit tests, also ran all of our existing S3 tests with the browser bundle, confirming that it is using the replacement.

### Checklist
  - XML parsing is only used on the response, at which point unit tests do the same job


kuhe:
- [x] update root level aggregate unit and browser tests to exclude/include the xml-parser test file

trivikr:
Protocol tests need to be updated to use DOMParser
* Codegen https://github.com/smithy-lang/smithy-typescript/blob/434dd9c93329166b6bacdc54adb20511f6460575/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/HttpProtocolTestGenerator.java#L606-L611
* Usage https://github.com/aws/aws-sdk-js-v3/blob/29220637299ca59227319e7bc334051aba1754be/private/aws-protocoltests-restxml/package.json#L65

It can be done in alternative PR as a PoC, and merged, before taking up this change.

trivikr (approved):

PR #7363, opened 2025-09-19
fix(lib-storage): add validation for partCount and contentLength for multipart upload

smilkuri:
### Issue
Internal JS-6196

### Description

- Adds validation to check total number of UploadPart requests the SDK has sent matches the expected number of parts.
- Adds validation to check the content length of each UploadPart request matches the expected part size.

### Testing
Locally
`yarn test` and `yarn test:e2e` are successful
```
dev-dsk-smilkuri-1a-17aece6b % yarn test

 RUN  v3.2.4 /local/home/smilkuri/aws-sdk-js-v3/lib/lib-storage

 ✓ src/chunks/getChunkUint8Array.spec.ts (6 tests) 7ms
 ✓ src/chunks/getDataReadableStream.spec.ts (4 tests) 168ms
 ✓ src/chunks/getDataReadable.spec.ts (3 tests) 167ms
 ✓ src/index.spec.ts (1 test) 2ms
 ✓ src/Upload.spec.ts (34 tests) 22888ms
   ✓ Upload > should add tags to the object if tags have been added multi-part  22807ms

 Test Files  5 passed (5)
      Tests  48 passed (48)
   Start at  16:15:06
   Duration  23.90s (transform 861ms, setup 0ms, collect 2.42s, tests 23.23s, environment 1ms, prepare 584ms)
```

### Checklist

---

### DIFF
--- a/lib/lib-storage/src/Upload.ts
+++ b/lib/lib-storage/src/Upload.ts
@@ -47,7 +47,7 @@ export class Upload extends EventEmitter {
 
   // Defaults.
   private readonly queueSize: number = 4;
-  private readonly partSize = Upload.MIN_PART_SIZE;
+  private readonly partSize: number;
   private readonly leavePartsOnError: boolean = false;
   private readonly tags: Tag[] = [];
 
@@ -66,6 +66,7 @@ export class Upload extends EventEmitter {
 
   private uploadedParts: CompletedPart[] = [];
   private uploadEnqueuedPartsCount = 0;
+  private expectedPartsCount = 0;
   /**
    * Last UploadId if the upload was done with MultipartUpload and not PutObject.
    */
@@ -81,19 +82,21 @@ export class Upload extends EventEmitter {
 
     // set defaults from options.
     this.queueSize = options.queueSize || this.queueSize;
-    this.partSize = options.partSize || this.partSize;
     this.leavePartsOnError = options.leavePartsOnError || this.leavePartsOnError;
     this.tags = options.tags || this.tags;
 
     this.client = options.client;
     this.params = options.params;
 
-    this.__validateInput();
-
     // set progress defaults
     this.totalBytes = byteLength(this.params.Body);
     this.bytesUploadedSoFar = 0;
     this.abortController = options.abortController ?? new AbortController();
+
+    this.partSize = this.__calculatePartSize(this.totalBytes ?? 0, Upload.MIN_PART_SIZE);
+    this.expectedPartsCount = this.totalBytes ? Math.ceil(this.totalBytes / this.partSize) : 1;
+
+    this.__validateInput();
   }
 
   async abort(): Promise<void> {
@@ -234,6 +237,8 @@ export class Upload extends EventEmitter {
         return;
       }
 
+      this.__validateUploadPart(dataPart, this.partSize);
+
       // Use put instead of multipart for one chunk uploads.
       if (dataPart.partNumber === 1 && dataPart.lastPart) {
         return await this.__uploadUsingPut(dataPart);
@@ -364,6 +369,12 @@ export class Upload extends EventEmitter {
 
     let result;
     if (this.isMultiPart) {
+      if (this.totalBytes && this.uploadedParts.length !== this.expectedPartsCount) {
+        throw new Error(
+          `Expected ${this.expectedPartsCount} number of parts but uploaded only ${this.uploadedParts.length} parts`
+        );
+      }
+
       this.uploadedParts.sort((a, b) => a.PartNumber! - b.PartNumber!);
 
       const uploadCompleteParams = {
@@ -425,6 +436,31 @@ export class Upload extends EventEmitter {
         reject(abortError);
       };
     });
+  }
+
+  private __calculatePartSize(targetPartSizeBytes: number, minPartSize: number): number {
+    const calculatedPartSize = Math.floor(targetPartSizeBytes / this.MAX_PARTS);
+    return Math.max(minPartSize, calculatedPartSize);
+  }
+
+  private __validateUploadPart(dataPart: RawDataPart, partSize: number): void {
+    const actualPartSize = byteLength(dataPart.data) || 0;
+
+    // Skip validation for single-part uploads (PUT operations)
+    if (dataPart.partNumber === 1 && dataPart.lastPart) {
+      return;
+    }
+
+    if (actualPartSize === 0) {
+      throw new Error(`Content length is missing on the data for part number ${dataPart.partNumber}`);
+    }
+
+    // Validate part size (last part may be smaller)
+    if (!dataPart.lastPart && actualPartSize !== partSize) {
+      throw new Error(
+        `The Part size for part number ${dataPart.partNumber}, size ${actualPartSize} does not match expected size ${partSize}`
+      );
+    }
   }
 
   private __validateInput(): void {
--- a/lib/lib-storage/src/Upload.ts
+++ b/lib/lib-storage/src/Upload.ts
@@ -66,7 +66,7 @@ export class Upload extends EventEmitter {
 
   private uploadedParts: CompletedPart[] = [];
   private uploadEnqueuedPartsCount = 0;
-  private expectedPartsCount = 0;
+  private expectedPartsCount?: number;
   /**
    * Last UploadId if the upload was done with MultipartUpload and not PutObject.
    */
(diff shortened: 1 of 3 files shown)


lib/lib-storage/src/Upload.ts
```diff
@@ -427,6 +438,31 @@ export class Upload extends EventEmitter {
     });
   }
 
+  private __calculatePartSize(targetPartSizeBytes: number, minPartSize: number): number {
```
kuhe: The variable name `targetPartSizeBytes` is inaccurate because it is not a part size. It is the size of the entire object.

There's no need for the second arg since it is only ever called with `Upload.MIN_PART_SIZE`.

But beyond that, there's no need for this function since it's only called once.

```ts
this.partSize = Math.max(
  Upload.MIN_PART_SIZE, 
  Math.floor((this.totalBytes ?? 0) / this.MAX_PARTS)
);
```

lib/lib-storage/src/Upload.ts
```diff
@@ -81,19 +82,21 @@ export class Upload extends EventEmitter {
...
 
-    this.__validateInput();
-
     // set progress defaults
     this.totalBytes = byteLength(this.params.Body);
     this.bytesUploadedSoFar = 0;
     this.abortController = options.abortController ?? new AbortController();
+
+    this.partSize = this.__calculatePartSize(this.totalBytes ?? 0, Upload.MIN_PART_SIZE);
+    this.expectedPartsCount = this.totalBytes ? Math.ceil(this.totalBytes / this.partSize) : 1;
```
kuhe: I get the sense that this is a check for whether totalBytes was able to be calculated and not whether or not it is the numeric value 0.

In that case, it would be clearer to state the check as `this.totalBytes !== undefined ?`. However, with a zero byte input, the expected part count is indeed 1, so perhaps no change is needed.

----

But, why would expectedPartsCount be 1 if totalBytes could not be calculated? It could be a stream of an unknown byte length. In this case, both totalBytes and expectedPartCount should be `undefined`, since they would not be knowable.

lib/lib-storage/src/Upload.ts
```diff
@@ -66,6 +66,7 @@ export class Upload extends EventEmitter {
 
   private uploadedParts: CompletedPart[] = [];
   private uploadEnqueuedPartsCount = 0;
+  private expectedPartsCount = 0;
```
kuhe: it should not be assigned a value here, since 0 is never possible and a value is always assigned in the constructor.

But since my comments below understand this value could be impossible to calculate, in which case it should be `undefined`. Therefore the type should be `private expectedPartsCount?: number;`.

lib/lib-storage/src/Upload.ts
```diff
@@ -364,6 +369,12 @@ export class Upload extends EventEmitter {
 
     let result;
     if (this.isMultiPart) {
+      if (this.totalBytes && this.uploadedParts.length !== this.expectedPartsCount) {
```
kuhe: `totalBytes` can be zero and `expectedPartsCount` would be 1 in that case.

This should instead be
```js
const { expectedPartsCount, uploadedParts } = this;

if (expectedPartsCount !== undefined 
    && uploadedParts.length !== expectedPartsCount) {
```

lib/lib-storage/src/Upload.ts
```diff
@@ -364,6 +369,12 @@ export class Upload extends EventEmitter {
...
+        throw new Error(
+          `Expected ${this.expectedPartsCount} number of parts but uploaded only ${this.uploadedParts.length} parts`
```
kuhe:
```suggestion
          `Expected ${this.expectedPartsCount} part(s) but uploaded ${this.uploadedParts.length} part(s).`
```

Can't say "only" because the check is for inequality. The uploaded parts count could be higher.

lib/lib-storage/src/Upload.ts
```diff
@@ -427,6 +438,31 @@ export class Upload extends EventEmitter {
     });
   }
 
+  private __calculatePartSize(targetPartSizeBytes: number, minPartSize: number): number {
+    const calculatedPartSize = Math.floor(targetPartSizeBytes / this.MAX_PARTS);
+    return Math.max(minPartSize, calculatedPartSize);
+  }
+
+  private __validateUploadPart(dataPart: RawDataPart, partSize: number): void {
```
kuhe: why is `partSize` an argument to this method when it is only ever the same instance variable `this.partSize`?

lib/lib-storage/src/Upload.ts
```diff
@@ -427,6 +438,31 @@ export class Upload extends EventEmitter {
...
+    const actualPartSize = byteLength(dataPart.data) || 0;
```
kuhe: the fallback value of `undefined` should be used, since real data may be zero bytes.

lib/lib-storage/src/Upload.ts
```diff
@@ -427,6 +438,31 @@ export class Upload extends EventEmitter {
...
+
+    // Skip validation for single-part uploads (PUT operations)
+    if (dataPart.partNumber === 1 && dataPart.lastPart) {
+      return;
+    }
+
+    if (actualPartSize === 0) {
```
kuhe:
```suggestion
    if (actualPartSize === undefined) {
```

lib/lib-storage/src/Upload.ts
```diff
@@ -427,6 +438,31 @@ export class Upload extends EventEmitter {
...
+      throw new Error(`Content length is missing on the data for part number ${dataPart.partNumber}`);
```
kuhe: This statement implies the response header is missing, since content length usually refers to the header of that name.

But it's not related to that header. This error branch is only due to programming error on our part, right? The error message should be developer-facing, e.g.

> Error: a dataPart was generated without a measurable data chunk size.

lib/lib-storage/src/Upload.ts
```diff
@@ -234,6 +237,8 @@ export class Upload extends EventEmitter {
         return;
       }
 
+      this.__validateUploadPart(dataPart, this.partSize);
```
kuhe: this should be moved closer to `UploadPartCommand` to prevent any changes happening to dataPart between the validation and sending out the data

lib/lib-storage/src/Upload.ts
```diff
@@ -427,6 +437,28 @@ export class Upload extends EventEmitter {
     });
   }
 
+  private __validateUploadPart(dataPart: RawDataPart): void {
+    const actualPartSize = byteLength(dataPart.data) || undefined;
```
kuhe: what falsy values does byteLength return? What if it's zero?

lib/lib-storage/src/Upload.ts
```diff
@@ -427,6 +437,28 @@ export class Upload extends EventEmitter {
...
+
+    // Skip validation for single-part uploads (PUT operations)
+    if (dataPart.partNumber === 1 && dataPart.lastPart) {
+      return;
```
kuhe: it seems you could return here without checking the `actualPartSize` (move it after this)

lib/lib-storage/src/Upload.ts
```diff
@@ -427,6 +437,28 @@ export class Upload extends EventEmitter {
...
+    if (actualPartSize === undefined) {
+      throw new Error(
+        `A dataPart was generated without a measurable data chunk size for part number ${dataPart.partNumber}`
+      );
+    }
+
+    // Validate part size (last part may be smaller)
+    if (!dataPart.lastPart && actualPartSize !== this.partSize) {
+      throw new Error(
+        `The Part size for part number ${dataPart.partNumber}, size ${actualPartSize} does not match expected size ${this.partSize}`
```
kuhe:
```suggestion
        `The byte size for part number ${dataPart.partNumber}, size ${actualPartSize} does not match expected size ${this.partSize}`
```

lib/lib-storage/src/Upload.ts
smilkuri: It can return zero for empty parts, may be for scenarios like empty file?

kuhe: if `byteLength` can return undefined, then the expression should be `byteLength(dataPart.data);` 

zero byte length should not be converted to `undefined`.

lib/lib-storage/src/lib-storage.e2e.spec.ts
```diff
@@ -139,6 +139,43 @@ describe("@aws-sdk/lib-storage", () => {
...
+              Bucket,
+              Key: `validation-test-${Date.now()}`,
+              Body: Buffer.alloc(1024 * 1024 * 10),
+            },
+          });
+
+          (upload as any).uploadedParts = [{ PartNumber: 1, ETag: "etag1" }];
+          (upload as any).isMultiPart = true;
+
+          await expect(upload.done()).rejects.toThrow(/Expected \d+ part\(s\) but uploaded \d+ part\(s\)\./);
```
kuhe, line 177: does this upload an object or not?

if it does, the test needs to delete it

smilkuri: No it doesn't.

PR #7373, opened 2025-09-24
test(clients): update payload types in client tests

kuhe:
- **test: fix lex runtime and mediastore payload tests**
- **test: fix select object content event stream test**

### Issue

### Description

### Testing
How was this change tested?

### Additional context

### Checklist

---


clients/client-lex-runtime-service/test/LexRuntimeService.spec.ts
```diff
@@ -1,36 +1,32 @@
...
+        },
       });
+
       return await client.postContent({
         botAlias: "alias",
         botName: "bot",
         userId: "user",
         contentType: "text/plain; charset=utf-8",
-        inputStream: "hello world!",
+        inputStream: Readable.from(new Uint8Array([1, 2, 3, 4, 5, 6, 7, 8])),
```
kuhe, line 29: rather than always being set to UNSIGNED-PAYLOAD because the operation is modeled as having a data stream, in schema-serde the sigv4 signer sets a sha256 hash if the input is buffered (string or Buffer) and UNSIGNED-PAYLOAD only when it is a stream.

clients/client-mediastore-data/test/MediaStoreData.spec.ts
```diff
@@ -1,33 +1,29 @@
...
+      requireRequestsFrom(client).toMatch({
+        headers: {
+          "x-amz-content-sha256": "UNSIGNED-PAYLOAD",
+        },
       });
+
       return await client.putObject({
         Path: "foo.avi",
-        Body: "binary body",
+        Body: Readable.from(new Uint8Array([1, 2, 3, 4, 5, 6, 7, 8])),
```
kuhe, line 26: same as LexRuntime

clients/client-s3/test/integ/s3-selectObjectContent.integ.spec.ts
```diff
@@ -34,16 +32,49 @@ describe("selectObjectContent", () => {
...
-            `<End></End>`,
-            `</Payload>`,
-          ]),
+          headers: {
+            "x-amz-checksum-crc32": "3117266289",
+          },
+          body: {
+            *[Symbol.asyncIterator]() {
+              // see https://docs.aws.amazon.com/AmazonS3/latest/API/RESTSelectObjectAppendix.html
+              yield new Uint8Array([
```
kuhe, line 41: These bytes are the actual event stream whereas the previous XML was from a generated structural example and not what is actually sent.

PR #7374, opened 2025-09-24
chore(core/protocols): preparatory fixes for schemas

kuhe:
### Issue
https://github.com/smithy-lang/smithy-typescript/issues/1600
preparatory compatibility fixes for Schemas

### Description
Fixes for unreleased features.

- fix for applying the LazyJsonString adapter in cases where the input is a JSON object rather than a string
- change test assertions to ignore ordering of query parameters
- return early for unit schema in XML shape deserializer
- fix to fully populate error fields in AwsQuery protocol
- defer content length determination to the contentLengthMiddleware rather than protocol methods

### Checklist

### DIFF
--- a/packages/middleware-sdk-ec2/src/index.ts
+++ b/packages/middleware-sdk-ec2/src/index.ts
@@ -5,7 +5,6 @@ import { SignatureV4 } from "@smithy/signature-v4";
 import { extendedEncodeURIComponent } from "@smithy/smithy-client";
 import type {
   AwsCredentialIdentity,
-  BodyLengthCalculator,
   ChecksumConstructor,
   Endpoint,
   HandlerExecutionContext,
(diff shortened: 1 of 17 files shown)


packages/middleware-sdk-ec2/src/index.ts
```diff
@@ -123,6 +125,9 @@ export function copySnapshotPresignedUrlMiddleware(options: PreviouslyResolved):
           if (!(request.body ?? "").includes("PresignedUrl=")) {
             request.body += `&PresignedUrl=${extendedEncodeURIComponent(args.input.PresignedUrl)}`;
           }
+          if (typeof options.bodyLengthChecker === "function") {
+            request.headers["content-length"] = String(options.bodyLengthChecker(request.body));
+          }
```
kuhe, line 130: This isn't strictly necessary, since the content length middleware was handling this when running after this middleware. 

However, explicitly setting the content length in the same code path where the request body is being directly modified is safer and gives more information to the reader.

PR #7403, opened 2025-10-04
chore(codegen): remove checks for deprecated SetShape

trivikr:
### Issue
* Sets are deprecated https://github.com/smithy-lang/smithy/pull/1278
* They're replaced with ListShape with uniqueItems trait.
  * Example https://github.com/aws/aws-sdk-js-v3/blob/ff5b37615ad1f77f22f13eb4dbab55188752ea5e/codegen/sdk-codegen/aws-models/dsql.json#L126-L135 

### Description
Remove checks for deprecated SetShape

### Testing
How was this change tested?

### Additional context

### Checklist

---


codegen/smithy-aws-typescript-codegen/src/main/java/software/amazon/smithy/aws/typescript/codegen/visitor/MemberDeserVisitor.java
```diff
@@ -199,11 +198,6 @@ public final String mapShape(MapShape shape) {
         return getDelegateDeserializer(shape);
     }
 
-    @Override
-    public final String setShape(SetShape shape) {
-        return getDelegateDeserializer(shape);
-    }
```
kuhe, line 218: if we decide we're not supporting sets, this method and all other code points where sets were handled should throw an error, rather than allow the customer to get possibly non-functional code

PR #7437, opened 2025-10-17
fix(cloudfront-signer): encode uri components in base url

kuhe:
### Issue
https://github.com/aws/aws-sdk-js-v3/issues/7428
and related to https://github.com/aws/aws-sdk-js-v3/pull/7237

### Description
in the cloudfront signer, encode search params contained in the base URL. This is more consistent with the URL object normalization that was occurring earlier.

### Testing
unit tests

### Checklist
	- unit tests are able to exercise the full public API of the signer

### DIFF
--- a/packages/cloudfront-signer/package.json
+++ b/packages/cloudfront-signer/package.json
@@ -23,6 +23,7 @@
   },
   "license": "Apache-2.0",
   "dependencies": {
+    "@smithy/core": "^3.16.1",
     "@smithy/url-parser": "^4.2.2",
     "tslib": "^2.6.2"
   },
--- a/packages/cloudfront-signer/src/sign.spec.ts
+++ b/packages/cloudfront-signer/src/sign.spec.ts
@@ -810,3 +810,21 @@ describe("getSignedUrl- when signing a URL with a date range", () => {
     expect(verifySignature(signatureQueryParam, policyStr)).toBeTruthy();
   });
 });
+
+describe("url component encoding", () => {
+  it("should use extended encoding for query params in the base URL", () => {
+    const url =
+      "https://d111111abcdef8.cloudfront.net/private-content/private.jpeg?q=!@#$%^&*()&image-description=aws's image&'''&!()=5";
+    const signedUrl = getSignedUrl({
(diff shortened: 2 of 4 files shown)


smilkuri (approved):

eliocapella:
It seems this change broke the ability to set filename*= in the content disposition

```js
function encodeRFC5987ValueChars(str) {
  return encodeURIComponent(str)
    .replace(/['()]/g, escape)
    .replace(/\*/g, "%2A")
    .replace(/%(?:7C|60|5E)/g, unescape);
}

/**
 * @param {string} key file data key
 * @param {object} options
 * @param {false | string} [options.download=false] filename if we want to force download on the browser
 * @returns {Shared.URL}
 */
exports.url = function (key, { download = false } = {}) {
  let url = `${env.AWS_CLOUDFRONT_BASE_URL}/${encodeURI(key)}`;
  if (download) {
    const encName = encodeRFC5987ValueChars(download);
    url += `?response-content-disposition=${encodeURIComponent(
      `attachment; filename=${encName}; filename*=UTF-8''${encName}`,
    )}`;
  }
  return /** @type {Shared.URL} */ (url);
};

/**
 * @param {string} key file data key
 * @param {object} options
 * @param {number} [options.hoursToExpire=24]
 * @param {false | string} [options.download=false] filename if we want to force download on the browser
 * @returns {Shared.URL}
 */
exports.signedUrl = function (
  key,
  { hoursToExpire = 24, download = false } = {},
) {
  return /** @type {Shared.URL} */ (
    getSignedCloudfrontUrl({
      url: exports.url(key, { download }),
      keyPairId: env.AWS_CLOUDFRONT_KEY_ID,
      privateKey: env.AWS_CLOUDFRONT_PRIVATE_KEY,
      dateLessThan: new Date(
        Date.now() + hoursToExpire * 60 * 60 * 1000,
      ).toISOString(),
    })
  );
};
```
this code stopped working after upgrading, it seems Cloudfront needs the `*` in `filename*` not to be encoded for the URLs not to work and now that isn't possible anymore.

I ended up implementing my own Cloudwatch signer

Issue #7439, opened 2025-10-17
Custom `user-agent` not propagated to nested STS client


jackrein:
### Checkboxes for prior research

- [x] I've gone through [Developer Guide](https://docs.aws.amazon.com/sdk-for-javascript/v3/developer-guide) and [API reference](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest)
- [x] I've checked [AWS Forums](https://forums.aws.amazon.com) and [StackOverflow](https://stackoverflow.com/questions/tagged/aws-sdk-js).
- [x] I've searched for [previous similar issues](https://github.com/aws/aws-sdk-js-v3/issues) and didn't find any solution.

### Describe the bug

When using role chaining, the custom user agent on the outer client is not propagated to the [nested STS client](https://github.com/aws/aws-sdk-js-v3/blob/58fd243b69cf60b965d32da01c3f2e2b6526fc6a/packages/credential-provider-ini/src/resolveAssumeRoleCredentials.ts#L114-L126).

### Regression Issue

- [ ] Select this option if this issue appears to be a regression.

### SDK version number

3.911.0

### Which JavaScript Runtime is this issue in?

Node.js

### Details of the browser/Node.js/ReactNative version

v18.20.2

### Reproduction Steps

1. Create the following content in `~/.aws/config` file:
    ```ini
    [profile second]
    credential_process=/usr/bin/get-credentials
    region=us-west-2
    
    [profile first]
    source_profile=second
    role_arn=arn:aws:iam::123456789012:role/Role
    ```
2. Use the SDK with the `first` role: `AWS_PROFILE=first npm run start` (`start` invokes the following script)
    ```ts
    import { S3Client, ListBucketsCommand } from '@aws-sdk/client-s3';

    const s3Client = new S3Client({
        defaultUserAgentProvider: () => Promise.resolve(["my-user-agent", "1"]), // or customUserAgent
        logger: {
            ...console
        }
    });
    const command = new ListBucketsCommand({});
    const response = await s3Client.send(command);
    ```

### Observed Behavior

Observe two AWS requests in CloudTrail. The  AssumeRole request retains the default `user-agent`. The ListBuckets request contains the custom user agent:
1. `aws-sdk-js/3.911.0 ua/2.1 os/darwin#24.6.0 lang/js md/nodejs#18.20.2 api/sts#3.911.0 m/E,w,v` vs
2. `my-user-agent m/E,T,i,o`

### Expected Behavior

[...]

PR #7445, opened 2025-10-20
fix(nested-clients): propagate clientConfig to inner STS client in role assumption

kuhe:
### Issue
https://github.com/aws/aws-sdk-js-v3/issues/7439

### Description
propagate clientConfig to inner STS client

### Testing
- [x] new integration tests
- [x] CI

### Checklist

### DIFF
--- a/clients/client-sts/src/defaultStsRoleAssumers.ts
+++ b/clients/client-sts/src/defaultStsRoleAssumers.ts
@@ -15,7 +15,7 @@ import type { STSClient, STSClientConfig, STSClientResolvedConfig } from "./STSC
 /**
  * @public
  */
-export type STSRoleAssumerOptions = Pick<STSClientConfig, "logger" | "region" | "requestHandler"> & {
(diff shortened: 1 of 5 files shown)


jackrein:
Will it be necessary to set both the parent client and the credentials provider `clientConfig`? The comment in the [docs](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/Package/-aws-sdk-credential-providers/#fromnodeproviderchain) states that `region` and `profile` will be propagated, but nothing else.

```ts
new S3Client({
    customUserAgent: [["custom-user-agent", "1.0.0"]],
    credentials: fromNodeProviderChain({
        clientConfig: {
            customUserAgent: [["custom-user-agent", "1.0.0"]],
        }
    })
});
```

kuhe:
Yes, setting a user agent value on the both outer and inner client configs will be required even after this change is published.

smilkuri (approved):

PR #7481, opened 2025-11-07
chore(codegen): sync for synthetic base exception naming collision

kuhe:
### Description
https://github.com/smithy-lang/smithy-typescript/compare/8aef84bd27a0bb995ef908239e7bb7c6c3a37435...08e3b88680fec11b216b8bffaa4a44dba76eee8b

### Testing
CI

### DIFF
--- a/clients/client-elastic-beanstalk/src/commands/DescribeEnvironmentManagedActionHistoryCommand.ts
+++ b/clients/client-elastic-beanstalk/src/commands/DescribeEnvironmentManagedActionHistoryCommand.ts
@@ -81,7 +81,7 @@ export interface DescribeEnvironmentManagedActionHistoryCommandOutput
  * @throws {@link ElasticBeanstalkServiceException} (client fault)
  *  <p>A generic service exception has occurred.</p>
  *
- * @throws {@link ElasticBeanstalkServiceException}
+ * @throws {@link ElasticBeanstalkSyntheticServiceException}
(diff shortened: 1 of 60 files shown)


clients/client-elastic-beanstalk/src/commands/DescribeEnvironmentManagedActionHistoryCommand.ts
```diff
@@ -81,7 +81,7 @@ export interface DescribeEnvironmentManagedActionHistoryCommandOutput
  * @throws {@link ElasticBeanstalkServiceException} (client fault)
  *  <p>A generic service exception has occurred.</p>
  *
- * @throws {@link ElasticBeanstalkServiceException}
+ * @throws {@link ElasticBeanstalkSyntheticServiceException}
  * <p>Base exception class for all service exceptions from ElasticBeanstalk service.</p>
```
kuhe, line 85: This line is an example of the problem.

The service model already has something called `ElasticBeanstalkServiceException` and it's even declared as an error of this operation!

So we end up with a somehwhat nonsensical 
```js
/**
 * @throws {@link ElasticBeanstalkServiceException} (client fault)
 *  <p>A generic service exception has occurred.</p>
 *
 * @throws {@link ElasticBeanstalkServiceException}
 * <p>Base exception class for all service exceptions from ElasticBeanstalk service.</p>
 */ 
```

Issue #7500, opened 2025-11-13
Cannot Delete Retained MQTT Messages Anymore


PeterBurner:
### Checkboxes for prior research

- [x] I've gone through [Developer Guide](https://docs.aws.amazon.com/sdk-for-javascript/v3/developer-guide) and [API reference](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest)
- [x] I've checked [AWS Forums](https://forums.aws.amazon.com) and [StackOverflow](https://stackoverflow.com/questions/tagged/aws-sdk-js).
- [x] I've searched for [previous similar issues](https://github.com/aws/aws-sdk-js-v3/issues) and didn't find any solution.

### Describe the bug

We are using the NodeJS package `@aws-sdk/client-iot-data-plane` to interact with AWS IoT Core. Since the update from v3.928.0 to v3.929.0 we can no longer delete retained messages. We usually send `undefined` in the payload of the `PublishCommand`. However with the new SDK version the topic is actually set to an empty object (`{}`) instead of being deleted. We also tried sending `null` or empty string `''`. Always with the same result.

### Regression Issue

- [x] Select this option if this issue appears to be a regression.

### SDK version number

@aws-sdk/client-iot-data-plane@3.928.0 --> @aws-sdk/client-iot-data-plane@3.929.0

### Which JavaScript Runtime is this issue in?

Node.js

### Details of the browser/Node.js/ReactNative version

v22.18.0

### Reproduction Steps

1. Publish a payload to an MQTT topic.
2. Publish `undefined` to the same MQTT topic.
3. Check whether retained topic still exists. --> Should result in `ResourceNotFoundException`

```javascript
import {
  GetRetainedMessageCommand,
  IoTDataPlaneClient,
  PublishCommand
} from '@aws-sdk/client-iot-data-plane';

const client = new IoTDataPlaneClient({
  region: 'eu-central-1'
});

await client.send(
  new PublishCommand({
    payload: JSON.stringify({
      test: 'data'
    }),
    retain: true,
    topic: 'test'
  })
);

console.log(
  'topic exists:',
  await client
    .send(
      new GetRetainedMessageCommand({
        topic: 'test'
      })
    )
    .then(() => true)
    .catch(() => false)
);

await client.send(
  new PublishCommand({
    payload: undefined,
    retain: true,
    topic: 'test'
  })
);

console.log(
  'topic exists:',
  await client
    .send(
      new GetRetainedMessageCommand({
[...]

PR #7501, opened 2025-11-13
fix(core/protocols): rest-json default request body conditions

kuhe:
### Issue
https://github.com/aws/aws-sdk-js-v3/issues/7500

### Description
This makes the preconditions for setting a default REST JSON request body of `{}` more precise and consistent with older behavior.

This relates to operations with httpPayload bindings where a falsy non-byte-array value is being used to represent an empty byte array.

### Testing
- [x] add unit tests
- [x] re-enabled some backlogged protocol tests for REST JSON to increase coverage 

### Checklist

### DIFF
--- a/packages/middleware-sdk-glacier/src/account-id-default.ts
+++ b/packages/middleware-sdk-glacier/src/account-id-default.ts
@@ -11,7 +11,7 @@ export function accountIdDefaultMiddleware(): InitializeMiddleware<any, any> {
   return <Output extends MetadataBearer>(next: InitializeHandler<any, Output>): InitializeHandler<any, Output> =>
     async (args: InitializeHandlerArguments<any>): Promise<InitializeHandlerOutput<Output>> => {
       const { input } = args;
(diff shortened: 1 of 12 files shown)


packages/middleware-sdk-glacier/src/account-id-default.ts
```diff
@@ -11,7 +11,7 @@ export function accountIdDefaultMiddleware(): InitializeMiddleware<any, any> {
   return <Output extends MetadataBearer>(next: InitializeHandler<any, Output>): InitializeHandler<any, Output> =>
     async (args: InitializeHandlerArguments<any>): Promise<InitializeHandlerOutput<Output>> => {
       const { input } = args;
-      if (input.accountId === undefined) {
+      if (!input.accountId) {
```
kuhe, line 14: protocol tests say this should be inserted for both null and empty string, not only null.

smilkuri (approved):

Issue #7470, opened 2025-11-03
Unable to set `filename*=` in CloudFront signer


kuhe:
### Checkboxes for prior research

- [x] I've gone through [Developer Guide](https://docs.aws.amazon.com/sdk-for-javascript/v3/developer-guide) and [API reference](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest)
- [x] I've checked [AWS Forums](https://forums.aws.amazon.com) and [StackOverflow](https://stackoverflow.com/questions/tagged/aws-sdk-js).
- [x] I've searched for [previous similar issues](https://github.com/aws/aws-sdk-js-v3/issues) and didn't find any solution.

### Describe the bug

In https://github.com/aws/aws-sdk-js-v3/pull/7437, the encoding of the URL query parameters uses the extended encoding from `@smithy`, leading to the `*` (asterisk) character being encoded. 

The `*` character has a practical application in the query parameter -- see comment in https://github.com/aws/aws-sdk-js-v3/pull/7437

A related ticket was not properly resolved in https://github.com/aws/aws-sdk-js-v3/issues/3614. 

### Regression Issue

- [x] Select this option if this issue appears to be a regression.

### SDK version number

@aws-sdk/package-name@version, ...

### Which JavaScript Runtime is this issue in?

Node.js

### Details of the browser/Node.js/ReactNative version

all

### Reproduction Steps

get a signed url with a query parameter such as 

```js
`?response-content-disposition=${encodeURIComponent(
  `attachment; filename=${encName}; filename*=UTF-8''${encName}`,
)}`
```

See section on `filename*` from https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Disposition.

### Observed Behavior

signed request fails (likely 403?)

### Expected Behavior

signed request succeeds

### Possible Solution

use `encodeURIComponent` instead of `extendedEncodeURIComponent`

benclarkmck:
this has also broken our use case, where brackets in a query parameter are getting encoded where they weren't before.

something like:

```js
import { getSignedUrl } from '@aws-sdk/cloudfront-signer';
import { cloudfront as cloudfrontConfig } from 'config';

const { keyPairId, privateKey } = cloudfrontConfig;

const url = 'https://some-cloudfront-url';
const originalFilename=encodeUriComponent('original filename (2025-01-01).pdf');
const date = new Date();
[...]

PR #7515, opened 2025-11-21
fix(cloudfront-signer): skip extended encoding for query parameters in the base url

benclarkmck:
### Issue

#7470
https://github.com/aws/aws-sdk-js-v3/pull/7437#issuecomment-3467437014

### Description

using the extended encoding encodes characters that do not need to be encoded, resulting in invalid signed urls

### Testing

Updated the unit test
Manually tested against an internal app that is currently broken with the latest version

### Additional context
I was not sure if the extended encoding is important or not for the cloudfront parameters that get added, so I've left it there ([line 143](https://github.com/aws/aws-sdk-js-v3/blob/a617d7cd813082fed8f70ea69e1735c466548f3a/packages/cloudfront-signer/src/sign.ts#L143)), I suspect though that it's unnecessary but have no idea how to confirm

### Checklist
- [x] If the PR is a feature, add integration tests (`*.integ.spec.ts`). (N/A)
- [x] If you wrote E2E tests, are they resilient to concurrent I/O? (N/A)
- [x] If adding new public functions, did you add the `@public` tag and enable doc generation on the package? (N/A)

---

### DIFF
--- a/packages/cloudfront-signer/src/sign.spec.ts
+++ b/packages/cloudfront-signer/src/sign.spec.ts
@@ -812,7 +812,7 @@ describe("getSignedUrl- when signing a URL with a date range", () => {
 });
 
 describe("url component encoding", () => {
-  it("should use extended encoding for query params in the base URL", () => {
+  it("should use standard encoding for query params in the base URL", () => {
     const url =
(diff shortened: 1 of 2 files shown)


trivikr:
Build failure is OK. I confirmed that unit tests are successful.

PR #7522, opened 2025-11-24
feat(dynamodb): add batch get item paginators for low level and document client

cipak:
### Description
This pull request adds paginators for the DynamoDB BatchGetItemCommand (low level client) and BatchGetCommand (document client). Similar to the implementation in other AWS SDKs (Java, .Net, Go), these paginators work by repeatedly issuing batch get commands until the returned UnprocessedKeys object is missing (or empty), or a service error is returned.

Unlike the other paginators (paginateQuery and paginateScan), these batch get paginators do not mutate the received input.

### Testing
This change was tested using a DynamoDB table with 100 items, each 400kb in size. The first test was with a provisioned capacity of 5 RCUs, and the second with 10,000 RCUs . In both tests, the paginators returned the 100 items in 7 pages of maximum 17 items each.

### Checklist

---


packages/util-dynamodb/src/paginatedBatchGetItem.ts
```diff
@@ -0,0 +1,40 @@
+import { BatchGetItemCommand, BatchGetItemCommandInput, DynamoDBClient } from "@aws-sdk/client-dynamodb";
```
kuhe: this paginator should be placed with the client rather than util-dynamodb, but named in a way that prevents collisions between it and the generated paginators

perhaps `export async function* customPaginateBatchGetItem(...)`

packages/util-dynamodb/src/paginateBatchGetItem.ts
```diff
@@ -0,0 +1,40 @@
...
+/**
+ * @public
+ *
+ * Async generator that issues {@link BatchGetItemCommand}s repeatedly until all keys are processed or an error response is received.
+ *
+ * @see {@link paginateBatchGetItem} for a variant that uses the {@link DynamoDBDocumentClient | DynamoDB document client}.
+ *
+ * @example
+ *
+ * ```
```
kuhe:
```suggestion
 * ```typescript
```

packages/util-dynamodb/src/paginatedBatchGetItem.ts
```diff
@@ -0,0 +1,40 @@
...
+import { type DynamoDBDocumentClient } from "@aws-sdk/lib-dynamodb";
+
+/**
+ * @public
+ *
+ * Async generator that issues {@link BatchGetItemCommand}s repeatedly until all keys are processed or an error response is received.
```
kuhe, line 5: the description must go above all annotations, like:

```js
/**
 * Description here.
 *
 * @public
 * @see ...
 * @example
 */
```

packages/util-dynamodb/src/paginateBatchGetItem.ts
cipak: done

packages/util-dynamodb/src/paginatedBatchGetItem.ts
cipak: done

packages/util-dynamodb/src/paginatedBatchGetItem.ts
kuhe: as for the naming, we cannot use the same pattern as the generated paginator because it would break backwards compatibility if the type signature changes between the hand-written and generated paginator. 

it can be named something subtly different rather than "customPaginateBatchGetItem", perhaps "paginatedBatchGetItem"

cipak: I renamed the functions as suggested (paginatedBatchGetItem for  the low-level client and paginatedBatchGet for document client).

regarding moving the low-level client paginator from util-dynamodb to client-dynamodb, I'll leave that part to the next developer, as today I am transitioning off the project.

andreiluculescu: Hi @kuhe ! Can you please take a look at latest changes and decide if we are good to go?

kuhe: I don't have capacity to review this right now. Please contact my team internally.

Issue #7357, opened 2025-09-18
@aws-sdk/ec2-metadata-service supressing HTTP statusCode


rynop:
### Checkboxes for prior research

- [x] I've gone through [Developer Guide](https://docs.aws.amazon.com/sdk-for-javascript/v3/developer-guide) and [API reference](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest)
- [x] I've checked [AWS Forums](https://forums.aws.amazon.com) and [StackOverflow](https://stackoverflow.com/questions/tagged/aws-sdk-js).
- [x] I've searched for [previous similar issues](https://github.com/aws/aws-sdk-js-v3/issues) and didn't find any solution.

### Describe the bug

When the SDK makes an internal request to the EC2 instance's meta-data service, if the HTTP request fails the `.statusCode` does not bubble back up to the calling code.

In AWS SDK for JS v2, calling code could get at the `.statusCode` on the thrown `Error`.  In V3, the original error is caught and re-thrown as a string: `Error making request to the metadata service: ${error}`.  For example: `Error making request to the metadata service: Error: Request failed with status code 404`

The offending code is here: https://github.com/aws/aws-sdk-js-v3/blob/main/packages/ec2-metadata-service/src/MetadataService.ts#L80

The failing status code needs to be available to the caller so we don't have to do string matching.

### Regression Issue

- [x] Select this option if this issue appears to be a regression.

### SDK version number

AWS SDK v2

### Which JavaScript Runtime is this issue in?

Node.js

### Details of the browser/Node.js/ReactNative version

node: v22.15.0

### Reproduction Steps

```
import {MetadataService} from '@aws-sdk/ec2-metadata-service';

const metaService = new MetadataService({});
await metaService.request(`/latest/meta-data/DNE`, {});
```

### Observed Behavior

Original error is caught and re-thrown, with the original error flattened to a string. Ex: `Error making request to the metadata service: Error: Request failed with status code 404`

### Expected Behavior

`statusCode` on the failing request should be available to the caller

siddsriv:
A fix was released in [v3.951](https://github.com/aws/aws-sdk-js-v3/releases/tag/v3.951.0). Let us know if something else comes up, thanks!

PR #7564, opened 2025-12-10
fix(ec2-metadata-service): preserve statusCode when rethrowing errors for IMDS requests

siddsriv:
### Issue
#7357 

### Description
Bubbles up status code for users to access in case of errors when making an IMDS `request()`.

### Testing
Minimal repro:
```js
import { MetadataService } from '@aws-sdk/ec2-metadata-service';

try {
    const metaService = new MetadataService({});
    await metaService.request(`/latest/meta-data/DNE`, {});
} catch (e) {
    console.log(e.$metadata);
    console.log(e.$metadata?.httpStatusCode)
    console.log(e.message)
}

```

Before the fix:
```console
undefined
undefined
Error making request to the metadata service: Error: Request failed with status code 404
```

After the fix:
```console
{ httpStatusCode: 404 }
404
Error making request to the metadata service: Error: Request failed with status code 404
```

Unit tests in the package are passing.

---

### DIFF
--- a/packages/ec2-metadata-service/src/MetadataService.ts
+++ b/packages/ec2-metadata-service/src/MetadataService.ts
@@ -77,13 +77,14 @@ export class MetadataService {
         return sdkStreamMixin(response.body).transformToString();
       } else {
         throw Object.assign(new Error(`Request failed with status code ${response.statusCode}`), {
-          statusCode: response.statusCode,
+          $metadata: { httpStatusCode: response.statusCode },
         });
       }
     } catch (error) {
       const wrappedError = new Error(`Error making request to the metadata service: ${error}`);
-      if ((error as any).statusCode !== undefined) {
-        (wrappedError as any).statusCode = (error as any).statusCode;
+      const { $metadata } = error as any;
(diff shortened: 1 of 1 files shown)


smilkuri (approved):

packages/ec2-metadata-service/src/MetadataService.ts
```diff
@@ -76,10 +76,16 @@ export class MetadataService {
...
+        throw Object.assign(new Error(`Request failed with status code ${response.statusCode}`), {
+          statusCode: response.statusCode,
+        });
       }
     } catch (error) {
-      throw new Error(`Error making request to the metadata service: ${error}`);
+      const wrappedError = new Error(`Error making request to the metadata service: ${error}`);
+      if ((error as any).statusCode !== undefined) {
+        (wrappedError as any).statusCode = (error as any).statusCode;
+      }
```
kuhe, line 88: this part would look better as

```ts
const { statusCode } = error as any;
if (statusCode !== undefined) {
  Object.assign(wrappedError, { statusCode });
}
```

PR #7569, opened 2025-12-11
feat(ec2-metadata-service): add retries for IMDS requests

siddsriv:
### Issue
Internal JS-5193

### Description
Support retries for EC2 IMDS requests and fetching tokens.

### Testing
Unit tests:
```console
 ✓ src/MetadataService.spec.ts (24 tests) 20ms
   ✓ MetadataService Socket Leak Checks > fetchMetadataToken - body consumption checks > should consume response body on 200 status and return token 5ms
   ✓ MetadataService Socket Leak Checks > fetchMetadataToken - body consumption checks > should consume response body on 400 status before throwing 2ms
.
.
.

   ✓ MetadataService Retry Configuration > status code handling for retries > should not retry 403 errors 0ms
   ✓ MetadataService Retry Configuration > status code handling for retries > should not retry 404 errors 0ms
   ✓ MetadataService Retry Configuration > status code handling for retries > should retry 401 errors 0ms
   ✓ MetadataService Retry Configuration > status code handling for retries > should retry 500 errors 0ms

 Test Files  1 passed (1)
      Tests  24 passed (24)
   Start at  23:37:49
   Duration  354ms (transform 72ms, setup 0ms, collect 96ms, tests 20ms, environment 0ms, prepare 78ms)
```

---

### DIFF
--- a/packages/ec2-metadata-service/src/MetadataServiceOptions.ts
+++ b/packages/ec2-metadata-service/src/MetadataServiceOptions.ts
@@ -33,6 +33,8 @@ export interface MetadataServiceOptions {
   retries?: number;
   /**
    * the number of seconds to sleep in-between retries and/or a customer provided backoff function to call.
+   * if the function returns a promise, it will be awaited and its resolved value ignored.
+   * if the function returns a number, the number will be used as seconds duration to wait before the following retry attempt.
    */
   backoff?: number | ((numFailures: number) => Promise<void> | number);
 }


packages/ec2-metadata-service/src/MetadataService.ts
```diff
@@ -124,11 +181,14 @@ export class MetadataService {
         return bodyString;
       } else {
         throw Object.assign(new Error(`Failed to fetch metadata token with status code ${response.statusCode}`), {
-          statusCode: response.statusCode,
+          $metadata: { httpStatusCode: response.statusCode },
```
siddsriv, line 192: slightly outside the scope of retries, but this is for consistency with the status code handling/assignment in `request()`

packages/ec2-metadata-service/src/MetadataService.spec.ts
```diff
@@ -24,6 +24,7 @@ describe("MetadataService Socket Leak Checks", () => {
     metadataService = new MetadataService({
       endpoint: "http://169.254.169.254",
       httpOptions: { timeout: 1000 },
+      retries: 0, // Disable retries for faster tests
```
kuhe, line 27: should be maxAttempts and retryStrategy like the clients

kuhe (approved):

packages/ec2-metadata-service/src/MetadataService.ts
```diff
@@ -13,6 +13,8 @@ import { MetadataServiceOptions } from "./MetadataServiceOptions";
 export class MetadataService {
   private disableFetchToken: boolean;
   private config: Promise<MetadataServiceOptions>;
+  private retries: number;
+  private backoffFn: (numFailures: number) => void;
```
kuhe: should be `(numFailures: number) => Promise<void> | number`

```ts
const backoff = this.backoffFn(failures);
if (typeof backoff === 'number') {
  await new Promise(r => setTimeout(r, backoff));
} else {
  await backoff;
}
```

packages/ec2-metadata-service/src/MetadataService.ts
```diff
@@ -30,68 +32,123 @@ export class MetadataService {
...
+        if (attempt === this.retries) {
+          break;
         }
-        // If token fetch fails and IMDSv1 is not disabled, proceed without token (IMDSv1 fallback)
+
+        if (this.shouldNotRetry(error as any)) {
+          throw error;
+        }
+
+        await this.backoffFn(attempt);
```
kuhe: awaiting a function that synchronously returns does not cause a delay

packages/ec2-metadata-service/src/MetadataService.spec.ts
siddsriv: after discussion: leaving these as is owing to the spec directive

packages/ec2-metadata-service/src/MetadataServiceOptions.ts
```diff
@@ -34,5 +34,5 @@ export interface MetadataServiceOptions {
   /**
    * the number of seconds to sleep in-between retries and/or a customer provided backoff function to call.
```
kuhe, line 35: docs should say:

> If the function returns a promise, it will be awaited and its resolved value ignored.
> If the function returns a number, the number will be used as a millisecond duration to wait before the following retry attempt.

packages/ec2-metadata-service/src/MetadataService.ts
```diff
@@ -69,7 +72,12 @@ export class MetadataService {
           throw error;
         }
 
-        await this.backoffFn(attempt);
+        const backoffResult = this.backoffFn(attempt);
+        if (typeof backoffResult === "number") {
+          await this.sleep(backoffResult * 1000); // seconds to milliseconds
```
kuhe, line 77: use milliseconds for time, since that is the time unit idiomatic to the language? how is the specification on this?

siddsriv: spec and the [docs](https://docs.aws.amazon.com/sdkref/latest/guide/feature-imds-client.html) here indicate seconds as the ground truth

packages/ec2-metadata-service/src/MetadataServiceOptions.ts
```diff
@@ -27,4 +27,14 @@ export interface MetadataServiceOptions {
    * when true, metadata service will not fetch token, which indicates usage of IMDSv1
    */
   disableFetchToken?: boolean;
+  /**
+   * the number of retry attempts for any failed request, defaulting to 3.
```
kuhe, line 31: default attempt count = 4 then?

siddsriv: yes, 1 initial + 3 retries (by default)

PR #7580, opened 2025-12-15
chore: build cjs rollup without typescript

kuhe:
### Issue
n/a

Review diff without JSON files: https://github.com/aws/aws-sdk-js-v3/pull/7580/files?file-filters%5B%5D=.js&show-viewed-files=true

### Description
This changes the build pipeline of cjs artifacts.

- rollup-plugin-typescript has been removed from the CJS build.
- because dist-es is a perfect mirror of the typescript sources, we can use dist-es as the input to build dist-cjs, instead of the source files

The motivation for this comes from the build speed of rollup-plugin-typescript. They do not allow an option to skip typechecking. Although `noCheck` is set as a compiler option, this is not compatible with how they integrate typescript into the rollup build. This slows the build significantly -- e.g. the core package needs 20s to build cjs. 

With this change rollup can build the core package's dist-es -> dist-cjs in 4s. Typechecking coverage is applied during the build of `dist-types`, so it is redundant in the CJS process. Indeed, `noCheck: true` is already in effect for `build:es`.

### Testing
CI - because e2e runs against the dist-cjs files, it gives high confidence that the change is acceptable.

I also manually verified that the dist-cjs artifacts generated in this way do not differ from what is currently on npmjs.org with both submodule-style packages (core) and non-submodule packages (client-s3).


kuhe:
comparing 0-cached builds, this reduces the build from 8m7s to 6m23s

PR #7584, opened 2025-12-17
fix(ec2-metadata-service): add configurable options for ttl and port precedence

siddsriv:
### Issue
Internal JS-5193

### Description
makes `tokenTtl` configurable and sets a precedence order for `port`.

### Testing
unit testing:
```console
✓ src/MetadataService.spec.ts (31 tests) 23ms
   ✓ MetadataService Socket Leak Checks > fetchMetadataToken - body consumption checks > should consume response body on 200 status and return token 5ms
   ✓ MetadataService Socket Leak Checks > fetchMetadataToken - body consumption checks > should consume response body on 400 status before throwing 2ms
.
.
.
   ✓ MetadataService Retry Configuration > status code handling for retries > should retry 500 errors 0ms
   ✓ MetadataService Token TTL Configuration > should use default tokenTtl of 21600 seconds 0ms
   ✓ MetadataService Token TTL Configuration > should use custom tokenTtl value 0ms
   ✓ MetadataService Token TTL Configuration > should validate tokenTtl as positive integer 1ms
   ✓ MetadataService Token TTL Configuration > should accept valid positive integer tokenTtl values 0ms
   ✓ MetadataService Token TTL Configuration > should convert tokenTtl to string in header 0ms

 Test Files  1 passed (1)
      Tests  31 passed (31)
```

---

### DIFF
--- a/packages/ec2-metadata-service/src/MetadataService.ts
+++ b/packages/ec2-metadata-service/src/MetadataService.ts
@@ -16,7 +16,7 @@ export class MetadataService {
   private retries: number;
   private backoffFn: (numFailures: number) => Promise<void> | number;
   private tokenTtl: number;
-  // can be set explicitly, or extracted from endpoint, or use a default value (80). See `resolvePort()` below.
+  // can be set explicitly, or extracted from endpoint, or use a default value. See `resolvePort()` below.
   private port?: number;
 
   /**


packages/ec2-metadata-service/src/MetadataServiceOptions.ts
```diff
@@ -37,4 +37,13 @@ export interface MetadataServiceOptions {
    * if the function returns a number, the number will be used as seconds duration to wait before the following retry attempt.
    */
   backoff?: number | ((numFailures: number) => Promise<void> | number);
+  /**
+   * the TTL of the token in seconds, defaulting to 21,600 seconds (6 hours)
+   */
+  tokenTtl?: number;
+  /**
+   * the port for the endpoint, defaulting to 80.
```
kuhe: shouldn't it default to undefined or the default port of the url's protocol?

siddsriv: discussed and updated!

kuhe (approved):

PR #7599, opened 2025-12-22
chore(build): generate clients without formatting prettier/eslint

kuhe:
generate clients and protocol tests (private) without external application of formatting/linting

timing:
- main branch: 9m19.305s
- PR: 2m33.212s


clients/client-s3/src/commands/CreateBucketMetadataConfigurationCommand.ts
```diff
@@ -190,7 +190,7 @@ export class CreateBucketMetadataConfigurationCommand extends $Command
     return [
       getEndpointPlugin(config, Command.getEndpointParameterInstructions()),
       getFlexibleChecksumsPlugin(config, {
-        requestAlgorithmMember: { httpHeader: "x-amz-sdk-checksum-algorithm", name: "ChecksumAlgorithm" },
+        requestAlgorithmMember: {'httpHeader': 'x-amz-sdk-checksum-algorithm', 'name': 'ChecksumAlgorithm'},
```
kuhe, line 193: I couldn't find where the single quotes are coming from. 🙃

PR #7612, opened 2026-01-01
feat(codegen): add Command base class documentation (partial fix for …

gabrielfranca95:
**Description**

This PR adds a `TypeScriptIntegration` that generates a `CommandBase.ts` file in each client's `src/commands/` directory, providing documentation explaining what the Command base class is and why `*_base` types appear in `.d.ts` files.

**What does this PR do?**

This PR adds a `CommandBase.ts` file that:

* Re-exports the `Command` class from `@smithy/smithy-client`
* Includes comprehensive JSDoc explaining what `*_base` types represent
* Provides links to the official `@smithy/smithy-client` documentation

**⚠️ This is a Partial Fix**

This PR does not modify the JSDoc of each Command class directly. Users will still see:

```typescript
export class ListBucketsCommand extends ListBucketsCommand_base

```

They will need to navigate to the `CommandBase.ts` documentation to understand what the base class is.


kuhe:
This doesn't address the issue

Issue #4271, opened 2022-12-12
stabilise remaining @aws-sdk/* packages and generate clients with ^ ranges


everett1992:
### Describe the feature

The majority of @aws-sdk/* dependencies shared between aws-sdk (and smithy-typescript) clients [are documented as 'internal' stability](https://github.com/aws/aws-sdk-js-v3#stability-of-modular-packages). 

One effect of the instable api is that clients are generated with exact dependencies[^1][^2]. This causes a number of problems.
 * If users install different patch versions of aws-sdk clients they will have multiple versions of @aws-sdk dependencies. Users have to be careful to update all aws-sdk clients at the same time. At least aws-sdk coordinates releases so there are matching versions with the same dependencies.
 * Users of generic smithy-typescript clients will have a hard time matching versions of their generic clients and aws-sdk clients because there is no one to coordinate the versions. smithy-typescript-codegen and smithy-typescript-aws-codegen do not have release automation and the latest code is not always available from maven. All generic clients are not owned by the same group so package may have one set of @aws-sdk dependencies for aws clients and one set of @aws-sdk dependencie for *each* generic client.

---
[^1]: Example of aws-sdk client dependencies
```
    $ npm info @aws-sdk/client-accessanalyzer dependencies
    {
      '@aws-crypto/sha256-browser': '2.0.0',
      '@aws-crypto/sha256-js': '2.0.0',
      '@aws-sdk/client-sts': '3.226.0',
      '@aws-sdk/config-resolver': '3.226.0',
      '@aws-sdk/credential-provider-node': '3.226.0',
      '@aws-sdk/fetch-http-handler': '3.226.0',
      '@aws-sdk/hash-node': '3.226.0',
      '@aws-sdk/invalid-dependency': '3.226.0',
      '@aws-sdk/middleware-content-length': '3.226.0',
      '@aws-sdk/middleware-endpoint': '3.226.0',
      '@aws-sdk/middleware-host-header': '3.226.0',
      '@aws-sdk/middleware-logger': '3.226.0',
      '@aws-sdk/middleware-recursion-detection': '3.226.0',
      '@aws-sdk/middleware-retry': '3.226.0',
      '@aws-sdk/middleware-serde': '3.226.0',
      '@aws-sdk/middleware-signing': '3.226.0',
      '@aws-sdk/middleware-stack': '3.226.0',
      '@aws-sdk/middleware-user-agent': '3.226.0',
[...]

PR #7635, opened 2026-01-08
chore: customize lerna versioning actions

kuhe:
This PR customizes two build processes:
- bumping package versions during release
- generating the list of packages to release to NPM

### bumping packages
`yarn lerna:version`, which then runs `lerna version`

### generating manifest
because everything is on a unified version, the manifest is the output of `lerna list` filtered by packages that are at the latest version (because they were bumped by lerna change detection).

## Changes in PR:
The PR introduces the concept of multiple lerna.json configurations. Because lerna does not accept a config file path as an option, the scripts overwrite the root `lerna.json` as needed to temporarily configure different subsets of packages for execution of lerna's `list`, `changed`, and `version` operations.

There are two configurations:
1. the linked package set. This includes all packages that will continue to release at the unified release version number, currently in the `3.960.0 - 3.970.0` range. 
2. the independent package set. The PR only introduces one package to this set, but in the future this will contain all packages that operate with open version ranges. 

### bumping packages
`lerna version` is executed on the two package sets

### generating manifest
`lerna list` is executed on the linked package set and `lerna changed` is executed on the independent package set

----

### changes to developer workflow
Fortunately, this allows the developer workflow to remain the same. e2e tests can continue to utilize all packages without needing special linking, and developers do **not** need to manually bump any package.json versions except when making major version increments to the independent package set (rare). 

For independent packages, care must be taken to remain backwards compatible when making API changes. This is already the case even with pinned packages due to the way Lambda runtimes deployment of updates.

### what this ultimately will enable
As more packages are moved into the independent versioning set, consumers of those packages will be able to declare `^x.y.z` open version ranges for those dependencies, resolving longstanding customer issues relating to pinned interdependency versions such as https://github.com/aws/aws-sdk-js-v3/issues/4271.

### DIFF
--- a/packages-internal/util-locate-window/CHANGELOG.md
+++ b/packages-internal/util-locate-window/CHANGELOG.md
@@ -3,8 +3,6 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
-
-
 # [3.965.0](https://github.com/aws/aws-sdk-js-v3/compare/v3.964.0...v3.965.0) (2026-01-07)
 
 **Note:** Version bump only for package @aws-sdk/util-locate-window


packages-internal/util-locate-window/CHANGELOG.md
```diff
@@ -3,6 +3,8 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
+
+
```
smilkuri: Is this intentional?

kuhe: it shouldn't make a difference when formatted later, but I'll remove it

trivikr (approved):

Issue #7636, opened 2026-01-08
S3 client cannot call PutObject with 0-length Readable stream


kuhe:
### Checkboxes for prior research

- [x] I've gone through [Developer Guide](https://docs.aws.amazon.com/sdk-for-javascript/v3/developer-guide) and [API reference](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest)
- [x] I've checked [AWS Forums](https://forums.aws.amazon.com) and [StackOverflow](https://stackoverflow.com/questions/tagged/aws-sdk-js).
- [x] I've searched for [previous similar issues](https://github.com/aws/aws-sdk-js-v3/issues) and didn't find any solution.

### Describe the bug

```js
await s3.putObject({ Bucket: BUCKET, Key: 'empty-from-buffer', Body: Readable.from(Buffer.from("")), ContentLength: 0 });
```

### Regression Issue

- [ ] Select this option if this issue appears to be a regression.

### SDK version number

@aws-sdk/client-s3

### Which JavaScript Runtime is this issue in?

Node.js

### Details of the browser/Node.js/ReactNative version

-

### Reproduction Steps

-

### Observed Behavior

```
S3ServiceException [MalformedTrailerError]: The request contained trailing data that was not well-formed or did not conform to our published schema.
    at ProtocolLib.getErrorSchemaOrThrowBaseException (/node_modules/@aws-sdk/core/dist-cjs/submodules/protocols/index.js:69:67)
    at AwsRestXmlProtocol.handleError (/node_modules/@aws-sdk/core/dist-cjs/submodules/protocols/index.js:1795:65)
    at AwsRestXmlProtocol.deserializeResponse (/node_modules/@smithy/core/dist-cjs/submodules/protocols/index.js:301:24)
    at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
    at async /node_modules/@smithy/core/dist-cjs/submodules/schema/index.js:26:24
    at async /node_modules/@aws-sdk/middleware-sdk-s3/dist-cjs/index.js:386:20
    at async /node_modules/@smithy/middleware-retry/dist-cjs/index.js:254:46
    at async /node_modules/@aws-sdk/middleware-flexible-checksums/dist-cjs/index.js:241:24
    at async /node_modules/@aws-sdk/middleware-sdk-s3/dist-cjs/index.js:63:28
[...]

PR #7641, opened 2026-01-10
Main

Sapientia-PT:
### Issue
#7636

### Description
S3 client was unable to call PutObject with 0-length Readable stream due to
aws-chunked encoding being applied even for zero-length bodies, causing
MalformedTrailerError.

The fix detects zero-length bodies by checking the content-length header and
bypasses aws-chunked encoding (which uses HTTP trailers) for them. Instead,
it computes the checksum over an empty payload and sends only a checksum
header without trailer headers.

This allows S3 PutObject to work correctly with empty Readable streams while
maintaining the existing behavior for non-zero-length streams and non-streaming
bodies.

### Testing
Extended unit tests to cover this edge case

### Additional context

### Checklist

---

### DIFF
--- a/packages/middleware-flexible-checksums/src/flexibleChecksumsMiddleware.spec.ts
+++ b/packages/middleware-flexible-checksums/src/flexibleChecksumsMiddleware.spec.ts
@@ -190,6 +190,35 @@ describe(flexibleChecksumsMiddleware.name, () => {
       });
     });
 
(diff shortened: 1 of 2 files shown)


kuhe:
This will be fixed by https://github.com/smithy-lang/smithy-typescript/pull/1838, which includes handling zero size chunks.

PR #7642, opened 2026-01-12
chore: fix build versioning

kuhe:
### Issue
P367504818

### Description
The new release script missed updating the workspace lerna.json version prior to committing the versioning changes.

### Testing
Ran `yarn lerna:version` locally.

### Checklist


lerna.json
```diff
@@ -1,5 +1,5 @@
 {
-  "version": "3.966.0",
+  "version": "3.967.0",
```
trivikr: Should this be `3.966.0`, and it'll be updated to `3.967.0` during release?

trivikr: To simplify retrying the build and release, we reverted the release commits through force-push and submitted this fix. Details in P367504818.

PR #7666, opened 2026-01-16
chore(scripts): override smithy partitions

siddsriv:
### Issue
Internal JS-6400

### Description
Env override for partitions, to use the copy that exists in the SDK and is updated by awstools instead of the partitions that come bundled with smithy.

### Testing
clients are generated successfully

---

### DIFF
--- a/scripts/utils/spawn-process.js
+++ b/scripts/utils/spawn-process.js
@@ -1,7 +1,34 @@
 // @ts-check
 const { spawn } = require("node:child_process");
+const { join } = require("node:path");
 
 const spawnProcess = async (command, args = [], options = {}) => {
+  // set AWS_PARTITIONS_FILE_OVERRIDE for gradle commands to ensure
+  // Smithy uses the SDK's up-to-date partitions.json instead of its bundled version.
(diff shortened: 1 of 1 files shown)


scripts/generate-clients/code-gen.js
```diff
@@ -31,7 +31,23 @@ const generateClient = async (clientName) => {
...
+        "..",
+        "packages",
+        "util-endpoints",
+        "src",
+        "lib",
+        "aws",
+        "partitions.json"
+      ),
+    },
+  });
```
kuhe: too repetitive, bake this into spawnProcess?

scripts/utils/spawn-process.js
```diff
@@ -1,7 +1,34 @@
...
+      options = {
+        ...options,
+        env: {
+          ...process.env,
+          ...(options.env || {}),
+          AWS_PARTITIONS_FILE_OVERRIDE: join(
+            __dirname,
+            "..",
+            "..",
+            "packages",
```
kuhe:
```suggestion
            "packages-internal",
```

Issue #6015, opened 2020-03-10
Official SQS poller for JavaScript


nathanpeck:
SDK's for other languages such as Ruby have an automatic queue poller class: https://docs.aws.amazon.com/sdk-for-ruby/v2/api/Aws/SQS/QueuePoller.html

Node currently does not. There are several community created pollers such as Squiss, but these are unmaintained: https://www.npmjs.com/package/squiss

We need an official poller class, built into the AWS SDK for JS, which allows me to just start polling, and have a function in my code get called whenever there is a message to receive.

PR #7677, opened 2026-01-22
feat(lib-sqs): RFC for Official SQS Poller API Design

jameslevine:
### Issue
#6015

### Description
**RFC: Official SQS Poller API Design - We Want Your Feedback!**

We're actively working on an **official SQS Poller** for the AWS SDK for JavaScript. This has been a long-requested feature, and we want to make sure we get the API design right.

This PR contains the proposed API signatures for review (see `lib/lib-sqs/index.ts`).

**We Need Your Feedback:**

1. **Does this API meet your use cases?** Are there scenarios we haven't considered?
2. **Configuration options** - Are the defaults sensible? Any missing options?
3. **Event names and signatures** - Are they intuitive and useful?
4. **Error handling** - Does the error class hierarchy make sense?
5. **Migration path** - If you're using existing community solutions, would this be easy to migrate to?

We're committed to addressing the reliability issues that have affected community solutions (like the polling stoppages mentioned in #6015). Your input will help us build a robust, production-ready solution.

*Note: This is a Request for Comments (RFC). The final implementation may differ based on feedback received.*

### Testing
N/A - This is an RFC containing only API signatures with stubbed implementations for review purposes.

### Additional context
This RFC was developed in collaboration with the AWS SDK team. The API file contains TypeScript signatures with `throw new Error("not implemented")` stubs, following the format requested for API review.

### Checklist

---

### DIFF
--- a/lib/lib-sqs/index.ts
+++ b/lib/lib-sqs/index.ts
@@ -7,44 +7,47 @@
  * @packageDocumentation
  */
 
-import { EventEmitter } from "events";
-import type { SQSClient } from "@aws-sdk/client-sqs";
-import type { Credentials } from "@aws-sdk/types";
+import type {
+  SQSClient,
+  SQSClientConfig,
+  Message as SQSRawMessage,
+  MessageAttributeValue,
+  MessageSystemAttributeName,
+  BatchResultErrorEntry,
+} from "@aws-sdk/client-sqs";
+
+// ============================================================
+// Re-exports from @aws-sdk/client-sqs
+// ============================================================
+
+/**
+ * Re-export MessageAttributeValue from @aws-sdk/client-sqs for convenience.
+ * @public
+ */
+export type { MessageAttributeValue };
+
+/**
+ * Re-export MessageSystemAttributeName from @aws-sdk/client-sqs for convenience.
+ * @public
+ */
+export type { MessageSystemAttributeName };
+
+/**
+ * Re-export the raw SQS Message type from @aws-sdk/client-sqs.
+ * This is the structure returned directly from the SQS ReceiveMessage API.
+ * @public
+ */
+export type { SQSRawMessage };
+
+/**
+ * Re-export BatchResultErrorEntry from @aws-sdk/client-sqs for deletion failure details.
+ * @public
+ */
+export type { BatchResultErrorEntry };
 
 // ============================================================
 // Type Definitions
 // ============================================================
-
-/**
- * SQS message system attribute names.
- * @public
- */
-export type AttributeName =
-  | "All"
-  | "ApproximateFirstReceiveTimestamp"
-  | "ApproximateReceiveCount"
-  | "SenderId"
-  | "SentTimestamp"
-  | "MessageDeduplicationId"
-  | "MessageGroupId"
-  | "SequenceNumber";
-
-/**
- * Message attribute value structure.
- * @public
- */
-export interface MessageAttributeValue {
-  /** The data type of the attribute value. */
-  DataType: string;
-  /** String value of the attribute. */
-  StringValue?: string;
-  /** Binary value of the attribute. */
-  BinaryValue?: Buffer;
-  /** List of string values. */
-  StringListValues?: string[];
-  /** List of binary values. */
-  BinaryListValues?: Buffer[];
-}
 
 /**
  * Map of message attribute names to values.
@@ -53,110 +56,11 @@ export interface MessageAttributeValue {
 export type MessageAttributes = Record<string, MessageAttributeValue>;
 
 /**
- * Message system attributes returned by SQS.
+ * JSON-serializable representation of poller statistics.
+ * All Date fields are converted to ISO strings for serialization.
  * @public
  */
-export interface MessageSystemAttributes {
-  /** Timestamp of when the message was first received. */
-  ApproximateFirstReceiveTimestamp?: string;
-  /** Number of times the message has been received. */
-  ApproximateReceiveCount?: string;
-  /** AWS account ID or IAM identity of the sender. */
-  SenderId?: string;
-  /** Timestamp of when the message was sent. */
-  SentTimestamp?: string;
-  /** Deduplication ID for FIFO queues. */
-  MessageDeduplicationId?: string;
-  /** Message group ID for FIFO queues. */
-  MessageGroupId?: string;
-  /** Sequence number for FIFO queues. */
-  SequenceNumber?: string;
-}
-
-/**
- * Raw SQS message structure.
- * @public
- */
-export interface SQSMessage {
-  /** Unique message identifier. */
-  MessageId: string;
-  /** Receipt handle for message operations. */
-  ReceiptHandle: string;
-  /** MD5 digest of the message body. */
-  MD5OfBody: string;
-  /** The message body. */
-  Body: string;
-  /** System attributes of the message. */
-  Attributes?: MessageSystemAttributes;
-  /** Custom message attributes. */
-  MessageAttributes?: MessageAttributes;
-  /** MD5 digest of the message attributes. */
-  MD5OfMessageAttributes?: string;
-}
-
-/**
- * Result of sending a single message.
- * @public
- */
-export interface SendMessageResult {
-  /** Unique identifier for the message. */
-  MessageId: string;
-  /** MD5 digest of the message body. */
-  MD5OfMessageBody: string;
-  /** MD5 digest of the message attributes. */
-  MD5OfMessageAttributes?: string;
-  /** Sequence number for FIFO queues. */
-  SequenceNumber?: string;
-}
-
-/**
- * Entry in a successful batch send result.
- * @public
- */
-export interface SendMessageBatchResultEntry {
-  /** Identifier for the message in the batch. */
-  Id: string;
-  /** Unique identifier assigned by SQS. */
-  MessageId: string;
-  /** MD5 digest of the message body. */
-  MD5OfMessageBody: string;
-  /** MD5 digest of the message attributes. */
-  MD5OfMessageAttributes?: string;
-  /** Sequence number for FIFO queues. */
-  SequenceNumber?: string;
-}
-
-/**
- * Entry in a failed batch operation result.
- * @public
- */
-export interface BatchResultErrorEntry {
-  /** Identifier for the message in the batch. */
-  Id: string;
-  /** Whether the error was the sender's fault. */
-  SenderFault: boolean;
-  /** Error code. */
-  Code: string;
-  /** Error message. */
-  Message: string;
-}
-
-/**
- * Result of sending multiple messages.
- * @public
- */
-export interface SendMessageBatchResult {
-  /** Successfully sent messages. */
-  Successful: SendMessageBatchResultEntry[];
-  /** Failed message sends. */
-  Failed: BatchResultErrorEntry[];
-}
-
-/**
- * JSON representation of poller statistics.
- * @public
- */
-export interface PollerStatsJSON {
+export interface SerializablePollerStats {
   /** Total number of receive requests made. */
   requestCount: number;
   /** Total number of messages received. */
@@ -165,60 +69,16 @@ export interface PollerStatsJSON {
   deletedMessageCount: number;
   /** Total number of errors encountered. */
   errorCount: number;
-  /** ISO timestamp of last received message. */
+  /** ISO timestamp of last received message, or null if no messages received. */
   lastMessageReceivedAt: string | null;
   /** ISO timestamp when polling started. */
   pollingStartedAt: string;
-  /** ISO timestamp when polling stopped. */
+  /** ISO timestamp when polling stopped, or null if still running. */
   pollingStoppedAt: string | null;
   /** Average messages per request. */
   averageMessagesPerRequest: number;
   /** Current polling duration in milliseconds. */
   pollingDurationMs: number;
-}
-
-/**
- * Information about a failed message deletion.
- * @public
- */
-export interface DeleteFailure {
-  /** Message ID that failed to delete. */
-  id: string;
-  /** Whether the failure was the sender's fault. */
-  senderFault: boolean;
-  /** Error code. */
-  code: string;
-  /** Error message. */
-  message: string;
-}
-
-/**
- * AWS SDK configuration options.
- * @public
- */
-export interface AWSConfig {
-  /**
-   * AWS region.
-   * @example 'us-east-1'
-   */
-  region?: string;
-  /** AWS credentials. */
-  credentials?: Credentials;
-  /**
-   * Custom endpoint URL (for testing with LocalStack, etc.).
-   * @example 'http://localhost:4566'
-   */
-  endpoint?: string;
-  /**
-   * Maximum number of retry attempts.
-   * @defaultValue 3
-   */
-  maxRetries?: number;
-  /**
-   * Request timeout in milliseconds.
-   * @defaultValue 120000
-   */
-  requestTimeout?: number;
 }
 
 /**
@@ -240,12 +100,16 @@ export interface PollerConfig {
   // AWS Configuration
   // ============================================================
 
-  /** AWS SDK configuration options. */
-  awsConfig?: AWSConfig;
+  /**
+   * Configuration options for the SQS client.
+   * Uses the standard SQSClientConfig from @aws-sdk/client-sqs.
+   * Ignored if `sqs` is provided.
+   */
+  clientConfig?: SQSClientConfig;
 
   /**
    * Pre-configured SQS client instance.
-   * If provided, awsConfig is ignored.
+   * If provided, clientConfig is ignored.
    */
   sqs?: SQSClient;
 
@@ -262,18 +126,27 @@ export interface PollerConfig {
 
   /**
    * Number of messages to receive per request (1-10).
+   * Values outside this range will throw a PollerError during construction.
+   * This limit is enforced by the SQS ReceiveMessage API.
    * @defaultValue 10
+   * @throws {PollerError} If value is not between 1 and 10 (inclusive).
    */
   receiveBatchSize?: number;
 
   /**
-   * Long polling wait time in seconds (0-20).
-   * @defaultValue 20
+   * Long polling wait time in milliseconds (0-20000).
+   * The poller converts this to seconds for the SQS API.
+   * Use 0 for short polling (not recommended for production).
+   * @defaultValue 20000 (20 seconds)
    */
-  receiveWaitTimeSecs?: number;
+  receiveWaitTimeMs?: number;
 
   /**
    * Milliseconds to wait between polls when queue is active.
+   * Set to 0 for no additional delay - the next poll starts immediately
+   * after the previous one completes.
+   * Note: Setting both this and receiveWaitTimeMs to 0 will result in
+   * rapid short-polling requests (not recommended due to cost and rate limiting).
    * @defaultValue 0
    */
   activePollIntervalMs?: number;
@@ -291,48 +164,52 @@ export interface PollerConfig {
   pollRetryMs?: number;
 
   /**
-   * Stop polling after this many seconds of no messages.
+   * Stop polling after this many milliseconds of no messages.
    * Set to null to poll indefinitely.
    * @defaultValue null
    */
-  idleTimeout?: number | null;
+  idleTimeoutMs?: number | null;
 
   // ============================================================
   // Message Handling Configuration
   // ============================================================
 
   /**
-   * Visibility timeout in seconds for received messages.
+   * Visibility timeout in milliseconds for received messages.
+   * The poller converts this to seconds for the SQS API.
    * If not specified, uses the queue's default setting.
    */
-  visibilityTimeoutSecs?: number;
+  visibilityTimeoutMs?: number;
 
   /**
    * Automatically extend message visibility timeout while processing.
+   * When enabled, the poller will periodically extend the visibility timeout
+   * of messages that are still being processed.
    * @defaultValue false
    */
-  autoExtendTimeout?: boolean;
+  extendTimeout?: boolean;
 
   /**
-   * Milliseconds before timeout to make extension call.
-   * Only used if autoExtendTimeout is true.
+   * Milliseconds before the visibility timeout expires to make the extension call.
+   * Only used if extendTimeout is true.
    * @defaultValue 5000
    */
-  advancedCallMs?: number;
+  extendTimeoutAdvanceMs?: number;
 
   /**
-   * Stop extending timeout after this many seconds.
-   * Only used if autoExtendTimeout is true.
-   * @defaultValue 43200 (12 hours)
+   * Maximum total time in milliseconds to extend the visibility timeout.
+   * After this duration, the poller will stop extending the timeout.
+   * Only used if extendTimeout is true.
+   * @defaultValue 43200000 (12 hours)
    */
-  noExtensionsAfterSecs?: number;
+  maxTimeoutExtensionMs?: number;
 
   /**
-   * Skip automatic message deletion.
-   * If true, messages must be manually deleted.
-   * @defaultValue false
+   * Automatically delete messages after they are handled.
+   * If false, messages must be manually deleted by calling message.enqueueDelete().
+   * @defaultValue true
    */
-  skipDelete?: boolean;
+  deleteMessages?: boolean;
 
   // ============================================================
   // Batch Operations Configuration
@@ -345,7 +222,9 @@ export interface PollerConfig {
   deleteBatchSize?: number;
 
   /**
-   * Milliseconds to wait before triggering batch delete.
+   * Maximum time in milliseconds to wait after the first message is queued
+   * for deletion before triggering the batch delete operation.
+   * This allows multiple messages to be batched together for efficiency.
    * @defaultValue 2000
    */
   deleteWaitMs?: number;
@@ -356,14 +235,16 @@ export interface PollerConfig {
 
   /**
    * Format of message body.
-   * - 'plain': Return body as string
+   * - 'text': Return body as string (default)
    * - 'json': Parse body as JSON
-   * @defaultValue 'plain'
+   * @defaultValue 'text'
    */
-  bodyFormat?: "plain" | "json";
+  bodyFormat?: "text" | "json";
 
   /**
    * Unwrap SNS message envelope.
+   * When true, if the message body is an SNS notification envelope,
+   * the poller will extract the inner message content.
    * @defaultValue false
    */
   unwrapSns?: boolean;
@@ -374,15 +255,18 @@ export interface PollerConfig {
 
   /**
    * System attributes to retrieve with messages.
+   * Uses MessageSystemAttributeName from @aws-sdk/client-sqs.
+   * These are attributes like ApproximateReceiveCount, SentTimestamp, etc.
    * @defaultValue []
-   * @example ['All', 'ApproximateReceiveCount', 'SentTimestamp']
+   * @example ['All'] or ['ApproximateReceiveCount', 'SentTimestamp']
    */
-  attributeNames?: AttributeName[];
+  systemAttributeNames?: MessageSystemAttributeName[];
 
   /**
    * Custom message attributes to retrieve.
+   * These are user-defined attributes set when sending messages.
    * @defaultValue []
-   * @example ['All', 'CustomAttribute.*']
+   * @example ['All'] or ['CustomAttribute.*', 'SpecificAttribute']
    */
   messageAttributeNames?: string[];
 
@@ -392,8 +276,8 @@ export interface PollerConfig {
 
   /**
    * Callback invoked before each polling request.
-   * Can be used to implement custom logic or stop polling.
-   * @example (stats) => { if (stats.errorCount > 10) throw ':stop_polling'; }
+   * Can be used to implement custom logic or stop polling by throwing an error.
+   * @example (stats) => { if (stats.errorCount > 10) throw new Error('Too many errors'); }
    */
   beforeRequest?: (stats: PollerStats) => void;
 
@@ -402,51 +286,44 @@ export interface PollerConfig {
   // ============================================================
 
   /**
-   * Automatically register signal handlers for graceful shutdown.
+   * Register signal handlers for graceful shutdown.
    * When enabled, the poller listens for SIGTERM and SIGINT signals
    * and performs a soft stop (waiting for in-flight messages to complete).
    * @defaultValue true
    */
-  autoGracefulShutdown?: boolean;
+  gracefulShutdown?: boolean;
 
   /**
    * Maximum time in milliseconds to wait for in-flight messages during shutdown.
    * After this timeout, the poller will force stop.
-   * Only used if autoGracefulShutdown is true.
+   * Only used if gracefulShutdown is true.
    * @defaultValue 30000 (30 seconds)
    */
   shutdownTimeoutMs?: number;
 
   // ============================================================
-  // Queue Creation Configuration
+  // Event Callbacks Configuration
   // ============================================================
 
   /**
-   * Delay in seconds for new messages (0-900).
-   * Only used when creating a queue.
-   * @defaultValue 0
+   * Event callbacks to register during construction.
+   * These can also be assigned directly to poller.on after construction.
+   *
+   * @example
+   * ```typescript
+   * const poller = new QueuePoller({
+   *   queueUrl: '...',
+   *   on: {
+   *     message: (message, stats) => {
+   *       console.log('Received:', message.body);
+   *       message.enqueueDelete();
+   *     },
+   *     error: (error) => console.error(error),
+   *   },
+   * });
+   * ```
    */
-  delaySecs?: number;
-
-  /**
-   * Maximum message size in bytes (1024-262144).
-   * Only used when creating a queue.
-   * @defaultValue 262144 (256 KB)
-   */
-  maxMessageBytes?: number;
-
-  /**
-   * Message retention period in seconds (60-1209600).
-   * Only used when creating a queue.
-   * @defaultValue 345600 (4 days)
-   */
-  messageRetentionSecs?: number;
-
-  /**
-   * Queue access policy.
-   * Only used when creating a queue.
-   */
-  queuePolicy?: string;
+  on?: Partial<QueuePollerEvents>;
 }
 
 // ============================================================
@@ -464,9 +341,10 @@ export class PollerError extends Error {
   public readonly cause?: Error;
 
   /**
-   * Error name.
+   * Error name - identifies the specific error type.
+   * Subclasses override this with their specific error name.
    */
-  public readonly name = "PollerError" as const;
+  public override readonly name: string = "PollerError";
 
   /**
    * Creates a new PollerError.
@@ -493,7 +371,7 @@ export class TimeoutError extends PollerError {
   /**
    * Error name.
    */
-  public readonly name = "TimeoutError" as const;
+  public override readonly name: string = "TimeoutError";
 
   /**
    * Creates a new TimeoutError.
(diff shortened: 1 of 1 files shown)


lib/lib-sqs/index.ts
```diff
@@ -0,0 +1,1022 @@
...
+ */
+export type AttributeName =
+  | "All"
+  | "ApproximateFirstReceiveTimestamp"
+  | "ApproximateReceiveCount"
+  | "SenderId"
+  | "SentTimestamp"
+  | "MessageDeduplicationId"
+  | "MessageGroupId"
+  | "SequenceNumber";
```
kuhe: This should be replaced by 

`import type { MessageSystemAttributeName } from "@aws-sdk/client-sqs";`

Currently the service type has 10 values, while this has 8. Is that intentional?

lib/lib-sqs/index.ts
```diff
@@ -0,0 +1,1022 @@
...
+
+/**
+ * Message attribute value structure.
+ * @public
+ */
+export interface MessageAttributeValue {
```
kuhe: `type MessageAttributeValue` from the client package

lib/lib-sqs/index.ts
```diff
@@ -0,0 +1,1022 @@
...
+  /** The data type of the attribute value. */
+  DataType: string;
+  /** String value of the attribute. */
+  StringValue?: string;
+  /** Binary value of the attribute. */
+  BinaryValue?: Buffer;
```
kuhe: All byte array types must be declared as `Uint8Array` for compatibility, even if this one is eventually replaced by the import from the client types.

lib/lib-sqs/index.ts
```diff
@@ -0,0 +1,1022 @@
...
+ * Map of message attribute names to values.
+ * @public
+ */
+export type MessageAttributes = Record<string, MessageAttributeValue>;
+
+/**
+ * Message system attributes returned by SQS.
+ * @public
+ */
+export interface MessageSystemAttributes {
```
kuhe: this doesn't align with what the service calls a "message system attribute", which only includes AWSTraceHeader. Can you explain the difference between that and this in the tsdoc description of this type?

lib/lib-sqs/index.ts
```diff
@@ -0,0 +1,1022 @@
...
+  MessageGroupId?: string;
+  /** Sequence number for FIFO queues. */
+  SequenceNumber?: string;
+}
+
+/**
+ * Raw SQS message structure.
+ * @public
+ */
+export interface SQSMessage {
```
kuhe: why is this prefixed with SQS when other things aren't? 

Also, this should be replaced by `type Message` from the client.

lib/lib-sqs/index.ts
```diff
@@ -0,0 +1,1022 @@
...
+  MessageAttributes?: MessageAttributes;
+  /** MD5 digest of the message attributes. */
+  MD5OfMessageAttributes?: string;
+}
+
+/**
+ * Result of sending a single message.
+ * @public
+ */
+export interface SendMessageResult {
```
kuhe: replace with client defined type

lib/lib-sqs/index.ts
```diff
@@ -0,0 +1,1022 @@
...
+  MD5OfMessageAttributes?: string;
+  /** Sequence number for FIFO queues. */
+  SequenceNumber?: string;
+}
+
+/**
+ * Entry in a successful batch send result.
+ * @public
+ */
+export interface SendMessageBatchResultEntry {
```
kuhe: replace with client type of same name

lib/lib-sqs/index.ts
```diff
@@ -0,0 +1,1022 @@
...
+  MD5OfMessageAttributes?: string;
+  /** Sequence number for FIFO queues. */
+  SequenceNumber?: string;
+}
+
+/**
+ * Entry in a failed batch operation result.
+ * @public
+ */
+export interface BatchResultErrorEntry {
```
kuhe: replace with client type

lib/lib-sqs/index.ts
```diff
@@ -0,0 +1,1022 @@
...
+  Code: string;
+  /** Error message. */
+  Message: string;
+}
+
+/**
+ * Result of sending multiple messages.
+ * @public
+ */
+export interface SendMessageBatchResult {
```
kuhe: replace with client type

lib/lib-sqs/index.ts
```diff
@@ -0,0 +1,1022 @@
...
+  Successful: SendMessageBatchResultEntry[];
+  /** Failed message sends. */
+  Failed: BatchResultErrorEntry[];
+}
+
+/**
+ * JSON representation of poller statistics.
+ * @public
+ */
+export interface PollerStatsJSON {
```
kuhe: this should not be called JSON because it is an object type and not a string

you could call it SerializablePollerStats

lib/lib-sqs/index.ts
```diff
@@ -0,0 +1,1022 @@
...
+ */
+export interface SendMessageBatchResult {
+  /** Successfully sent messages. */
+  Successful: SendMessageBatchResultEntry[];
+  /** Failed message sends. */
+  Failed: BatchResultErrorEntry[];
+}
+
+/**
+ * JSON representation of poller statistics.
```
kuhe:
```suggestion
 * JSON-serializable representation of poller statistics.
```

lib/lib-sqs/index.ts
```diff
@@ -0,0 +1,1022 @@
...
+  averageMessagesPerRequest: number;
+  /** Current polling duration in milliseconds. */
+  pollingDurationMs: number;
+}
+
+/**
+ * Information about a failed message deletion.
+ * @public
+ */
+export interface DeleteFailure {
```
kuhe: we have e.g. `BatchResultErrorEntry` from the service in which all members are title cased. This has the same fields but in camelCase. Is that necessary?

lib/lib-sqs/index.ts
```diff
@@ -0,0 +1,1022 @@
...
+ *
+ * This file contains the public API signatures for the SQS Poller library.
+ * For feedback and discussion, please visit: https://github.com/aws/aws-sdk-js-v3/issues/6015
+ *
+ * @packageDocumentation
+ */
+
+import { EventEmitter } from "events";
+import type { SQSClient } from "@aws-sdk/client-sqs";
+import type { Credentials } from "@aws-sdk/types";
```
kuhe:
```suggestion
import type { AwsCredentialIdentity } from "@aws-sdk/types";
```

`Credentials` is deprecated

lib/lib-sqs/index.ts
```diff
@@ -0,0 +1,1022 @@
...
+  code: string;
+  /** Error message. */
+  message: string;
+}
+
+/**
+ * AWS SDK configuration options.
+ * @public
+ */
+export interface AWSConfig {
```
kuhe: There should not be an attempt to redefine the AWS Config, and these types don't match the SQSClient constructor parameter type.

To access the constructor arg type

```ts
import type { SQSClientConfig } from "@aws-sdk/client-sqs";
```

lib/lib-sqs/index.ts
```diff
@@ -0,0 +1,1022 @@
...
+   * @example 'https://sqs.us-east-1.amazonaws.com/123456789/my-queue'
+   */
+  queueUrl: string;
+
+  // ============================================================
+  // AWS Configuration
+  // ============================================================
+
+  /** AWS SDK configuration options. */
+  awsConfig?: AWSConfig;
```
kuhe:
```suggestion
  clientConfig?: SQSClientConfig;
```

lib/lib-sqs/index.ts
```diff
@@ -0,0 +1,1022 @@
...
+
+  /**
+   * Long polling wait time in seconds (0-20).
+   * @defaultValue 20
+   */
+  receiveWaitTimeSecs?: number;
+
+  /**
+   * Milliseconds to wait between polls when queue is active.
+   * @defaultValue 0
```
kuhe, line 148: what does poll interval 0 mean?

kuhe: what if the user sets this and receiveWaitTime to 0?

lib/lib-sqs/index.ts
```diff
@@ -0,0 +1,1022 @@
...
+  idlePollIntervalMs?: number;
+
+  /**
+   * Milliseconds to wait before retrying after an error.
+   * @defaultValue 2000
+   */
+  pollRetryMs?: number;
+
+  /**
+   * Stop polling after this many seconds of no messages.
```
kuhe: should be milliseconds like the rest of the API

lib/lib-sqs/index.ts
```diff
@@ -0,0 +1,1022 @@
...
+   * Number of messages to receive per request (1-10).
+   * @defaultValue 10
+   */
+  receiveBatchSize?: number;
+
+  /**
+   * Long polling wait time in seconds (0-20).
+   * @defaultValue 20
+   */
+  receiveWaitTimeSecs?: number;
```
kuhe: this API should accept milliseconds for all timing related things, since it is the basic time unit of the existing Date system.

lib/lib-sqs/index.ts
```diff
@@ -0,0 +1,1022 @@
...
+
+  // ============================================================
+  // Message Handling Configuration
+  // ============================================================
+
+  /**
+   * Visibility timeout in seconds for received messages.
+   * If not specified, uses the queue's default setting.
+   */
+  visibilityTimeoutSecs?: number;
```
kuhe: Should be milliseconds.

Granted that SQS has API fields in seconds, but this sits adjacent to configuration that is in ms.

lib/lib-sqs/index.ts
```diff
@@ -0,0 +1,1022 @@
...
+   * @defaultValue false
+   */
+  autoExtendTimeout?: boolean;
+
+  /**
+   * Milliseconds before timeout to make extension call.
+   * Only used if autoExtendTimeout is true.
+   * @defaultValue 5000
+   */
+  advancedCallMs?: number;
```
kuhe: This should be called `autoExtendTimeoutAdvancedCallMs` or merged into the boolean

lib/lib-sqs/index.ts
```diff
@@ -0,0 +1,1022 @@
...
+
+  /**
+   * Stop extending timeout after this many seconds.
+   * Only used if autoExtendTimeout is true.
+   * @defaultValue 43200 (12 hours)
+   */
+  noExtensionsAfterSecs?: number;
```
kuhe:
```suggestion
  maxExtendedTimeoutMs?: number;
```
or `maxTimeoutExtensionMs`

`noExtensionsAfter` can grammatically imply a timestamp rather than a duration is accepted

lib/lib-sqs/index.ts
```diff
@@ -0,0 +1,1022 @@
...
+
+  /**
+   * Skip automatic message deletion.
+   * If true, messages must be manually deleted.
+   * @defaultValue false
+   */
+  skipDelete?: boolean;
```
kuhe:
```ts
/**
 * @defaultValue true
 */
delete?: boolean;
```

or `deleteMessages?: true | boolean`

lib/lib-sqs/index.ts
```diff
@@ -0,0 +1,1022 @@
...
+  // ============================================================
+
+  /**
+   * Number of messages to batch for deletion (1-10).
+   * @defaultValue 10
+   */
+  deleteBatchSize?: number;
+
+  /**
+   * Milliseconds to wait before triggering batch delete.
```
kuhe: to wait after what?

lib/lib-sqs/index.ts
```diff
@@ -0,0 +1,1022 @@
...
+  // Message Format Configuration
+  // ============================================================
+
+  /**
+   * Format of message body.
+   * - 'plain': Return body as string
+   * - 'json': Parse body as JSON
+   * @defaultValue 'plain'
+   */
+  bodyFormat?: "plain" | "json";
```
kuhe: `plain` should be called `text`, like in content type headers

lib/lib-sqs/index.ts
```diff
@@ -0,0 +1,1022 @@
...
+  // ============================================================
+  // Message Attributes Configuration
+  // ============================================================
+
+  /**
+   * System attributes to retrieve with messages.
+   * @defaultValue []
+   * @example ['All', 'ApproximateReceiveCount', 'SentTimestamp']
+   */
+  attributeNames?: AttributeName[];
```
kuhe: should be prefixed with the specific type of attribute, since the language of the service contains MessageAttributeName, SystemAttributename, QueueAttributeName

lib/lib-sqs/index.ts
```diff
@@ -0,0 +1,1022 @@
...
+  // Lifecycle Configuration
+  // ============================================================
+
+  /**
+   * Automatically register signal handlers for graceful shutdown.
+   * When enabled, the poller listens for SIGTERM and SIGINT signals
+   * and performs a soft stop (waiting for in-flight messages to complete).
+   * @defaultValue true
+   */
+  autoGracefulShutdown?: boolean;
```
kuhe: This, and other parts of the API language, don't need the `auto` prefix because you are instructing it to do these things or not do them.

Everything the poller does on the user's behalf is "automatic", so there is no meaningful distinction between saying e.g. `autoExtend: true` / `extend: true`.

lib/lib-sqs/index.ts
```diff
@@ -0,0 +1,1022 @@
...
+   * @defaultValue 262144 (256 KB)
+   */
+  maxMessageBytes?: number;
+
+  /**
+   * Message retention period in seconds (60-1209600).
+   * Only used when creating a queue.
+   * @defaultValue 345600 (4 days)
+   */
+  messageRetentionSecs?: number;
```
kuhe:
```suggestion
  messageRetentionPeriodMs?: number;
```

use `ms` for this and delay since this does not map directly to the createQueue signature.

The SQS API has the addition of the word period to indicate the number is a duration.

lib/lib-sqs/index.ts
```diff
@@ -0,0 +1,1022 @@
...
+  // ============================================================
+  // Queue Creation Configuration
+  // ============================================================
+
+  /**
+   * Delay in seconds for new messages (0-900).
+   * Only used when creating a queue.
+   * @defaultValue 0
+   */
+  delaySecs?: number;
```
kuhe:
```suggestion
  delayMs?: number;
```

(Acknowledging that the SQS API has this in seconds, this signature is very different from the SQS API for createQueue and therefore should consistently use ms across itself).


This createQueue API is an alternate signature and subset of the SDK's CreateQueue operation. What is the use case for this shortened syntax, and having only this specific subset of attributes?

lib/lib-sqs/index.ts
```diff
@@ -0,0 +1,1022 @@
...
+  /**
+   * Maximum time in milliseconds to wait for in-flight messages during shutdown.
+   * After this timeout, the poller will force stop.
+   * Only used if autoGracefulShutdown is true.
+   * @defaultValue 30000 (30 seconds)
+   */
+  shutdownTimeoutMs?: number;
+
+  // ============================================================
+  // Queue Creation Configuration
```
kuhe: why is this in the poller config at all? automatic creation of nonexistent queues?

lib/lib-sqs/index.ts
```diff
@@ -0,0 +1,1022 @@
...
+  /**
+   * Resets all statistics to initial values.
+   * @public
+   */
+  public reset(): void {
+    throw new Error("not implemented");
+  }
+
+  /**
+   * Returns a plain object representation of the statistics.
```
kuhe: description can be omitted since it is the same as the returns text

lib/lib-sqs/index.ts
```diff
@@ -0,0 +1,1022 @@
...
+ * @public
+ */
+export class Message {
+  /**
+   * Parsed message body.
+   * - If bodyFormat is 'json', this will be the parsed object.
+   * - If bodyFormat is 'plain', this will be the raw string.
+   * - If unwrapSns is true, this will be the SNS message content.
+   */
+  public readonly body: unknown;
```
kuhe: this should have a generic type if a type parameter is supplied to the Poller, and its messages.

lib/lib-sqs/index.ts
```diff
@@ -0,0 +1,1022 @@
...
+   */
+  constructor(raw: SQSMessage) {
+    throw new Error("not implemented");
+  }
+
+  /**
+   * Queues the message for deletion.
+   * @public
+   */
+  public del(): void {
```
kuhe: `enqueueDelete`, otherwise `delete` implies it is fully deleted after the method returns

lib/lib-sqs/index.ts
```diff
@@ -0,0 +1,1022 @@
...
+  /** Emitted when a message is marked as handled. */
+  handled: (message: Message) => void;
+  /** Emitted when a message is successfully deleted. */
+  deleted: (message: Message) => void;
+  /** Emitted when a message is queued for deletion. */
+  delQueued: (message: Message) => void;
+  /** Emitted when a message is released back to the queue. */
+  released: (message: Message) => void;
+  /** Emitted when messages are received from SQS. */
+  gotMessages: (count: number) => void;
```
kuhe:
```suggestion
  receivedMessages: (count: number) => void;
```

lib/lib-sqs/index.ts
```diff
@@ -0,0 +1,1022 @@
...
+ */
+export interface QueuePollerEvents {
+  /** Emitted when a message is received from the queue. */
+  message: (message: Message, stats: PollerStats) => void;
+  /** Emitted when a message is marked as handled. */
+  handled: (message: Message) => void;
+  /** Emitted when a message is successfully deleted. */
+  deleted: (message: Message) => void;
+  /** Emitted when a message is queued for deletion. */
+  delQueued: (message: Message) => void;
```
kuhe:
```suggestion
  deletionEnqueued: (message: Message) => void;
```

lib/lib-sqs/index.ts
```diff
@@ -0,0 +1,1022 @@
...
+  /** Emitted when all in-flight messages have been handled. */
+  drained: () => void;
+  /** Emitted when polling is aborted (hard stop). */
+  aborted: () => void;
+  /** Emitted when graceful shutdown is initiated. */
+  shutdown: (signal: string) => void;
+  /** Emitted when a message's visibility timeout is extended. */
+  timeoutExtended: (message: Message) => void;
+  /** Emitted when automatic timeout extension fails. */
+  autoExtendFail: (data: { message: Message; error: Error }) => void;
```
kuhe:
```suggestion
  extensionError: (data: { message: Message; error: Error }) => void;
```

lib/lib-sqs/index.ts
```diff
@@ -0,0 +1,1022 @@
...
+  /** Emitted when a polling error occurs. */
+  error: (error: Error) => void;
+  /** Emitted when message deletion fails. */
+  delError: (error: DeleteError) => void;
```
kuhe:
```suggestion
  deleteError: (error: DeleteError) => void;
```

lib/lib-sqs/index.ts
```diff
@@ -0,0 +1,1022 @@
...
+  /** Emitted when polling is aborted (hard stop). */
+  aborted: () => void;
+  /** Emitted when graceful shutdown is initiated. */
+  shutdown: (signal: string) => void;
+  /** Emitted when a message's visibility timeout is extended. */
+  timeoutExtended: (message: Message) => void;
+  /** Emitted when automatic timeout extension fails. */
+  autoExtendFail: (data: { message: Message; error: Error }) => void;
+  /** Emitted when a polling error occurs. */
+  error: (error: Error) => void;
```
kuhe, line 677: Why is poll error simply "error", and processing error has a more specific event name?

I would expect the `error` event to include _all_ errors. Polling error can be a separate more-specific error event type.

lib/lib-sqs/index.ts
```diff
@@ -0,0 +1,1022 @@
...
+  /** Emitted when message deletion fails. */
+  delError: (error: DeleteError) => void;
+  /** Emitted when message processing throws an error. */
+  processing_error: (error: Error, message: Message) => void;
+  /** Emitted when message processing times out. */
+  timeout_error: (error: TimeoutError, message: Message) => void;
```
kuhe: these last two events should be in camelCase

lib/lib-sqs/index.ts
```diff
@@ -0,0 +1,1022 @@
...
+   * @throws {PollerError} If poller is already running.
+   */
+  public start(): void {
+    throw new Error("not implemented");
+  }
+
+  /**
+   * Stops polling for messages.
+   * @public
+   * @param soft - If true, waits for in-flight messages to complete.
```
kuhe: default value?

lib/lib-sqs/index.ts
```diff
@@ -0,0 +1,1022 @@
...
+  // ============================================================
+  // Message Operations
+  // ============================================================
+
+  /**
+   * Deletes a message from the queue.
+   * @public
+   * @param message - The message to delete.
+   */
+  public deleteMessage(message: Message): void {
```
kuhe: what is the difference between this and `Message.prototype.del()`?

If messages are primarily expected to have delete called on themselves then this method should not exist.

lib/lib-sqs/index.ts
```diff
@@ -0,0 +1,1022 @@
...
+  public deleteQueue(): Promise<void> {
+    throw new Error("not implemented");
+  }
+
+  /**
+   * Gets the URL of the configured queue.
+   * @public
+   * @returns Promise that resolves with the queue URL.
+   */
+  public getQueueUrl(): Promise<string> {
```
kuhe: when does this ever differ from `public readonly queueUrl`?

lib/lib-sqs/index.ts
```diff
@@ -0,0 +1,1022 @@
...
+
+  /**
+   * Releases a message back to the queue immediately.
+   * @public
+   * @param message - The message to release.
+   * @returns Promise that resolves when release is complete.
+   */
+  public releaseMessage(message: Message): Promise<void> {
+    throw new Error("not implemented");
+  }
```
kuhe: many of these poller message operation methods appear redundant with message methods

lib/lib-sqs/index.ts
```diff
@@ -0,0 +1,1022 @@
...
+   * Gets the URL of the configured queue.
+   * @public
+   * @returns Promise that resolves with the queue URL.
+   */
+  public getQueueUrl(): Promise<string> {
+    throw new Error("not implemented");
+  }
+
+  // ============================================================
+  // Message Sending
```
kuhe: This (and creation) are outside the implied scope of a message poller and should be removed.

What does the poller simplify about creation or message publishing that differs from the SDK client?

lib/lib-sqs/index.ts
```diff
@@ -0,0 +1,1022 @@
...
+  processing_error: (error: Error, message: Message) => void;
+  /** Emitted when message processing times out. */
+  timeout_error: (error: TimeoutError, message: Message) => void;
+}
+
+/**
+ * The main class for polling messages from an SQS queue.
+ * @public
+ */
+export class QueuePoller extends EventEmitter {
```
kuhe: To reduce API surface, Poller should not extend EventEmitter, but instead accept a callback function for each event type.

```ts
type PollerConfig = {
  // ...
  on: Partial<QueuePollerEvents>
  // ...
}

class QueuePoller {
  public readonly on: QueuePollerEvents; // has default functions for each event type
}
```

users may opt in for EventEmitter by adapting it

```ts
const poller = new QueuePoller();
const emitter = new EventEmitter();

for (const event in poller.on) {
  poller.on[event] = (...args) => emitter.emit(event, ...args);
}

emitter.on("message", (message) => {
  // do something
});
```

lib/lib-sqs/index.ts
```diff
@@ -0,0 +1,1022 @@
...
+  public readonly topicName?: string;
+
+  /**
+   * Timestamp when message was received by poller.
+   */
+  public readonly receivedAt: Date;
+
+  /**
+   * Creates a new Message instance.
+   * @internal
```
siddsriv: the class is marked `@public` but its constructor is `@internal`, perhaps leave a comment explaining why, if it is intentional.

lib/lib-sqs/index.ts
```diff
@@ -0,0 +1,1022 @@
...
+
+  /**
+   * Maximum number of messages to process simultaneously.
+   * Set to 0 for unlimited.
+   * @defaultValue 100
+   */
+  maxInFlight?: number;
+
+  /**
+   * Number of messages to receive per request (1-10).
```
siddsriv, line 128: nit: what is expected behavior if the value is not between 1-10? maybe an inline comment here would be useful for validation documentation. Depending on where else that documentation might exist, this might be a necessity

lib/lib-sqs/index.ts
jameslevine: Done! Now importing MessageSystemAttributeName directly from @aws-sdk/client-sqs to stay in sync with the service model and automatically pick up any new attributes.

lib/lib-sqs/index.ts
jameslevine: Updated to re-export from @aws-sdk/client-sqs for consistency.

lib/lib-sqs/index.ts
jameslevine: Addressed by using the imported types from the client package, which correctly use Uint8Array.

lib/lib-sqs/index.ts
jameslevine: Good point about the naming confusion. I've updated the Message class to use attributes: Partial<Record<MessageSystemAttributeName, string>> which represents the system attributes returned by the ReceiveMessage API (like ApproximateReceiveCount, SentTimestamp, etc.).

I've added clarifying TSDoc:

Note: This contains attributes like ApproximateReceiveCount, SentTimestamp, etc.
This is different from MessageSystemAttributeValue (used for AWSTraceHeader when sending).

lib/lib-sqs/index.ts
jameslevine: Removed 👍

lib/lib-sqs/index.ts
jameslevine: Removed 👍

lib/lib-sqs/index.ts
jameslevine: replaced 👍

lib/lib-sqs/index.ts
jameslevine: replaced 👍

lib/lib-sqs/index.ts
jameslevine: Good point! Renamed to SerializablePollerStats with updated TSDoc: "JSON-serializable representation of poller statistics."

lib/lib-sqs/index.ts
jameslevine: Updated the TSDoc to "JSON-serializable representation of poller statistics. All Date fields are converted to ISO strings for serialization."

lib/lib-sqs/index.ts
jameslevine: Good observation. I've removed the custom DeleteFailure interface and now use BatchResultErrorEntry directly from @aws-sdk/client-sqs for consistency with SDK patterns.

lib/lib-sqs/index.ts
jameslevine: Addressed by removing the custom AWSConfig interface entirely and using SQSClientConfig from the client package, which handles credentials correctly.

lib/lib-sqs/index.ts
jameslevine: Removed the custom AWSConfig interface. Users can now pass clientConfig?: SQSClientConfig directly, which provides full access to all SQS client configuration options.

lib/lib-sqs/index.ts
jameslevine: Done 👍

lib/lib-sqs/index.ts
jameslevine: `0` means no additional delay between polls - the next poll starts immediately after the previous one completes.

If both `activePollIntervalMs` and `receiveWaitTimeMs` are set to 0, it would result in rapid short-polling. Should we add validation to prevent this, or is it acceptable to let users make that choice? Open to your recommendation.

lib/lib-sqs/index.ts
jameslevine: Renamed to SQSRawMessage when re-exporting to clearly distinguish the raw client type from our Message wrapper class. Open to suggestions if you feel this still isn't suitable.

lib/lib-sqs/index.ts
jameslevine: Updated to Ms for consistency with the rest of the API.

lib/lib-sqs/index.ts
jameslevine: Updated to Ms for consistency with the rest of the API.

lib/lib-sqs/index.ts
jameslevine: Updated to Ms for consistency with the rest of the API.

lib/lib-sqs/index.ts
jameslevine: Renamed to extendTimeoutAdvanceMs to clarify its relationship to the visibility timeout extension feature. Also removed the 'auto' prefix per your other feedback.

lib/lib-sqs/index.ts
jameslevine: Renamed to maxTimeoutExtensionMs which clearly indicates it's a duration limit, with default 43200000 (12 hours).

lib/lib-sqs/index.ts
jameslevine: Good suggestion for using positive boolean naming. Changed to deleteMessages?: boolean with default value true.

lib/lib-sqs/index.ts
jameslevine: Clarified the TSDoc:

```
/**
 * Maximum time in milliseconds to wait after the first message is queued
 * for deletion before triggering the batch delete operation.
 * This allows multiple messages to be batched together for efficiency.
 * @defaultValue 2000
 */
deleteWaitMs?: number;
```

lib/lib-sqs/index.ts
jameslevine: Updated to 'text' | 'json' to align with content-type header conventions.

lib/lib-sqs/index.ts
jameslevine: Renamed to systemAttributeNames to clearly indicate these are system attributes (as opposed to custom message attributes). Added clarifying TSDoc explaining these are attributes like ApproximateReceiveCount, SentTimestamp, etc.

lib/lib-sqs/index.ts
jameslevine: Removed the 'auto' prefix:

autoGracefulShutdown → gracefulShutdown
autoExtendTimeout → extendTimeout

lib/lib-sqs/index.ts
jameslevine: Done 👍

lib/lib-sqs/index.ts
jameslevine: Added generic type parameter:

```
export class Message<T = unknown> {
  public readonly body: T;
  // ...
}

export class QueuePoller<T = unknown> {
  public readonly on: QueuePollerEvents<T>;
  // ...
}
```

lib/lib-sqs/index.ts
jameslevine: Renamed to enqueueDelete() to clearly indicate the message is queued for batch deletion rather than immediately deleted. Updated TSDoc to explain the batching behavior.

lib/lib-sqs/index.ts
jameslevine: Renamed 👍

lib/lib-sqs/index.ts
jameslevine: Renamed to `deletionEnqueued`.

lib/lib-sqs/index.ts
jameslevine: Renamed to `extensionError`.

lib/lib-sqs/index.ts
jameslevine: Renamed to `deleteError`.

lib/lib-sqs/index.ts
jameslevine: I've restructured the error events:

- The `error` event now emits for ALL errors (polling, processing, deletion, extension)
- Specific error events (`deleteError, extensionError, processingError, timeoutError`) are also emitted for users who want to handle specific error types
- Added TSDoc explaining that error types can be distinguished by the error's `name` property

lib/lib-sqs/index.ts
jameslevine: Updated to camelCase: processingError, timeoutError.

lib/lib-sqs/index.ts
jameslevine: Added `@defaultValue false` and updated the signature to `stop(soft: boolean = false)`. By default, `stop()` performs a hard stop. Pass `true` for a soft stop that waits for in-flight messages.

lib/lib-sqs/index.ts
jameslevine: Removed redundant methods:

- `deleteMessage()` - users should call message.enqueueDelete() directly
- `getQueueUrl()` - the queueUrl property is sufficient
- `releaseMessage()` - users should call message.release() directly

Kept `deleteMessages(messages[])` as a batch operation for cases where users want to delete multiple messages in a single API call.

lib/lib-sqs/index.ts
jameslevine: Agree on this. Removed message sending and queue creation functionality. The poller should now focus solely on polling and message consumption.

lib/lib-sqs/index.ts
jameslevine: Implemented 👍. The poller now uses a callback-based on property instead of extending `EventEmitter`:

```
class QueuePoller {
  public readonly on: QueuePollerEvents;
}
```
Added an example in TSDoc showing how users can adapt to `EventEmitter` if preferred:

```
const poller = new QueuePoller({ queueUrl: "..." });
const emitter = new EventEmitter();

poller.on.message = (message, stats) => emitter.emit("message", message, stats);
poller.on.error = (error) => emitter.emit("error", error);
```

lib/lib-sqs/index.ts
jameslevine: Added clarifying comment in the class TSDoc:

```
/**
 * Wrapper class for SQS messages with convenience methods.
 *
 * Note: The constructor is marked @internal because Message instances
 * should only be created by the QueuePoller when messages are received.
 * Users should not instantiate Message objects directly.
 */
```

lib/lib-sqs/index.ts
jameslevine: Added `@throws {PollerError}` documentation:

```
/**
 * Number of messages to receive per request (1-10).
 * Values outside this range will throw a PollerError during construction.
 * This limit is enforced by the SQS ReceiveMessage API.
 * @defaultValue 10
 * @throws {PollerError} If value is not between 1 and 10 (inclusive).
 */
receiveBatchSize?: number;
```

lib/lib-sqs/index.ts
jameslevine: Removed - the queueUrl property is sufficient. 👍

lib/lib-sqs/index.ts
jameslevine: Agreed. Removed redundant methods - users should call message.release() directly.

lib/lib-sqs/index.ts
jameslevine: Good point. Removed queue creation configuration entirely. Users should create queues using the SQS client directly.

lib/lib-sqs/index.ts
jameslevine: Removed queue creation configuration entirely.

lib/lib-sqs/index.ts
jameslevine: Removed queue creation configuration entirely.

lib/lib-sqs/index.ts
kuhe: Yes, to prevent e.g. service errors from causing client side requests to scale up limitlessly, the poller implementation should enforce a minimum time between outgoing retrieval requests.

lib/lib-sqs/index.ts
kuhe: would you expect soft stops to be more common? perhaps it should be the default

PR #7689, opened 2026-01-28
chore: test new model generation in ci

smilkuri:
### Issue
Internal JS-6480

### Description
Adds `new-client-test-codegen` test model with Weather service and test the generation of client in `pr-build` workflow.

### Testing
CI

### Checklist

---

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,6 +22,13 @@ test-unit: bundles
 	yarn g:vitest run -c vitest.config.clients.unit.mts
 	npx jest -c jest.config.js
 
+test-client-codegen:
+	rm -rf codegen/new-client-test-codegen/build
+	rm -rf clients/client-test-weather
+	cd codegen && ./gradlew :new-client-test-codegen:build
+	mkdir -p clients/client-test-weather
+	cp -r codegen/new-client-test-codegen/build/smithyprojections/new-client-test-codegen/source/typescript-client-codegen/* clients/client-test-weather/
+	
 # typecheck for test code.
 test-types:
 	npx tsc -p tsconfig.test.json
--- a/codegen/new-client-test-codegen/smithy-build.json
+++ b/codegen/new-client-test-codegen/smithy-build.json
@@ -5,7 +5,8 @@
     "typescript-client-codegen": {
       "service": "example.weather#Weather",
       "package": "@aws-sdk/client-test-weather",
-      "packageVersion": "0.0.1"
+      "packageVersion": "0.0.1",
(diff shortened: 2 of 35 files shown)


codegen/new-client-test-codegen/build.gradle.kts
```diff
@@ -0,0 +1,58 @@
...
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
```
kuhe, line 4: use the modern license header

.github/workflows/codegen-ci.yml
```diff
@@ -47,3 +52,33 @@ jobs:
...
+          cd codegen/new-client-test-codegen/build/smithyprojections/new-client-test-codegen/source/typescript-client-codegen
+          touch yarn.lock
+          yarn install --no-immutable
+          yarn build
+
+      - name: Run validation checks on generated client
+        run: |
+          cd codegen/new-client-test-codegen/build/smithyprojections/new-client-test-codegen/source/typescript-client-codegen
+          node ../../../../../../../scripts/runtime-dependency-version-check/runtime-dep-version-check.js
+          node ../../../../../../../scripts/runtime-dependency-version-check/check-dependencies.js
```
kuhe: why CD in if you're going to execute things outside the folder?

kuhe: are these the only validations run on generated clients?

this should be generated to the clients folder, a local commit made (for commit hooks), and then running integration tests (no e2e)

smilkuri: Just realised the validation scripts use [dirname](https://github.com/aws/aws-sdk-js-v3/blob/main/scripts/runtime-dependency-version-check/check-dependencies.js#L10) to find the repo root, not process.cwd(). So cd isn't required. But, this script only scan `packages`, `packages-internal`, and `private` folders, so they won't validate the generated client in `codegen/.../build/` anyway. Per the other comment, will generate this to clients/ folder and let the commit hooks run all validations instead.

kuhe: generate into the `/private` folder

Issue #7496, opened 2025-11-12
Uncatchable `ERR_STREAM_PREMATURE_CLOSE` in @aws-sdk/eventstream-handler-node crashes Node.js process


monarchwadia:
### Describe the feature

# Problem

When using `@aws-sdk/client-transcribe-streaming` with expired credentials, an `ERR_STREAM_PREMATURE_CLOSE` error is thrown in a callback context that cannot be caught by try/catch, promise handlers, or stream error listeners. This causes an unavoidable Node.js process crash.

## Impact

Impossible to build resilient production applications when stream errors crash the entire server instead of allowing graceful error handling.

# Originating file

`@aws-sdk/eventstream-handler-node@3.922.0/dist-cjs/index.js:89`

```javascript
stream.pipeline(payloadStream, signingStream, request.body, (err) => {
    if (err) {
        throw err;  // ⚠️ Throws in callback, bypasses all error handling
    }
});
```

# Reproduction

1. Create `TranscribeStreamingClient` with expired credentials
2. Start streaming with `StartStreamTranscriptionCommand`
3. Observe uncaught exception crash

# Current Workaround

```typescript
process.on('uncaughtException', (error: Error) => {
  if ((error as any).code === 'ERR_STREAM_PREMATURE_CLOSE') {
    console.error('Stream error:', error)
    return // Prevent crash
  }
  throw error
})
```

### Use Case

Impossible to build resilient production applications when stream errors crash the entire server instead of allowing graceful error handling.

### Proposed Solution

1. Use promises or event emitters instead of killing the raw stream like this.
2. On the thrown error, add `error.source = '@aws-sdk/eventstream-handler-node'` for identification
3. Document that `process.on('uncaughtException')` is required

### Other Information

- `@aws-sdk/client-transcribe-streaming@3.925.0`
- Node.js v23.9.0

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### SDK version used

"@aws-sdk/client-transcribe-streaming": "^3.917.0",

### Environment details (OS name and version, etc.)

Node.js v23.9.0, Mac OS 15.6.1

aBurmeseDev:
Thanks for reaching out. I'm taking a look. Does this only apply to the use case with expired credentials?

vcbui-GAH:
> Thanks for reaching out. I'm taking a look. Does this only apply to the use case with expired credentials?

[...]

PR #7703, opened 2026-02-02
fix(eventstream-handler-node): fix uncatchable error in stream pipeline

siddsriv:
### Issue
#7496 
https://github.com/aws/aws-sdk-js-v3/issues/7706

### Description
fixes issue where `ERR_STREAM_PREMATURE_CLOSE` and other pipeline errors in `@aws-sdk/eventstream-handler-node` would crash the Node.js process when connections closed prematurely (e.g., expired credentials). Changed pipeline callback to reject a Promise instead of throwing, making errors catchable via `Promise.race`. 

### Testing
Unit test case added
```console
 ✓ src/EventSigningStream.spec.ts (1 test) 6ms
 ✓ src/EventStreamPayloadHandler.spec.ts (7 tests) 26ms

 Test Files  2 passed (2)
      Tests  8 passed (8)
```

---

### DIFF
--- a/packages-internal/eventstream-handler-node/src/EventStreamPayloadHandler.ts
+++ b/packages-internal/eventstream-handler-node/src/EventStreamPayloadHandler.ts
@@ -74,14 +74,16 @@ export class EventStreamPayloadHandler implements IEventStreamPayloadHandler {
       systemClockOffsetProvider: this.systemClockOffsetProvider,
     });
 
-    // this promise is used in a race against the http layer's response below
+    // this promise rejects on pipeline error, resolves after http layer completes
+    // this is used in a race against the http layer's response below
+    let resolvePipeline: () => void;
     const pipelineError = new Promise<FinalizeHandlerOutput<any>>((resolve, reject) => {
+      resolvePipeline = () => resolve(undefined as any);
       pipeline(payloadStream, signingStream, request.body, (err: NodeJS.ErrnoException | null) => {
         if (err) {
           reject(new Error(`Pipeline error in @aws-sdk/eventstream-handler-node: ${err.message}`, { cause: err }));
-        } else {
-          resolve(undefined as any); // won't win race, but prevents GC issues
         }
+        // it resolves after next(args) completes
       });
     });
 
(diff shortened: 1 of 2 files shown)


packages-internal/eventstream-handler-node/src/EventStreamPayloadHandler.ts
```diff
@@ -74,15 +74,25 @@ export class EventStreamPayloadHandler implements IEventStreamPayloadHandler {
...
+    // this is used in a race against the http layer's response below
+    const pipelineError = new Promise<never>((_, reject) => {
+      pipeline(payloadStream, signingStream, request.body, (err: NodeJS.ErrnoException | null) => {
+        if (err) {
+          // metadata to identify pipeline errors
+          const pipelineErr = err as any;
+          pipelineErr.$source = "@aws-sdk/eventstream-handler-node";
+          reject(err);
+        }
+        // don't resolve on success - let the http request complete naturally
```
kuhe: this Promise should resolve when next(args) resolves/rejects. 

is an unresolved Promise GC'd? I'd rather not have to remember

packages-internal/eventstream-handler-node/src/EventStreamPayloadHandler.ts
```diff
@@ -74,15 +74,25 @@ export class EventStreamPayloadHandler implements IEventStreamPayloadHandler {
...
-        throw err;
-      }
+    // this promise only rejects on error, never resolves on success
+    // this is used in a race against the http layer's response below
+    const pipelineError = new Promise<never>((_, reject) => {
+      pipeline(payloadStream, signingStream, request.body, (err: NodeJS.ErrnoException | null) => {
+        if (err) {
+          // metadata to identify pipeline errors
+          const pipelineErr = err as any;
+          pipelineErr.$source = "@aws-sdk/eventstream-handler-node";
```
kuhe: don't need a special field for this, we don't do this anywhere else.

if you want additional context prefix the error message or wrap the error in another one

packages-internal/eventstream-handler-node/src/EventStreamPayloadHandler.ts
siddsriv: changed, promise will resolve in latest revision

PR #7708, opened 2026-02-03
chore(eventstream-handler-node): eventstream code maintenance

kuhe:
### Issue
Our event stream handling code needs updates to improve testability. 

### Testing
new unit and integration tests

### Checklist


packages-internal/eventstream-handler-node/src/EventStreamPayloadHandler.ts
```diff
@@ -28,12 +28,6 @@ export interface EventStreamPayloadHandlerOptions {
 
 /**
  * @internal
- *
- * A handler that control the eventstream payload flow:
- * 1. Pause stream for initial attempt.
- * 2. Close the stream is attempt fails.
- * 3. Start piping payload when connection is established.
- * 4. Sign the payload after payload stream starting to flow.
```
kuhe, line 36: removed comments because they are not true anymore

The stream begins flowing whenever the connection is established, regardless of receiving a server ack event. There is no pause or delay in signing.

packages-internal/eventstream-handler-node/src/EventStreamPayloadHandler.ts
```diff
@@ -67,7 +61,7 @@ export class EventStreamPayloadHandler implements IEventStreamPayloadHandler {
     const match = request.headers?.authorization?.match(/Signature=([\w]+)$/);
     // Sign the eventstream based on the signature from initial request.
     const priorSignature = match?.[1] ?? (query?.["X-Amz-Signature"] as string) ?? "";
-    const signingStream = new EventSigningStream({
+    const signingStream = new EventSigningTransformStream({
```
kuhe, line 64: class can be renamed because it was never exported

packages-internal/middleware-websocket/src/WebSocketFetchHandler.ts
```diff
@@ -151,50 +186,76 @@ export class WebSocketFetchHandler {
...
-          value: undefined as any, // unchecked because done=true.
-        });
-      }
+      push({ done: true });
     };
 
     const outputStream: AsyncIterable<Uint8Array> = {
       [Symbol.asyncIterator]: () => ({
-        next: () => {
-          return new Promise((_resolve, _reject) => {
```
kuhe, line 195: this previous implementation could have dropped messages because `onmessage` could call `resolve()` twice, which is meaningless for a single Promise.

PR #7712, opened 2026-02-04
Fix #6426: S3 commands hang after uploading an empty stream

### LINKED ISSUE #6426: S3 commands hang after performing an empty upload using `@aws-sdk/lib-storage`

- [X] I've gone through [Developer Guide](https://docs.aws.amazon.com/sdk-for-javascript/v3/developer-guide) and [API reference](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest)
- [X] I've checked [AWS Forums](https://forums.aws.amazon.com) and [StackOverflow](https://stackoverflow.com/questions/tagged/aws-sdk-js).
- [X] I've searched for [previous similar issues](https://github.com/aws/aws-sdk-js-v3/issues) and didn't find any solution.

### Describe the bug

Immediately ending the stream feeding [the `Upload` utility from `@aws-sdk/lib-storage`](https://github.com/aws/aws-sdk-js-v3/tree/main/lib/lib-storage) sometimes leads to the S3 client getting "locked up": a subsequent command sent through it never finishes.

### SDK version number

@aws-sdk/client-s3@3.637.0, @aws-sdk/lib-storage@3.637.0

Node.js

v20.15.0

### Reproduction Steps

Detailed steps to reproduce are listed in https://github.com/martinslota/aws-sdk-lib-storage-race. Those steps run against Localstack, but the same code also works against actual AWS S3.

### Observed Behavior

After an upload of an empty stream, commands sent through the S3 client never finish.

### Expected Behavior

After an upload of an empty stream, commands sent through the S3 client should continue to work as usual.

### Possible Solution

_No response_

_No response_

Ashutosh0x:
# PR Description: Fix S3 commands hang after uploading an empty stream (Fixes #6426)

## Problem
Uploading an empty stream or file using `@aws-sdk/lib-storage` (which internally uses `PutObject`) can cause the S3 client to hang indefinitely on subsequent requests. 

This is triggered by a known race condition in Node.js (reported as [Node.js issue #60001](https://github.com/nodejs/node/issues/60001)). When `httpRequest.end(emptyBuffer)` is called on a socket that is being reused, a race between the 'free' event and the 'finish' event can leave the socket in a state where it no longer receives data.

## Solution
This PR implements a robust workaround within the `@aws-sdk/lib-storage` package:
- **Detection**: When `__uploadUsingPut` is called with an empty `Uint8Array` (length 0).
- **Environment Check**: The workaround is only applied in **Node.js** environments (using `runtimeConfig.runtime === "node"`).
- **Stream Substitution**: Instead of passing the empty buffer to the S3 client, we pass an empty `Readable` stream (`Readable.from([])`).
- **Explicit Content-Length**: We ensure `Content-Length` is explicitly set to `0` to maintain compatibility.

By using a `Readable` stream, the underlying `NodeHttpHandler` will wait for the stream to end and then call `req.end()` without any arguments. This specifically avoids the buggy code path in Node.js and resolves the hang.

## Changes
- Modified `lib/lib-storage/src/Upload.ts`:
    - Added `Readable` import from `stream`.
    - Integrated `runtimeConfig` to detect the environment.
    - Updated `__uploadUsingPut` to apply the stream-based workaround for empty bodies.

## Testing
- Verified that the logic correctly identifies empty bodies.
- Ensured `Content-Length` is preserved or calculated correctly.
- This workaround has been validated by community members in the original issue thread as effective against the Node.js race condition.
- Verified that the change is transparent to browser/native environments.

## Fixes
Fixes #6426

### DIFF
--- a/lib/lib-storage/src/Upload.ts
+++ b/lib/lib-storage/src/Upload.ts
@@ -26,9 +26,11 @@ import { HttpRequest } from "@smithy/protocol-http";
 import { extendedEncodeURIComponent } from "@smithy/smithy-client";
 import type { AbortController as IAbortController, AbortSignal as IAbortSignal, Endpoint } from "@smithy/types";
 import { EventEmitter } from "events";
+import { Readable } from "stream";
 
 import { byteLength } from "./byteLength";
 import { BYTE_LENGTH_SOURCE, byteLengthSource } from "./byteLengthSource";
+import { runtimeConfig } from "./runtimeConfig";
 import { getChunk } from "./chunker";
 import { BodyDataTypes, Options, Progress } from "./types";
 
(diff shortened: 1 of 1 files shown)


kuhe:
the fix should be applied in https://github.com/smithy-lang/smithy-typescript/blob/main/packages/node-http-handler/src/write-request-body.ts

PR #7731, opened 2026-02-11
feat(lib-storage): S3TransferManager

smilkuri:
### Issue
Internal JS-6610

### Description
This PR adds upload functionality to the S3 Transfer Manager, building upon previous [PR](https://github.com/aws/aws-sdk-js-v3/pull/7238) that wasn't merged. This implementation focuses specifically on single object and multipart upload capabilities [commit](https://github.com/aws/aws-sdk-js-v3/pull/7731/changes/56d88900a6706966f0026ce7cea37442566ff26f).

### Testing
```
 yarn test:e2e
  ✓ src/s3-transfer-manager/S3TransferManager.e2e.spec.ts (26 tests) 274836ms

 Test Files  1 passed | 1 skipped (2)
      Tests  26 passed | 92 skipped (118)
   Start at  16:42:31
   Duration  275.51s (transform 665ms, setup 0ms, import 1.13s, tests 274.84s, environment 0ms)

 RUN  v4.0.18 /local/home/smilkuri/aws-sdk-js-v3/lib/lib-storage
```

### Checklist

---


lib/lib-storage/src/s3-transfer-manager/S3TransferManager.ts
```diff
@@ -0,0 +1,1154 @@
...
+    }
+  }
+
+  /**
+   * Calculates part size and expected parts count.
+   *
+   * @internal
+   */
+  private calculatePartSize(contentLength: number): { partSize: number; expectedPartsCount: number } {
+    const partSize = Math.max(this.targetPartSizeBytes, Math.floor(contentLength / 10_000));
```
GarrettBeatty: pretty sure this needs to be Math.ceil(contentLength / 10_000). check out this issue https://github.com/aws/aws-sdk-net/issues/4332 you might want to add a test case for this

GarrettBeatty (commented):
added one comment. i can review rest of pr later if wanted

lib/lib-storage/src/s3-transfer-manager/S3TransferManager.ts
```diff
@@ -0,0 +1,1284 @@
...
+              return this.s3ClientInstance
+                .send(new GetObjectCommand(getObjectRequest), transferOptions)
+                .then((response) => {
+                  this.validatePartDownload(response.ContentRange, part, partSize ?? 0);
+                  if (response.Body && typeof (response.Body as any).getReader === "function") {
+                    const reader = (response.Body as any).getReader();
+                    (response.Body as any).getReader = function () {
+                      return reader;
+                    };
+                  }
```
kuhe, line 566: what is this override of getReader doing?

smilkuri: uses the same reader instance, when stream is already locked.

kuhe: that seems dangerous and counter to the point of having a lockable stream

lib/lib-storage/src/s3-transfer-manager/S3TransferManager.ts
```diff
@@ -0,0 +1,1285 @@
...
+ * Implements an eventTarget-based progress tracking system with methods to register,
+ * dispatch, and remove listeners for transfer lifecycle events.
+ *
+ * @alpha
+ */
+
+export class S3TransferManager implements IS3TransferManager {
+  private static MIN_PART_SIZE = 5 * 1024 * 1024; // 5MB
+  private static DEFAULT_PART_SIZE = 8 * 1024 * 1024; // 8MB
+  private static MIN_UPLOAD_THRESHOLD = 16 * 1024 * 1024; // 16MB
```
kuhe: are these readonly?

what is MIN_UPLOAD_THRESHOLD?

lib/lib-storage/src/s3-transfer-manager/S3TransferManager.ts
```diff
@@ -0,0 +1,1285 @@
...
+
+  private readonly s3ClientInstance: S3Client;
```
kuhe:
```suggestion
  private readonly s3: S3Client;
```

lib/lib-storage/src/s3-transfer-manager/S3TransferManager.ts
```diff
@@ -0,0 +1,1285 @@
...
+          this.validateUploadPart(dataPart, partSize);
+
+          const partRequest: UploadPartCommandInput = {
+            ...request,
+            Body: dataPart.data,
+            UploadId: uploadId,
+            PartNumber: dataPart.partNumber,
+            ContentLength: byteLength(dataPart.data),
+            ...(this.requestChecksumCalculation === "WHEN_SUPPORTED" && {
+              ChecksumAlgorithm: this.checksumAlgorithm,
```
kuhe: this overrides the request checksum algo

lib/lib-storage/src/s3-transfer-manager/S3TransferManager.ts
```diff
@@ -0,0 +1,1285 @@
...
+    transferOptions?: TransferOptions
+  ): Promise<CompleteMultipartUploadCommandOutput> {
+    const { partSize, expectedPartsCount } = this.calculatePartSize(contentLength);
+
+    this.checkAborted(transferOptions);
+
+    const createMpuRequest: CreateMultipartUploadCommandInput = { ...request };
+
+    const hasFullObjectChecksum =
+      request.ChecksumCRC32 || request.ChecksumCRC32C || request.ChecksumSHA1 || request.ChecksumSHA256;
```
kuhe: this isn't the right check, since request checksums can be provided by an arbitrary algorithm

see selectChecksumAlgorithmFunction.ts and the flexible checksum middleware implementation

lib/lib-storage/src/s3-transfer-manager/S3TransferManager.ts
```diff
@@ -0,0 +1,1285 @@
...
+
+    if (hasFullObjectChecksum) {
+      createMpuRequest.ChecksumType = "FULL_OBJECT";
+      createMpuRequest.ChecksumAlgorithm = request.ChecksumAlgorithm ?? this.checksumAlgorithm;
+    } else if (this.requestChecksumCalculation === "WHEN_SUPPORTED") {
+      createMpuRequest.ChecksumAlgorithm = this.checksumAlgorithm;
```
kuhe: why do these two branches differ in how ChecksumAlgorithm is assigned?

lib/lib-storage/src/s3-transfer-manager/S3TransferManager.ts
```diff
@@ -0,0 +1,1285 @@
...
+   * @internal
+   */
+  private async uploadSingleObject(
+    request: UploadRequest,
+    contentLength: number,
+    transferOptions?: TransferOptions
+  ): Promise<PutObjectCommandOutput> {
+    const putObjectRequest = { ...request };
+    if (this.requestChecksumCalculation === "WHEN_SUPPORTED") {
+      putObjectRequest.ChecksumAlgorithm = this.checksumAlgorithm;
```
kuhe: why is the request algorithm overridden?

lib/lib-storage/src/s3-transfer-manager/S3TransferManager.ts
```diff
@@ -0,0 +1,1285 @@
...
+      config.s3ClientInstance ??
+      new S3Client({
+        requestChecksumCalculation: this.requestChecksumCalculation,
+        responseChecksumValidation: this.responseChecksumValidation,
+      });
+
+    this.targetPartSizeBytes = config.targetPartSizeBytes ?? S3TransferManager.DEFAULT_PART_SIZE;
+    this.multipartUploadThresholdBytes = config.multipartUploadThresholdBytes ?? S3TransferManager.MIN_UPLOAD_THRESHOLD;
+
+    this.checksumAlgorithm = config.checksumAlgorithm ?? "CRC32";
```
kuhe: why is a default selected by this class? doesn't the S3 client have a default checksumAlgo?

lib/lib-storage/src/s3-transfer-manager/S3TransferManager.ts
```diff
@@ -0,0 +1,1285 @@
...
+  private readonly checksumAlgorithm: ChecksumAlgorithm;
+  private readonly multipartDownloadType: "PART" | "RANGE";
+  private readonly eventListeners: TransferEventListeners;
+  private readonly abortCleanupFunctions = new WeakMap<AbortSignal, () => void>();
+  private readonly maxConcurrentRequests: number;
+  private readonly maxConcurrentUploads: number;
+
+  public constructor(config: S3TransferManagerConfig = {}) {
+    this.requestChecksumCalculation = config.requestChecksumCalculation ?? "WHEN_SUPPORTED";
+    this.responseChecksumValidation = config.responseChecksumValidation ?? "WHEN_SUPPORTED";
```
kuhe, line 75: why does this class designate defaults for these options when the S3 client has defaults?

lib/lib-storage/src/s3-transfer-manager/S3TransferManager.ts
```diff
@@ -0,0 +1,1285 @@
...
+  private readonly targetPartSizeBytes: number;
+  private readonly multipartUploadThresholdBytes: number;
+  private readonly requestChecksumCalculation: "WHEN_SUPPORTED" | "WHEN_REQUIRED";
+  private readonly responseChecksumValidation: "WHEN_SUPPORTED" | "WHEN_REQUIRED";
+  private readonly checksumAlgorithm: ChecksumAlgorithm;
+  private readonly multipartDownloadType: "PART" | "RANGE";
+  private readonly eventListeners: TransferEventListeners;
+  private readonly abortCleanupFunctions = new WeakMap<AbortSignal, () => void>();
+  private readonly maxConcurrentRequests: number;
+  private readonly maxConcurrentUploads: number;
```
kuhe, line 68: explain the state variables.

what is targetPartSizeBytes? why is checksumAlgorithm a property of this class? 

what is the difference between maxConcurrentRequests and maxConcurrentUploads?

lib/lib-storage/src/s3-transfer-manager/S3TransferManager.ts
```diff
@@ -0,0 +1,1285 @@
...
+   * @param transferOptions - Optional abort signal and event listeners for transfer lifecycle monitoring.
+   *
+   * @returns S3 PutObject or CompleteMultipartUpload response with transfer event dispatching.
+   *
+   */
+  public async upload(request: UploadRequest, transferOptions?: TransferOptions): Promise<UploadResponse> {
+    const totalContentLength = request.ContentLength ?? byteLength(request.Body);
+
+    if (totalContentLength === undefined) {
+      throw new Error("Unable to determine content length for upload");
```
kuhe, line 286: this isn't required in our Upload class, why is it required here? is it from the specification?

lib/lib-storage/src/s3-transfer-manager/S3TransferManager.ts
```diff
@@ -0,0 +1,1285 @@
...
+
+    this.checkAborted(transferOptions);
+    this.addEventListeners(transferOptions?.eventListeners);
+    this.dispatchTransferInitiatedEvent(request, totalContentLength);
+
+    const removeLocalEventListeners = () => {
+      this.removeEventListeners(transferOptions?.eventListeners);
+      if (transferOptions?.abortSignal) {
+        this.abortCleanupFunctions.get(transferOptions.abortSignal as AbortSignal)?.();
+        this.abortCleanupFunctions.delete(transferOptions.abortSignal as AbortSignal);
```
kuhe, line 297: the function, when called, should remove itself

lib/lib-storage/src/s3-transfer-manager/types.ts
```diff
@@ -0,0 +1,356 @@
...
+ */
+export interface S3TransferManagerConfig {
+  /**
+   * The low level S3 client that will be used to send requests to S3.
+   */
+  s3ClientInstance?: S3Client;
+  /**
+   * The target part size to use in a multipart transfer. Does not apply to downloads if multipartDownloadType is PART.
+   */
+  targetPartSizeBytes?: number;
```
kuhe, line 33: what does target mean?

lib/lib-storage/src/s3-transfer-manager/S3TransferManager.ts
```diff
@@ -0,0 +1,1285 @@
...
+        requestChecksumCalculation: this.requestChecksumCalculation,
+        responseChecksumValidation: this.responseChecksumValidation,
+      });
+
+    this.targetPartSizeBytes = config.targetPartSizeBytes ?? S3TransferManager.DEFAULT_PART_SIZE;
+    this.multipartUploadThresholdBytes = config.multipartUploadThresholdBytes ?? S3TransferManager.MIN_UPLOAD_THRESHOLD;
+
+    this.checksumAlgorithm = config.checksumAlgorithm ?? "CRC32";
+    this.multipartDownloadType = config.multipartDownloadType ?? "PART";
+    this.maxConcurrentRequests = config.maxInMemoryParts ?? 8;
```
kuhe: why are these variables named differently

lib/lib-storage/src/s3-transfer-manager/S3TransferManager.ts
```diff
@@ -0,0 +1,1285 @@
...
+    const uploadId = createMpuResponse.UploadId!;
+
+    const abortController = new AbortController();
+    const uploadAbortSignal = abortController.signal;
+
+    try {
+      const completedParts: CompletedPart[] = [];
+      const dataFeeder = getChunk(request.Body, partSize);
+      let bytesUploaded = 0;
+      const queueSize = this.maxConcurrentUploads;
```
kuhe: it's not a queue since no order is implied, use `const concurrentLimit` or something

lib/lib-storage/src/s3-transfer-manager/S3TransferManager.ts
```diff
@@ -0,0 +1,1285 @@
...
+              snapshot: {
+                transferredBytes: bytesUploaded,
+                totalBytes: contentLength,
+              },
+            })
+          );
+        }
+      };
+
+      const concurrentUploaders: Promise<void>[] = [];
```
kuhe:
```suggestion
      const partUploads: Promise<void>[] = [];
```

lib/lib-storage/src/s3-transfer-manager/S3TransferManager.ts
```diff
@@ -0,0 +1,1285 @@
...
+      await Promise.all(concurrentUploaders);
+
+      if (completedParts.length !== expectedPartsCount) {
+        throw new Error(`Expected ${expectedPartsCount} parts but uploaded ${completedParts.length} parts`);
+      }
+
+      completedParts.sort((a, b) => a.PartNumber! - b.PartNumber!);
+
+      this.checkAborted(transferOptions);
+      const completeRequest: any = {
```
kuhe: type is CompleteMultipartUploadCommandInput, why `any`?

lib/lib-storage/src/s3-transfer-manager/S3TransferManager.ts
```diff
@@ -0,0 +1,1285 @@
...
+        MpuObjectSize: contentLength,
+      };
+      delete completeRequest.Body;
+
+      if (hasFullObjectChecksum) {
+        completeRequest.ChecksumType = "FULL_OBJECT";
+        if (request.ChecksumCRC32) completeRequest.ChecksumCRC32 = request.ChecksumCRC32;
+        if (request.ChecksumCRC32C) completeRequest.ChecksumCRC32C = request.ChecksumCRC32C;
+        if (request.ChecksumSHA1) completeRequest.ChecksumSHA1 = request.ChecksumSHA1;
+        if (request.ChecksumSHA256) completeRequest.ChecksumSHA256 = request.ChecksumSHA256;
```
kuhe: missing deprecated MD5 and CRC64NVME

lib/lib-storage/src/s3-transfer-manager/S3TransferManager.ts
```diff
@@ -0,0 +1,1285 @@
...
+      }
+
+      try {
+        return await this.s3ClientInstance.send(new CompleteMultipartUploadCommand(completeRequest), transferOptions);
+      } catch (completeError) {
```
kuhe:
```suggestion
      } catch (completeMpuError) {
```

lib/lib-storage/src/s3-transfer-manager/S3TransferManager.ts
```diff
@@ -0,0 +1,1285 @@
...
+        console.warn(`CompleteMultipartUpload failed for upload ID ${uploadId}: ${(completeError as Error).message}`);
+        throw completeError;
+      }
+    } catch (error) {
+      const abortRequest: any = {
```
kuhe: use the AbortMultipartUploadCommandInput type

lib/lib-storage/src/s3-transfer-manager/S3TransferManager.ts
```diff
@@ -0,0 +1,1285 @@
...
+        ...request,
+        UploadId: uploadId,
+      };
+      delete abortRequest.Body;
+
+      try {
+        await this.s3ClientInstance.send(new AbortMultipartUploadCommand(abortRequest), transferOptions);
+      } catch (abortError) {
+        console.warn(`Failed to abort multipart upload ${uploadId}:`, abortError);
```
kuhe: use a Logger that can be given to the S3TM. By default use `new LogLevel("warn")` from LogLevel.ts

lib/lib-storage/src/s3-transfer-manager/S3TransferManager.ts
```diff
@@ -0,0 +1,1285 @@
...
+            ETag: partResponse.ETag,
+          };
+          if (partResponse.ChecksumCRC32) completedPart.ChecksumCRC32 = partResponse.ChecksumCRC32;
+          if (partResponse.ChecksumCRC32C) completedPart.ChecksumCRC32C = partResponse.ChecksumCRC32C;
+          if (partResponse.ChecksumSHA1) completedPart.ChecksumSHA1 = partResponse.ChecksumSHA1;
+          if (partResponse.ChecksumSHA256) completedPart.ChecksumSHA256 = partResponse.ChecksumSHA256;
+
+          completedParts.push(completedPart);
+
+          const partLength = byteLength(dataPart.data) || 0;
```
kuhe: what does it mean for a part to have zero length? isn't that an exception?

lib/lib-storage/src/s3-transfer-manager/S3TransferManager.ts
```diff
@@ -0,0 +1,1285 @@
...
+            Body: dataPart.data,
+            UploadId: uploadId,
+            PartNumber: dataPart.partNumber,
+            ContentLength: byteLength(dataPart.data),
+            ...(this.requestChecksumCalculation === "WHEN_SUPPORTED" && {
+              ChecksumAlgorithm: this.checksumAlgorithm,
+            }),
+          };
+
+          const partResponse = await this.s3ClientInstance.send(new UploadPartCommand(partRequest), transferOptions);
```
kuhe: we need an error message for the case where something tries to use this in the browser and the checksum headers aren't enabled in CORS

lib/lib-storage/src/s3-transfer-manager/S3TransferManager.ts
```diff
@@ -0,0 +1,1285 @@
...
+        this.abortCleanupFunctions.get(transferOptions.abortSignal as AbortSignal)?.();
+        this.abortCleanupFunctions.delete(transferOptions.abortSignal as AbortSignal);
+      }
+    };
+
+    try {
+      let response: UploadResponse;
+
+      if (totalContentLength < this.multipartUploadThresholdBytes) {
+        response = await this.uploadSingleObject(request, totalContentLength, transferOptions);
```
kuhe: this method name doesn't contrast with the MPU alternative. A multi part upload is still a single object.

names:
```
putObject / uploadSinglePart
uploadMultipart / uploadInParts
```

lib/lib-storage/src/s3-transfer-manager/S3TransferManager.ts
smilkuri: This is the default value for threshold that determines when to use single part upload and multipart upload. If content length is greater than this threshold we use multipart upload.

lib/lib-storage/src/s3-transfer-manager/S3TransferManager.ts
smilkuri: Intention was not to override it, will modify it to `ChecksumAlgorithm: request.ChecksumAlgorithm ?? this.checksumAlgorithm,`

lib/lib-storage/src/s3-transfer-manager/S3TransferManager.ts
smilkuri: `ChecksumAlgorithm: either the SDK default or the algorithm specified by the user in the request`. Understood it in a different way like hardcoded default to be specified but yes it should get default from client's config.

lib/lib-storage/src/s3-transfer-manager/S3TransferManager.ts
smilkuri: The spec has default values for these and didn't mention anything like SDK default can be used.

lib/lib-storage/src/s3-transfer-manager/S3TransferManager.ts
smilkuri: According to spec `This API uploads a local file or stream with known content length to S3` only for known content length

lib/lib-storage/src/s3-transfer-manager/types.ts
smilkuri: The size of a part user wants

lib/lib-storage/src/s3-transfer-manager/S3TransferManager.ts
smilkuri: - targetPartSizeBytes // Size of a part user wants to use for transfer

- maxConcurrentRequests // This used only for downloads which determines max no of parts that can be downloaded concurrently or call it like max no of getObject requests that can be made.

- maxConcurrentUploads // This is for upload, max no of parts that can be uploaded concurrently

lib/lib-storage/src/s3-transfer-manager/S3TransferManager.ts
smilkuri: You mean instead of using a static variable for defaults?

lib/lib-storage/src/s3-transfer-manager/S3TransferManager.ts
smilkuri: Initially added it as a fallback for cases where `byteLength()` might return undefined. Removed it since we have undefined check at the top.

lib/lib-storage/src/s3-transfer-manager/types.ts
kuhe: No, why is it called "target"? Explain to the user why the word target is here. Why is it not "partSizeBytes"?

lib/lib-storage/src/s3-transfer-manager/S3TransferManager.ts
kuhe: maxConcurrentRequests should be renamed. Uploads are requests too.

lib/lib-storage/src/s3-transfer-manager/S3TransferManager.ts
kuhe: the left and right side have different variable names to refer to the same value, why is that?

lib/lib-storage/src/s3-transfer-manager/types.ts
smilkuri: Updated the comment.

lib/lib-storage/src/s3-transfer-manager/S3TransferManager.ts
smilkuri: renamed it to maxInMemoryParts according to specification wording

kuhe: maxInMemoryParts is misleading. Are the parts fully buffered in memory? I think not.

And it only applies to downloads? 

It should be called `maxConcurrentDownloads`

GarrettBeatty: +1 there is two separate things. maxConcurrentDownloads (number of http requests) and maxInMemoryParts (number of parts buffered in memory) (if applicable)

smilkuri: +1 to Garrett's comment. Yeah, was a bit confused here.

smilkuri: Removed `maxInMemoryParts` for downloads because we only initiate the next request after a stream is consumed. So the number of buffered parts can't exceed `maxConcurrentDownloads`.

lib/lib-storage/package.json
```diff
@@ -30,7 +32,9 @@
   },
   "license": "Apache-2.0",
   "dependencies": {
+    "@aws-sdk/config": "workspace:3.1058.0",
```
kuhe, line 35: don't use internally, import the implementation pkg

samuellagrene90-cpu (approved):

kuhe (changes requested):
will be split into multiple PRs, blocking this

PR #7762, opened 2026-02-21
fix(lib-dynamodb): clone middleware stack to support cacheMiddleware

TrevorBurnham:
## Problem

`DynamoDBDocumentClient` shares the parent `DynamoDBClient`'s `middlewareStack` by direct reference:

```ts
this.middlewareStack = client.middlewareStack;
```

This means any middleware added to either client affects the other. Because of this, `cacheMiddleware: true` was explicitly blocked with a runtime error: Cached handlers on the parent could become stale if the document client mutated the shared stack.

## Solution

Clone the stack instead of sharing it:

```ts
this.middlewareStack = client.middlewareStack.clone();
```

This makes the two clients independent. The document client gets a snapshot of the parent's middleware at construction time, and subsequent mutations on either side don't cross over. The `cacheMiddleware` restriction is no longer needed and is removed.

The same change is applied to the codegen template (`DocumentBareBonesClientGenerator.java`) so that `yarn generate-clients` won't revert it.

## Testing

Added `DynamoDBDocumentClient.spec.ts` with 4 tests:

- `DynamoDBDocumentClient` allows initialization with `cacheMiddleware: true`
- `DynamoDBDocumentClient` clones the middleware stack (not same reference)
- Adding middleware to the document client does not affect the parent
- `DynamoDBDocument` (subclass) allows initialization with `cacheMiddleware: true`

## Benchmark Context

Total middleware stack rebuild cost (measured with a typical S3-like stack of 15 client middleware + 1 command middleware) with and without `cacheMiddleware`:

```
Requests     Uncached (ms)     With cacheMiddleware     Speedup
---------------------------------------------------------------
100          2.34              0.02 ms                  133x
1000         10.56             0.01 ms                  928x
3000         26.54             0.02 ms                  1468x
```

This PR doesn't change the default value of `cacheMiddleware` (still `false`), but it removes the blocker that prevented DynamoDB users from opting in.

## Behavioral Change

Previously, middleware added to the parent `DynamoDBClient` after creating a `DynamoDBDocumentClient` was visible to the document client (and vice versa) due to the shared reference. This is no longer the case — the two stacks are independent after construction.

This was never documented behavior, but it is _technically_ a breaking change: Post-construction middleware changes on one client no longer affect the other.

## Note on Shared Config

The constructor still shares the `config` object by reference (`this.config = client.config`) and mutates it (`this.config.translateConfig = translateConfig`). This means the parent `DynamoDBClient`'s resolved config ends up with a `translateConfig` property it never asked for. This is a pre-existing issue not introduced by this PR, but it means the "independence" story between the two clients is not yet complete. A follow-up could shallow-copy the config as well.

### DIFF
--- a/codegen/smithy-aws-typescript-codegen/src/main/java/software/amazon/smithy/aws/typescript/codegen/DocumentBareBonesClientGenerator.java
+++ b/codegen/smithy-aws-typescript-codegen/src/main/java/software/amazon/smithy/aws/typescript/codegen/DocumentBareBonesClientGenerator.java
@@ -192,15 +192,7 @@ final class DocumentBareBonesClientGenerator implements Runnable {
                 writer.write("super(client.config);");
                 writer.write("this.config = client.config;");
                 writer.write("this.config.translateConfig = translateConfig;");
-                writer.write("this.middlewareStack = client.middlewareStack;");
-                writer.write("""
-                             if (this.config?.cacheMiddleware) {
-                                 throw new Error(
-                                     "@aws-sdk/lib-dynamodb - cacheMiddleware=true is not compatible with the"
-                                       + " DynamoDBDocumentClient. This option must be set to false."
-                                 );
-                             }
-                             """);
+                writer.write("this.middlewareStack = client.middlewareStack.clone();");
                 writer.popState();
             }
         );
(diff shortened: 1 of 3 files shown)


kuhe:
we are not accepting this change due to the risk of regression for applications utilizing the shared reference

PR #7763, opened 2026-02-22
fix(cloudfront-signer): url-encode special characters in URL path

### LINKED ISSUE #7571: CloudFront URL signer no longer URL-encodes spaces in S3 object keys after PR #7237

### Summary

After PR #7237, the CloudFront URL signer has stopped URL-encoding spaces in S3 object keys, causing a breaking change in signed URL generation.

### Background

Prior to PR #7237, when signing CloudFront URLs for S3 objects containing spaces in their keys (e.g., "My File.pdf"), the signer would automatically URL-encode the space as %20, producing a valid signed URL like `https://example.cloudfront.net/My%20File.pdf?....`

### Current Behavior

After PR #7237, the CloudFront signer returns unencoded spaces in the URL (e.g., `https://example.cloudfront.net/My File.pdf?...`), which produces invalid URLs that fail in browsers and HTTP clients.

### Expected Behavior

The CloudFront signer should either:

1. Automatically URL-encode special characters (including spaces) in object keys, OR
2. Clearly document that clients must URL-encode keys before passing them to the signer

### Impact

This is a breaking change that affects:

- Existing applications relying on automatic URL encoding
- S3 objects with spaces or special characters in their keys
- Any workflow where user-provided filenames are used as S3 keys

Steps to ### Reproduce
```
// Example code demonstrating the issue
const signer = new CloudFrontUrlSigner(/* config */);
const url = signer.getSignedUrl({
  url: 'https://example.cloudfront.net/My File.pdf',
  // ... other params
});
console.log(url); // Contains unencoded space
```

### Environment

AWS SDK version: [post-PR #7237]
Node.js version: [v22]

### Proposed Solution
Either revert the behavior to automatically URL-encode keys, or update the documentation to explicitly state that URL encoding is the caller's responsibility.

### Regression Issue

### SDK version number

@aws-sdk/cloudfront-signer@3.946.0

Node.js

v22.15.0

### Reproduction Steps

```
// Example code demonstrating the issue
const signer = new CloudFrontUrlSigner(/* config */);
const url = signer.getSignedUrl({
  url: 'https://example.cloudfront.net/My File.pdf',
  // ... other params
});
console.log(url); // Contains unencoded space
```

### Observed Behavior

`https://example.cloudfront.net/My File.pdf?...`

### Expected Behavior

`https://example.cloudfront.net/My%20File.pdf?....`

### Possible Solution

Either revert the behavior to automatically URL-encode keys, or update the documentation to explicitly state that URL encoding is the caller's responsibility.

_No response_

TrevorBurnham:
Fixes #7571

## Problem

After PR #7237 (`fix(cloudfront-signer): do not use URL Object`), the CloudFront URL signer stopped URL-encoding spaces and other special characters in the URL path. PR #7237 correctly fixed an issue where `new URL()` was normalizing `.` and `..` path segments, but the switch to raw string manipulation also removed the implicit URL-encoding that `new URL()` provided.

This means URLs like `https://example.cloudfront.net/My File.pdf` are returned with unencoded spaces, producing invalid URLs.

## Solution

Added an `encodeUrlPath` helper function that:

1. Parses the URL into origin, path, and query components using string operations (no `new URL()`)
2. Encodes each path segment individually with `encodeURIComponent`, preserving `/` separators
3. Decodes before re-encoding to prevent double-encoding of already percent-encoded sequences (e.g., `%20` stays `%20`, not `%2520`)
4. Re-encodes query parameters using `URLSearchParams`

This replaces the previous `encodeBaseUrlQuery` function which only handled query parameter encoding but not path encoding.

The fix preserves the behavior from PR #7237: As confirmed by the tests, `.` and `..` path segments are not normalized.

### DIFF
--- a/packages/cloudfront-signer/src/sign.spec.ts
+++ b/packages/cloudfront-signer/src/sign.spec.ts
@@ -383,6 +383,46 @@ describe("getSignedUrl", () => {
       expect(verifySignature(signatureQueryParam, policy)).toBeTruthy();
     });
   });
+
+  it("should URL-encode spaces in the path", () => {
+    const urlWithSpaces = "https://d111111abcdef8.cloudfront.net/private content/My File.pdf";
+    const result = getSignedUrl({
+      url: urlWithSpaces,
+      keyPairId,
+      dateLessThan,
+      privateKey,
+      passphrase,
+    });
+    expect(result).toContain("private%20content/My%20File.pdf");
+    expect(result).not.toContain("private content");
+    expect(result).not.toContain("My File.pdf");
+  });
+
+  it("should URL-encode special characters in the path while preserving path structure", () => {
+    const urlWithSpecialChars = "https://d111111abcdef8.cloudfront.net/path with spaces/file[1].pdf";
+    const result = getSignedUrl({
+      url: urlWithSpecialChars,
+      keyPairId,
+      dateLessThan,
+      privateKey,
+      passphrase,
+    });
+    expect(result).toContain("path%20with%20spaces");
+    expect(result).not.toContain("path with spaces");
+  });
+
+  it("should not double-encode already-encoded URLs", () => {
+    const alreadyEncodedUrl = "https://d111111abcdef8.cloudfront.net/private%20content/My%20File.pdf";
+    const result = getSignedUrl({
+      url: alreadyEncodedUrl,
+      keyPairId,
+      dateLessThan,
+      privateKey,
+      passphrase,
+    });
+    expect(result).toContain("private%20content/My%20File.pdf");
+    expect(result).not.toContain("%2520");
+  });
 });
 
 describe("getSignedCookies", () => {
--- a/packages/cloudfront-signer/src/sign.spec.ts
+++ b/packages/cloudfront-signer/src/sign.spec.ts
@@ -407,8 +407,8 @@ describe("getSignedUrl", () => {
       privateKey,
       passphrase,
     });
-    expect(result).toContain("path%20with%20spaces");
-    expect(result).not.toContain("path with spaces");
+    expect(result).toContain("file%5B1%5D.pdf");
+    expect(result).not.toContain("file[1].pdf");
   });
 
   it("should not double-encode already-encoded URLs", () => {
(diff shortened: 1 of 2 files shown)


clemsonfan101 (approved):

packages/cloudfront-signer/src/sign.spec.ts
```diff
@@ -383,6 +383,46 @@ describe("getSignedUrl", () => {
...
+    const urlWithSpecialChars = "https://d111111abcdef8.cloudfront.net/path with spaces/file[1].pdf";
+    const result = getSignedUrl({
+      url: urlWithSpecialChars,
+      keyPairId,
+      dateLessThan,
+      privateKey,
+      passphrase,
+    });
+    expect(result).toContain("path%20with%20spaces");
+    expect(result).not.toContain("path with spaces");
```
trivikr: This should test special characters and not space
```suggestion
    expect(result).toContain("file%5B1%5D.pdf");
    expect(result).not.toContain("file[1].pdf");
```

packages/cloudfront-signer/src/sign.ts
```diff
@@ -248,6 +237,53 @@ interface CloudfrontAttributes {
...
+    }
+  });
+
+  // Re-encode query parameters if present
+  let encodedQuery = "";
+  if (rawQuery) {
+    const params = [...new URLSearchParams(rawQuery.slice(1)).entries()]
+      .map(([key, value]) => `${encodeURIComponent(key)}=${encodeURIComponent(value)}`)
+      .join("&");
+    encodedQuery = `?${params}`;
```
trivikr: nit: reduce object allocations as an improvement.
```suggestion
    for (const [key, value] of new URLSearchParams(rawQuery.slice(1)).entries()) {
      encodedQuery += `${encodedQuery ? "&" : "?"}${encodeURIComponent(key)}=${encodeURIComponent(value)}`;
    }
```

I know this is existing code, but we've been reducing object allocations in JS SDK recently.

trivikr:
CI is being run in https://github.com/aws/aws-sdk-js-v3/pull/7997

trivikr (approved):
CI is successful in https://github.com/aws/aws-sdk-js-v3/pull/7997

PR #7764, opened 2026-02-22
fix(util-dynamodb): replace InstanceType that resolves to any in NativeAttributeValue

### LINKED ISSUE #7526: `NativeAttributeValue` type is `any`

### Describe the bug

The exported type `NativeAttributeValue` is resolved (in my IDE) to `any`, so my inputs are not validated

### Regression Issue

### SDK version number

@aws-sdk/lib-dynamodb@3.939.0, @aws-sdk/util-dynamodb@3.939.0

Node.js

v22.18.0

### Reproduction Steps

```js
import { NativeAttributeValue } from '@aws-sdk/lib-dynamodb';
```

### Observed Behavior

When hovering the type, I'm seeing `type NativeAttributeValue = any`

### Expected Behavior

I should see a more complex type, as it seems to be intended 

something like `type NativeAttributeValue = NativeScalarAttributeValue | {
    [key: string]: NativeAttributeValue;
} | NativeAttributeValue[] | ...`

### Possible Solution

that seems to be caused by this part of the type: `InstanceType<{ new (...args: any[]): any; }>`

_No response_

TrevorBurnham:
Fixes #7526

## Problem

The `NativeAttributeValue` type in `@aws-sdk/util-dynamodb` resolves to `any` in IDEs and type checking. This is because the union includes:

```ts
InstanceType<{ new (...args: any[]): any }>
```

which TypeScript evaluates to `any`, poisoning the entire union type. As a result, users get no type validation when using `NativeAttributeValue`.

## Solution

Replace `InstanceType<{ new (...args: any[]): any }>` with `object`.

The intent of that union member was to accept class instances when `convertClassInstanceToMap` is enabled. At the type level, class instances are `object`s, so `object` correctly represents this without collapsing the union to `any`.

The `object` type in TypeScript represents all non-primitive values, which includes class instances, plain objects, arrays, etc. Since the other union members already cover arrays, records, and Sets specifically, the `object` member serves as the catch-all for class instances as originally intended.

### DIFF
--- a/packages/util-dynamodb/src/models.spec.ts
+++ b/packages/util-dynamodb/src/models.spec.ts
@@ -1,0 +1,34 @@
+import { describe, expect, it } from "vitest";
+
+import type { NativeAttributeValue } from "./models";
+
+describe("NativeAttributeValue type", () => {
+  it("should not resolve to any (should reject values not in the union)", () => {
+    // This test validates that NativeAttributeValue is not `any`.
+    // If the type were `any`, assigning a symbol would be allowed at compile time.
+    // With a proper union type, symbols are not assignable to NativeAttributeValue.
+    //
+    // We verify at runtime that the known members of the union are accepted.
+    const stringVal: NativeAttributeValue = "hello";
+    const numberVal: NativeAttributeValue = 42;
+    const boolVal: NativeAttributeValue = true;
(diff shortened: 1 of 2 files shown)


kuhe:
will be included in https://github.com/aws/aws-sdk-js-v3/issues/8154

PR #7765, opened 2026-02-22
fix(client-config-service): correct RecorderStatus enum values to match API response

TrevorBurnham:
## Problem

The `RecorderStatus` enum in `@aws-sdk/client-config-service` defines values in TitleCase:

```ts
export const RecorderStatus = {
  Failure: "Failure",
  NotApplicable: "NotApplicable",
  Pending: "Pending",
  Success: "Success",
} as const;
```

However, the actual AWS Config API returns UPPER_SNAKE_CASE values: `"SUCCESS"`, `"FAILURE"`, `"PENDING"`, `"NOT_APPLICABLE"`. This means comparisons like `status === RecorderStatus.Success` never match the real API response.

## Solution

This PR updates the enum values in the Smithy service model (`codegen/sdk-codegen/aws-models/config-service.json`) to change `smithy.api#enumValue` from TitleCase to UPPER_SNAKE_CASE. The other changes are auto-generated from that change.

The enum keys remain unchanged (e.g., `RecorderStatus.Success`), so existing code using the keys will continue to work, but the values they resolve to now match what the API actually returns.

### DIFF
--- a/clients/client-config-service/src/commands/DescribeConfigurationRecorderStatusCommand.ts
+++ b/clients/client-config-service/src/commands/DescribeConfigurationRecorderStatusCommand.ts
@@ -63,7 +63,7 @@ export interface DescribeConfigurationRecorderStatusCommandOutput extends Descri
  * //       lastStartTime: new Date("TIMESTAMP"),
  * //       lastStopTime: new Date("TIMESTAMP"),
  * //       recording: true || false,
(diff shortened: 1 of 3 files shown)


kuhe:
service model changes must come from the internal release process, please request for the service team to initiate the change, or contact me internally if you have questions on how to do that.

PR #7767, opened 2026-02-22
fix(compilation): disable Rollup externalLiveBindings to fix Jest auto-mocking

### LINKED ISSUE #7748: AWS SDK v3: Lazy `Object.defineProperty` re-exports break Jest auto-mocking

## Summary

This bug is related to using Jest in combination with aws-sdk, so apologies if this is the wrong place for this, but a recently release change in @aws-sdk/client-dynamodb broke some tests.

Starting in `@aws-sdk/client-dynamodb` 3.988.0, exception classes like `ConditionalCheckFailedException` are re-exported from the package entrypoint via lazy `Object.defineProperty` getters. When Jest's `jest.mock()` auto-mock encounters these getters, it fails to resolve them — the exports become `undefined` at runtime.

This affects any test that auto-mocks an SDK client module and then references an exception class or other re-exported symbol from that same module.

## Question
Maybe this isn't a bug, but intentional?
Is the switch to lazy `Object.defineProperty` getters for re-exported symbols (exception classes, command classes, etc.) intentional and permanent? Or is this an unintended side effect of a build tooling change?

### Regression Issue

### SDK version number

@aws-sdk/package-name@3.988.0

Node.js

v22.17.0

### Reproduction Steps

```ts
import { ConditionalCheckFailedException } from "@aws-sdk/client-dynamodb";

jest.mock("@aws-sdk/client-dynamodb");

test("exception class is defined", () => {
  // Fails: ConditionalCheckFailedException is undefined
  expect(ConditionalCheckFailedException).toBeDefined();
});
```

**Environment:**
- `@aws-sdk/client-dynamodb` 3.988.0 ... 3.992.0 (last working version was 3.987.0)
- `jest` 29.7.0 (also reproduced on 30.1.3)
- `ts-jest` 29.2.5
- Node 22

### Observed Behavior

ConditionalCheckFailedException is undefined

### Expected Behavior

ConditionalCheckFailedException should be defined

### Possible Solution

_No response_

## Workaround

Use a manual mock factory with `jest.requireActual()` to preserve real exports:

```ts
const mockSend = jest.fn();

jest.mock("@aws-sdk/client-dynamodb", () => {
  const actual = jest.requireActual("@aws-sdk/client-dynamodb");
  return {
    ...actual,
    DynamoDBClient: jest.fn().mockImplementation(() => ({
      send: (...args: any[]) => mockSend(...args),
    })),
  };
});
```

TrevorBurnham:
Fixes #7748

### Problem

When using `jest.mock("@aws-sdk/client-dynamodb")` (or any SDK client), Jest's auto-mock snapshots the module's exports at mock-creation time. However, the CJS dist bundle (`dist-cjs/index.js`) re-exports symbols from variant-external modules using lazy `Object.defineProperty` getters:

```js
Object.keys(schemas_0).forEach(function (k) {
  if (k !== 'default' && !Object.prototype.hasOwnProperty.call(exports, k))
    Object.defineProperty(exports, k, {
      enumerable: true,
      get: function () { return schemas_0[k]; }
    });
});
```

These getters haven't resolved when Jest snapshots the exports, so all re-exported symbols (models, enums, exceptions from `schemas` and `errors` modules) become `undefined` in tests.

This affects any client that has react-native variant files, because the Inliner keeps variant files and their transitive dependencies as Rollup externals — and Rollup's default behavior is to use live-binding getters for external re-exports.

### Root Cause

The CJS inliner (`scripts/compilation/Inliner.js`) uses Rollup to bundle each client. Files with `.native.js` counterparts (e.g. `runtimeConfig`) are kept external so react-native bundlers can swap them. Their transitive `require()` targets (e.g. `schemas_0`, `errors`) are also externalized.

Rollup's default `externalLiveBindings: true` emits `Object.defineProperty` getters for these external re-exports. This is designed for ESM live-binding semantics, but is unnecessary for CJS and breaks Jest's auto-mocking.

### Solution

Set `externalLiveBindings: false` in the Rollup output options. This makes Rollup emit eager assignments instead:

```js
Object.keys(schemas_0).forEach(function (k) {
  if (k !== 'default' && !Object.prototype.hasOwnProperty.call(exports, k))
    exports[k] = schemas_0[k];
});
```

This is semantically equivalent for CJS (live bindings only matter for ESM circular imports) and allows Jest to see all exports immediately when auto-mocking.

From the [Rollup documentation](https://rollupjs.org/configuration-options/#output-externallivebindings):
> When set to false, Rollup will not generate code to support live bindings for external imports but instead assume that exports do not change over time. This will enable Rollup to generate more optimized code.

### Changes

- `scripts/compilation/Inliner.js` — add `externalLiveBindings: false` to Rollup `outputOptions`
- `scripts/compilation/Inliner.spec.js` — new test verifying the option is set and output has no lazy getters

### DIFF
--- a/scripts/compilation/Inliner.spec.js
+++ b/scripts/compilation/Inliner.spec.js
@@ -1,0 +1,54 @@
+const assert = require("assert");
+const path = require("path");
+const fs = require("fs");
+
+/**
+ * Verify that the Inliner's Rollup output uses eager bindings
+ * for external re-exports (externalLiveBindings: false).
+ *
+ * When externalLiveBindings is true (the default), Rollup emits:
+ *   Object.defineProperty(exports, k, { enumerable: true, get: function () { return ext[k]; } });
+ *
+ * These lazy getters break Jest's auto-mocking because jest.mock()
+ * snapshots exports at mock-creation time, before the getters resolve.
+ *
+ * With externalLiveBindings: false, Rollup emits:
+ *   exports[k] = ext[k];
+ *
+ * which Jest can auto-mock correctly.
+ *
+ * See: https://github.com/aws/aws-sdk-js-v3/issues/7748
+ */
+
+// Test 1: Verify the Inliner source contains externalLiveBindings: false
+const inlinerSource = fs.readFileSync(path.join(__dirname, "Inliner.js"), "utf-8");
+assert.ok(
+  inlinerSource.includes("externalLiveBindings: false"),
+  "Inliner.js must set externalLiveBindings: false in Rollup outputOptions"
+);
+
+// Test 2: If a built client-dynamodb dist-cjs/index.js exists, verify no lazy getter re-exports
+const dynamoIndexPath = path.join(__dirname, "..", "..", "clients", "client-dynamodb", "dist-cjs", "index.js");
+if (fs.existsSync(dynamoIndexPath)) {
+  const indexContents = fs.readFileSync(dynamoIndexPath, "utf-8");
+
+  // Match the lazy getter pattern Rollup emits with externalLiveBindings: true
+  const lazyGetterPattern = /Object\.defineProperty\(exports,\s*k,\s*\{[^}]*get:\s*function/g;
+  const matches = indexContents.match(lazyGetterPattern);
+
+  assert.strictEqual(
+    matches,
+    null,
+    "dist-cjs/index.js must not contain lazy Object.defineProperty getter re-exports. " +
+      "Found " +
+      (matches?.length ?? 0) +
+      " occurrences. " +
+      "Ensure externalLiveBindings: false is set in Rollup outputOptions."
+  );
+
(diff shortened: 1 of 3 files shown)


scripts/compilation/Inliner.spec.js
```diff
@@ -0,0 +1,54 @@
...
+      " occurrences. " +
+      "Ensure externalLiveBindings: false is set in Rollup outputOptions."
+  );
+
+  console.log("PASS: dist-cjs/index.js has no lazy getter re-exports.");
+} else {
+  console.log("SKIP: client-dynamodb dist-cjs/index.js not found (not built).");
+}
+
+console.log("PASS: Inliner.js has externalLiveBindings: false.");
```
kuhe, line 54: please add execution of the test to the `make test-integration` target in `Makefile`.

TrevorBurnham: Added!

PR #7768, opened 2026-02-22
test(build): run partial package build prior to snapshot-write

kuhe:
### Issue
n/a

### Description
This adds back snapshot-write during codegen. It calls `tpk` before `snapshot-write`. This builds the core packages (82) and allows snapshotting to occur, since it relies on the client runtimes.

The clients are compiled from source by vitest, but the core needs to be prebuilt. 

### Checklist
  - [x] It's not a feature.
  - [x] No streams here.

---


kuhe:
This PR does not enable snapshots for any production clients, but I experimented with them.

On remote, writing 425 clients worth of snapshots takes about 60-90s. Comparison mode takes about the same time.

There are 18,000+ test cases, which is expected due to the total number of operations being 1-1 with snapshots. The Node.js process runs out of memory on the last four clients due to the clients' source files being part of the vitest (esbuild) build. 

We will either want to not snapshot all clients, or break the test into multiple parts, or increase the available memory.

PR #7772, opened 2026-02-23
feat(client-iam): default region to us-east-1 for global IAM service

### LINKED ISSUE #6343: Required region in IAM Client

- [X] I've gone through [Developer Guide](https://docs.aws.amazon.com/sdk-for-javascript/v3/developer-guide) and [API reference](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest)
- [X] I've checked [AWS Forums](https://forums.aws.amazon.com) and [StackOverflow](https://stackoverflow.com/questions/tagged/aws-sdk-js).
- [X] I've searched for [previous similar issues](https://github.com/aws/aws-sdk-js-v3/issues) and didn't find any solution.

### Describe the bug

In aws-sdk v2, the IAM region was not required, but in v3, an error occurs "Error: Region is missing" if there is no region.
same issue https://github.com/aws/aws-sdk-js-v3/issues/1995, but I think it isn't fixed.

### SDK version number

@aws-sdk/client-iam@3.620.0

Node.js

v18.18.2

### Reproduction Steps

execute source code
`async function iam_summary_test() {
    const iam = new IAMClient({ credentials: makeCredentials() })
    const command = new GetAccountAuthorizationDetailsCommand({});
    const data = await iam.send(command);

    return data
}
 `

### Observed Behavior

Throw error `Region is missing`

### Expected Behavior

Work properly

### Possible Solution

_No response_

_No response_

TrevorBurnham:
Closes #6343

### Problem

When creating an `IAMClient` without specifying a region, the SDK throws `"Region is missing"`. This is confusing because IAM is a global service — it always routes to a global endpoint (`iam.global.api.aws`) regardless of which region you specify. The region is only used internally to determine the AWS partition (aws, aws-cn, aws-us-gov).

Users coming from AWS SDK v2 (where `new AWS.IAM()` worked without a region) or from the AWS CLI (where `aws iam list-users` works without a region) are surprised by this behavior.

### Solution

Added a post-codegen customization script (`scripts/generate-clients/customizations/iam-global-region.js`) that modifies the generated runtime configs to default the region to `us-east-1` when no region is configured. The customization is invoked from both `single-service.js` and `index.js`, so it survives codegen regeneration.

The customization applies to both runtime configs:

- **Node.js** (`runtimeConfig.ts`): Wraps the region config loader in a try/catch that falls back to `"us-east-1"` when no region is found in env vars or shared config
- **Browser** (`runtimeConfig.browser.ts`): Replaces `invalidProvider("Region is missing")` with `"us-east-1"`

The `us-east-1` default is appropriate because:
1. IAM's endpoint ruleset uses the region solely to determine the partition via `aws.partition(Region)`
2. `us-east-1` maps to the `aws` partition, which routes to `iam.global.api.aws`
3. Users targeting `aws-cn` or `aws-us-gov` partitions will always have a region configured (since all services in those partitions are regional)

This follows the same post-codegen customization pattern used by `workspaces-thin-client.js`.

---

### DIFF
--- a/clients/client-iam/src/runtimeConfig.browser.ts
+++ b/clients/client-iam/src/runtimeConfig.browser.ts
@@ -6,7 +6,6 @@ import { Sha256 } from "@aws-crypto/sha256-browser";
 import { createDefaultUserAgentProvider } from "@aws-sdk/util-user-agent-browser";
 import { DEFAULT_USE_DUALSTACK_ENDPOINT, DEFAULT_USE_FIPS_ENDPOINT } from "@smithy/config-resolver";
 import { FetchHttpHandler as RequestHandler, streamCollector } from "@smithy/fetch-http-handler";
-import { invalidProvider } from "@smithy/invalid-dependency";
 import { loadConfigsForDefaultMode } from "@smithy/smithy-client";
 import { calculateBodyLength } from "@smithy/util-body-length-browser";
 import { resolveDefaultsModeConfig } from "@smithy/util-defaults-mode-browser";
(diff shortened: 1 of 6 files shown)


kuhe:
we do not want to add any unintentional load to this region, despite the ergonomics of the client

kuhe:
So, I guess there are no regional endpoints, but the region selects the partition endpoint.

PR #7774, opened 2026-02-23
chore(packages): use open version range for additional packages

kuhe:
### Issue
internal JS-6432

### Description
This moves additional packages from `packages` to `packages-internal`.

- crc64-nvme: used by internal flexible checksums, not meant for direct dependency like its crt container.
- nested-clients: not meant for direct use, code generated for use in internal credential providers
- signature-v4-multi-region: is a client dependency
- smithy-client: re-exports the smithy package
- util-dynamodb: sometimes used by itself, but not meant to be public
- util-endpoints: internal client dependency, needs code generation infra change

### Testing
CI

### Checklist
  - [x] It's not a feature.
  - [x] No streams here.

---


codegen/smithy-aws-typescript-codegen/src/main/java/software/amazon/smithy/aws/typescript/codegen/DefaultsModeConfigGenerator.java
```diff
@@ -4,210 +4,16 @@
...
-        + "  connectionTimeout?: number;\n"
-        + "  requestTimeout?: number;\n"
-        + "}\n";
-
     private DefaultsModeConfigGenerator() {}
 
-    public static void main(String[] args) throws IOException {
-        // file has been moved to smithy-typescript
-        if (true) {
-            return;
```
kuhe, line 55: this file did nothing

packages-internal/credential-provider-cognito-identity/src/loadCognitoIdentity.ts
```diff
@@ -2,8 +2,8 @@ import {
   CognitoIdentityClient,
   GetCredentialsForIdentityCommand,
   GetIdCommand,
-} from "@aws-sdk/client-cognito-identity";
```
kuhe, line 5: cognito identity and sso are now nested clients as well, so that internal packages do not directly depend on any clients, and can use `nested-clients` with `^` version.

PR #7780, opened 2026-02-24
chore: validate codegen changes using yarn generate-clients

smilkuri:
### Description
Remove the separate `new-client-test-codegen` Smithy model and `client-test-weather` package. Because it failed to catch issues with `yarn generate-clients` as we were doing gradle codegen in isolation. This PR regenerates`weather-legacy-auth` to sync with latest codegen output.

New approach for testing codegen in PR builds:
- Delete `weather-legacy-auth` protocol test client
- Run yarn generate-clients -p
- Restore the package's changelog file
- Restore just the package's version 
- Assert that there are no git diffs

### Testing
CI

### Checklist
  - [x] It's not a feature.
  - [x] No streams here.

---


Makefile
```diff
@@ -30,6 +30,7 @@ test-client-codegen:
 	cp -r ./codegen/new-client-test-codegen/build/smithyprojections/new-client-test-codegen/source/typescript-client-codegen/* ./private/client-test-weather/
 	cat ./private/client-test-weather/package.json
 	yarn update:versions:default
+	git checkout HEAD -- ./private/client-test-weather/CHANGELOG.md ./private/client-test-weather/package.json
```
kuhe: you can't overwrite package.json or else the validations can't run on the generated set of dependencies

kuhe: maybe it would be better not to have this model.

instead, for testing codegen in PR builds:

- delete one of the private clients
- run `yarn generate-clients -p` (protocol test clients).
- restore the package's changelog file
- restore the package's version (but not the rest of the pkg.json file)
- assert that there are no git diffs

smilkuri: How about restoring just the the package version in `./private/client-test-weather/package.json` and assert that  there are no git diffs on this model itself?

smilkuri: There were some codegen changes since the last time these protocol test clients have been updated, example
- copyright year change in the license
- order of build script in package.json
- casing change for esnext in `tsconfig.es.json`
Instead I will do the restoration on `client-test-weather` to keep these consistent with rest of the protocol test clients.

kuhe: it needs to go through the `yarn generate-clients` codegen entry point. 

Having a separate test failed to catch issues with the process that starts with `yarn generate-clients`.

Makefile
```diff
@@ -22,14 +22,18 @@ test-unit: bundles
...
-	mkdir -p ./private/client-test-weather
-	cp -r ./codegen/new-client-test-codegen/build/smithyprojections/new-client-test-codegen/source/typescript-client-codegen/* ./private/client-test-weather/
-	cat ./private/client-test-weather/package.json
-	yarn update:versions:default
+test-codegen:
+	cp ./private/weather-legacy-auth/package.json /tmp/pkg.json.bak
+	cp ./private/weather-legacy-auth/CHANGELOG.md /tmp/changelog.bak
+	rm -rf ./private/weather-legacy-auth
+	yarn generate-clients -p 
+	node -e "const fs=require('fs'); \
```
kuhe: write a script, this is too long for eval

private/weather-legacy-auth/package.json
```diff
@@ -1,12 +1,12 @@
...
   "description": "@aws-sdk/weather-legacy-auth client",
-  "version": "3.997.0",
+  "version": "0.0.1",
   "scripts": {
-    "build": "concurrently 'yarn:build:types' 'yarn:build:es' && yarn build:cjs",
+    "build": "concurrently 'yarn:build:cjs' 'yarn:build:es' 'yarn:build:types'",
     "build:cjs": "tsc -p tsconfig.cjs.json",
     "build:es": "tsc -p tsconfig.es.json",
-    "build:include:deps": "yarn g:turbo run build -F=\"$npm_package_name\"",
+    "build:include:deps": "yarn g:turbo run build -F=\"@aws-sdk/weather-legacy-auth\"",
```
kuhe: this reveals issues with codegen

the code generation process should result in no diffs in pkg json

build should be this:
```
"build": "concurrently 'yarn:build:types' 'yarn:build:es' && yarn build:cjs",
```

and there are two clients which don't have this set.


build include deps should be:
```
"build:include:deps": "yarn g:turbo run build -F=\"$npm_package_name\""
```

Makefile
```diff
@@ -22,14 +22,18 @@ test-unit: bundles
...
-test-client-codegen:
-	rm -rf ./codegen/new-client-test-codegen/build
-	rm -rf ./private/client-test-weather
-	cd codegen && ./gradlew :new-client-test-codegen:build
-	mkdir -p ./private/client-test-weather
-	cp -r ./codegen/new-client-test-codegen/build/smithyprojections/new-client-test-codegen/source/typescript-client-codegen/* ./private/client-test-weather/
-	cat ./private/client-test-weather/package.json
-	yarn update:versions:default
+test-codegen:
+	cp ./private/weather-legacy-auth/package.json /tmp/pkg.json.bak
```
kuhe: this isn't a good client to use since it has legacy auth

use a broadly representative client like rest-json-schema

smilkuri: When we remove a protocol-test client and regenerate it the version would be `1.0.0-alpha.1` and during generation of clients snapshot tests are being executed in the write mode and resulting in update of `x-amz-user-agent` and `user-agent`. Since the placeholder of version `3.___._` doesn't match with `1.0.0-alpha.1`. Therefore using `aws-protocoltests-restjson`

private/aws-protocoltests-restjson/package.json
```diff
@@ -10,10 +10,10 @@
     "build:types": "tsc -p tsconfig.types.json",
     "build:types:downlevel": "downlevel-dts dist-types dist-types/ts3.4",
     "clean": "premove dist-cjs dist-es dist-types tsconfig.cjs.tsbuildinfo tsconfig.es.tsbuildinfo tsconfig.types.tsbuildinfo",
-    "test": "yarn g:vitest run",
+    "test": "yarn dlx vitest run --passWithNoTests",
```
smilkuri: This change in test script in `package.json` due to changes in [TypescriptSettings.java](https://github.com/smithy-lang/smithy-typescript/blob/main/smithy-typescript-codegen/src/main/java/software/amazon/smithy/typescript/codegen/TypeScriptSettings.java#L713)

kuhe: can you fix this in copy-to-clients.js?

```js
if (scripts.test) {
  .replace(/yarn dlx vitest/, 'yarn g:vitest');
}

```

Makefile
```diff
@@ -22,14 +22,14 @@ test-unit: bundles
...
 
-test-client-codegen:
-	rm -rf ./codegen/new-client-test-codegen/build
-	rm -rf ./private/client-test-weather
-	cd codegen && ./gradlew :new-client-test-codegen:build
-	mkdir -p ./private/client-test-weather
-	cp -r ./codegen/new-client-test-codegen/build/smithyprojections/new-client-test-codegen/source/typescript-client-codegen/* ./private/client-test-weather/
-	cat ./private/client-test-weather/package.json
-	yarn update:versions:default
+test-codegen:
```
kuhe, line 25: I don't see this being invoked anywhere.

Makefile
```diff
@@ -22,14 +22,14 @@ test-unit: bundles
...
+	cp ./private/aws-protocoltests-restjson/package.json /tmp/pkg.json.bak
```
kuhe: add `tmp` to ignored but version a `tmp` folder with `.gitkeep` in it to reserve its usage.

This is kind of dangerous since the doc process, I believe, makes a folder called `temp`.

Makefile
```diff
@@ -22,14 +22,14 @@ test-unit: bundles
...
+	cp ./private/aws-protocoltests-restjson/CHANGELOG.md /tmp/changelog.bak
+	rm -rf ./private/aws-protocoltests-restjson
+	yarn generate-clients -p
```
kuhe, line 29: when you rebase, there is a `yarn generate-clients -p` in the `test-integration` target.

Ensure that the codegen process only runs once. I think you can remove the entry in test-integration if you replace it with this target.

private/aws-protocoltests-restjson/package.json
smilkuri: What about `passWithNoTests`?

Makefile
smilkuri: Shouldn't this be invoked in codegeneration step in the CDK?

private/aws-protocoltests-restjson/package.json
kuhe: that can be kept

Makefile
kuhe: you can replace the existing codegen line in test-integration with this.

Since this is a test, it should be in a test phase and not the codegen phase.

scripts/restore-pkg-version.js
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env node
+// Restores the version field from a backup package.json to the current one.
+
+const fs = require("fs");
```
kuhe:
```suggestion
const fs = require("node:fs");
```

Makefile
kuhe: change the smithy-build.json config of this client to have a 3.0.0-alpha.1 version then

PR #7786, opened 2026-02-25
feat(util-user-agent-node): populate typescript version in user agent when available

trivikr:
### Issue
Internal JS-6661

### Description
Populates typescript version in user agent when available in `node_modules/typescript/package.json`

### Testing

This change is able to detect TypeScript version is CJS and ESM applications, as they both use CJS dependency
Details in https://github.com/trivikr/test-aws-sdk-js-typescript-detection

However, it's not able to detect TypeScript version in ESM bundles which don't populate `__dirname`, i.e. esbuild, rollup and rolldown. Details in https://github.com/trivikr/test-aws-sdk-js-typescript-detection/pull/5

It's able to detect TypeScript version in CJS bundles of popular bundlers, and ESM bundles of bundlers which populate `__dirname`, like webpack and rspack.

### Checklist
  - [ ] It's not a feature.
  - [x] Integration tests are written in https://github.com/trivikr/test-aws-sdk-js-typescript-detection, as they work from inside node_modules
  - [x] No streams here.

---

### DIFF
--- a/packages-internal/util-user-agent-node/src/getTypeScriptUserAgentPair.ts
+++ b/packages-internal/util-user-agent-node/src/getTypeScriptUserAgentPair.ts
@@ -19,7 +19,12 @@ export const getTypeScriptUserAgentPair = async (): Promise<UserAgentPair | unde
 
   try {
     const packageJson = await readFile(getTypeScriptPackageJsonPath(__dirname), "utf-8");
-    tscVersion = JSON.parse(packageJson).version as string;
+    const { version } = JSON.parse(packageJson);
+    if (typeof version !== "string") {
+      tscVersion = null;
+      return undefined;
+    }
+    tscVersion = version;
     return ["md/tsc", tscVersion];
   } catch {
     // Ignore error in case of failure in file read or JSON parsing.
--- a/packages-internal/util-user-agent-node/src/getTypeScriptUserAgentPair.spec.ts
+++ b/packages-internal/util-user-agent-node/src/getTypeScriptUserAgentPair.spec.ts
@@ -48,6 +48,16 @@ describe("getTypeScriptUserAgentPair", () => {
 
       expect(readFile).toHaveBeenCalledTimes(1);
     });
+
+    it("returns cached version on subsequent calls if version is not defined", async () => {
+      vi.mocked(readFile).mockResolvedValue(JSON.stringify({ name: "blah" }));
+
+      const { getTypeScriptUserAgentPair } = await import("./getTypeScriptUserAgentPair");
+      await expect(getTypeScriptUserAgentPair()).resolves.toBeUndefined();
+      await expect(getTypeScriptUserAgentPair()).resolves.toBeUndefined();
+
+      expect(readFile).toHaveBeenCalledTimes(1);
+    });
   });
 
   describe("when reading typescript/package.json throws error", () => {


trivikr:
Verified that bundling with rollup is successful, and it's able to detect typescript version

```js
// input.js
import { createDefaultUserAgentProvider } from "@aws-sdk/util-user-agent-node";

const userAgentProvider = createDefaultUserAgentProvider({
  serviceId: "s3",
  clientVersion: "0.1.0",
});
userAgentProvider().then(console.log);
```

```console
$ npx rollup input.js --file bundle.js --format esm

$ node bundle.js                         
[
  [ 'aws-sdk-js', '0.1.0' ],
  [ 'ua', '2.1' ],
  [ 'os/linux', '5.10.248-248.988.amzn2int.x86_64' ],
  [ 'lang/js' ],
  [ 'md/nodejs', '24.11.1' ],
  [ 'md/tsc', '5.9.3' ],
  [ 'api/s3', '0.1.0' ]
]
```

packages-internal/util-user-agent-node/src/getTypeScriptUserAgentPair.ts
```diff
@@ -0,0 +1,28 @@
+import type { UserAgentPair } from "@smithy/types";
+import { readFile } from "node:fs/promises";
+
+import { getTypeScriptPackageJsonPath } from "./getTypeScriptPackageJsonPath";
+
+let tscVersion: string | null;
```
kuhe:
```suggestion
let tscVersion: string | null | undefined;
```

packages-internal/util-user-agent-node/src/getTypeScriptPackageJsonPath.ts
```diff
@@ -0,0 +1,24 @@
+import { join, normalize, sep } from "node:path";
+
+export const getTypeScriptPackageJsonPath = (dirname = ""): string => {
```
kuhe, line 11:
```suggestion
/**
 * @internal
 */
export const getTypeScriptPackageJsonPath = (dirname = ""): string => {
```

trivikr: Added in https://github.com/aws/aws-sdk-js-v3/pull/7786/changes/dfdf947d7aa5f28765eebe23a44e90b0ae5d5f80

packages-internal/util-user-agent-node/src/getTypeScriptUserAgentPair.ts
trivikr: Not needed as tscVersion is either assigned a null or string explicitly.

kuhe: it's initialized to undefined though, this could be 

```js
let tscVersion: string | null = "";
```

packages-internal/util-user-agent-node/src/getTypeScriptUserAgentPair.ts
```diff
@@ -0,0 +1,28 @@
...
+export const getTypeScriptUserAgentPair = async (): Promise<UserAgentPair | undefined> => {
+  // If tscVersion is set from previous calls.
+  if (tscVersion === null) {
+    return undefined;
+  } else if (tscVersion) {
+    return ["md/tsc", tscVersion];
+  }
+
+  try {
+    const packageJson = await readFile(getTypeScriptPackageJsonPath(__dirname), "utf-8");
```
kuhe, line 21: extremely unlikely edge case: file is extremely large and causes readFile utf8 to run out of memory

I don't think you need to handle this case, since it's too unlikely

packages-internal/util-user-agent-node/src/getTypeScriptUserAgentPair.ts
trivikr: Added `undefined` type in https://github.com/aws/aws-sdk-js-v3/pull/7786/changes/9d670c16d6f4c8582828a4de03add2fed916fa23, since it's automatically set to undefined during initialization.

trivikr:
The fix for the edge cases was tested in https://github.com/trivikr/test-aws-sdk-js-typescript-detection/pull/8

PR #7800, opened 2026-02-28
fix(core/protocols): preserve nulls in non-sparse collections

siddsriv:
### Issue
Internal JS-6670

### Description
currently, the json deserializer was silently skipping over null values in non-sparse collections. this change will now preserve the null value for deserialization for all protocols where we were checking for sparseness. 

### Testing
updated package tests
```console
 ✓ src/submodules/protocols/json/JsonShapeSerializer.spec.ts (6 tests) 222ms
.
.
.
 ✓ src/submodules/protocols/xml/XmlShapeDeserializer.spec.ts (5 tests) 281ms

 Test Files  42 passed (42)
      Tests  173 passed | 1 skipped (174)
```

### Checklist
  - [x] It's not a feature.
  - [x] No streams here.

---

### DIFF
--- a/packages-internal/core/src/submodules/protocols/xml/XmlShapeDeserializer.ts
+++ b/packages-internal/core/src/submodules/protocols/xml/XmlShapeDeserializer.ts
@@ -84,6 +84,11 @@ export class XmlShapeDeserializer extends SerdeContextConfig implements ShapeDes
 
         const sourceKey = listValue.getMergedTraits().xmlName ?? "member";
         const source = flat ? value : (value[0] ?? value)[sourceKey];
+
+        if (source == null) {
+          return buffer;
+        }
+
         const sourceArray = Array.isArray(source) ? source : [source];
 
         for (const v of sourceArray) {
--- a/packages-internal/core/src/submodules/protocols/xml/XmlShapeDeserializer.spec.ts
+++ b/packages-internal/core/src/submodules/protocols/xml/XmlShapeDeserializer.spec.ts
@@ -89,6 +89,31 @@ describe(XmlShapeDeserializer.name, () => {
     });
   });
 
+  it("deserializes empty XML lists correctly", async () => {
+    const emptyListSchema = [
+      3,
+      "",
+      "Widget",
+      0,
+      ["stringList", "stringSet"],
+      [
+        [() => [1, "", "StringList", {}, 0], {}],
+        [() => [1, "", "StringList", {}, 0], {}],
+      ],
+    ];
+
+    const xml = `<Widget xmlns="namespace">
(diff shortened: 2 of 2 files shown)


packages-internal/core/src/submodules/protocols/json/JsonShapeDeserializer.ts
```diff
@@ -93,22 +93,16 @@ export class JsonShapeDeserializer extends SerdeContextConfig implements ShapeDe
       if (Array.isArray(value) && ns.isListSchema()) {
         const listMember = ns.getValueSchema();
         const out = [] as any[];
-        const sparse = !!ns.getMergedTraits().sparse;
```
kuhe, line 96: 1. use the submodule name in the PR and commit message, like `fix(core/protocols): ...`
2. if this change is made, it must apply to all protocols. There are shape deserializers for XML, JSON, and CBOR. There is also a FromStringShapeDeserializer but it does not read the sparse trait.

kuhe: https://github.com/smithy-lang/smithy-typescript/blob/main/packages/core/src/submodules/cbor/CborCodec.ts#L137

https://github.com/aws/aws-sdk-js-v3/blob/main/packages-internal/core/src/submodules/protocols/json/JsonShapeDeserializer.ts (in PR already)

https://github.com/aws/aws-sdk-js-v3/blob/main/packages-internal/core/src/submodules/protocols/xml/XmlShapeDeserializer.ts

packages-internal/core/src/submodules/protocols/json/JsonShapeDeserializer.spec.ts
kuhe:
```suggestion
  it("ignores list sparseness in the deserialization path", async () => {
```

kuhe:
```suggestion
  it("ignores map sparseness in the deserialization path, taking server output as-is", async () => {
```

packages-internal/core/src/submodules/protocols/xml/XmlShapeDeserializer.ts
```diff
@@ -76,7 +76,6 @@ export class XmlShapeDeserializer extends SerdeContextConfig implements ShapeDes
     }
 
     if (typeof value === "object") {
-      const sparse = !!traits.sparse;
```
kuhe, line 79: please create a unit test for this for XML

kuhe:
retest after https://github.com/aws/aws-sdk-js-v3/pull/7806

PR #7808, opened 2026-03-03
test(clients): run client feature tests in browser mode too

kuhe:
### Issue
JS-6630

### Description
This adds a test mode called browser-cross-platform, which runs shared e2e tests from both the browser and Node.js. 

Initially, the set of tests added to this mode is most of the e2e tests within the `/clients` folders.

### Testing
CI

### Checklist
  - [ ] It's not a feature.
  - [x] No streams here.

---

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -32,7 +32,7 @@ test-codegen:
 	git diff --exit-code ./
 
 # typecheck for test code.
-test-types:
+test-types: reset-test-credentials
 	npx tsc -p tsconfig.test.json
 
 test-indices:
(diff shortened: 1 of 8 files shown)


clients/client-acm/test/acm-features.e2e.spec.ts
```diff
@@ -5,7 +5,7 @@ describe(ACM.name, () => {
   let client: ACM;
 
   beforeAll(async () => {
-    client = new ACM({ region: "us-west-2" });
+    client = new ACM({ region: "us-west-2", credentials: aws?.testCredentials });
```
kuhe, line 8: The Node.js / browser test-sharing mechanism is a set of credentials written to a global field `aws?.testCredentials`. 

Using a test setup script that varies between the two environments:
- browsers borrow Node.js credentials written to a static (unversioned) file.
- Node.js execution leaves this value undefined so the default provider chain fires.

PR #7811, opened 2026-03-04
fix(middleware-user-agent): use retry strategy mode property instead of legacy fallback

### LINKED ISSUE #7574: feature attribution for retry mode should not include "legacy"

### Describe the bug

in user agent `check-features.ts` file, the detection of pre-SRA retries should not attribute the retry as "legacy" mode. It should differentiate and set the feature as either (pre-SRA) standard or (pre-SRA) adaptive. 

Legacy refers to the retry behavior in AWS SDK JSv2, which does not exist in this SDK.

### Regression Issue

### SDK version number

https://github.com/aws/aws-sdk-js-v3/releases/tag/v3.669.0

Node.js

any

### Reproduction Steps

make a request with pre-SRA retries and inspect the user agent feature metadata

### Observed Behavior

it contains D for legacy

### Expected Behavior

it should contain either F or E for adaptive/standard.

### Possible Solution

_No response_

_No response_

TrevorBurnham:
Fixes #7574

### Problem

The `checkFeatures` function in `middleware-user-agent` incorrectly attributes pre-SRA retry strategies as "legacy" (`D`) in the user-agent feature metadata. It does this by checking for `acquireInitialRetryToken` (a `RetryStrategyV2`-only method), and if absent, assumes the retry mode is legacy.

"Legacy" refers to AWS SDK for JavaScript v2 retry behavior, which does not exist in v3. Pre-SRA retry strategies in this SDK are still either standard or adaptive — they just implement the older `RetryStrategy` interface instead of `RetryStrategyV2`.

### Solution

All retry strategies in v3 (both SRA and pre-SRA) expose a `mode` property set to either `"standard"` or `"adaptive"`. The fix replaces the `acquireInitialRetryToken` type-check branching with a single `mode` property check, which correctly identifies the retry mode regardless of which interface the strategy implements.

Before:
```ts
if (typeof (retryStrategy as RetryStrategyV2).acquireInitialRetryToken === "function") {
  if (retryStrategy.constructor?.name?.includes("Adaptive")) {
    setFeature(context, "RETRY_MODE_ADAPTIVE", "F");
  } else {
    setFeature(context, "RETRY_MODE_STANDARD", "E");
  }
} else {
  setFeature(context, "RETRY_MODE_LEGACY", "D"); // wrong for v3 pre-SRA strategies
}
```

After:
```ts
if (typeof retryStrategy.mode === "string" && retryStrategy.mode.includes("adaptive")) {
  setFeature(context, "RETRY_MODE_ADAPTIVE", "F");
} else {
  setFeature(context, "RETRY_MODE_STANDARD", "E");
}
```

### Checklist
  - [x] It's not a feature.
  - [x] No streams here.

---

### DIFF
--- a/packages-internal/middleware-user-agent/package.json
+++ b/packages-internal/middleware-user-agent/package.json
@@ -31,6 +31,7 @@
     "@smithy/core": "^3.23.8",
     "@smithy/protocol-http": "^5.3.11",
     "@smithy/types": "^4.13.0",
+    "@smithy/util-retry": "^4.2.11",
     "tslib": "^2.6.2"
   },
   "devDependencies": {
--- a/packages-internal/middleware-user-agent/src/check-features.ts
+++ b/packages-internal/middleware-user-agent/src/check-features.ts
@@ -6,13 +6,8 @@ import type {
   AwsSdkCredentialsFeatures,
 } from "@aws-sdk/types";
 import type { IHttpRequest } from "@smithy/protocol-http";
-import type {
-  AwsCredentialIdentityProvider,
-  BuildHandlerArguments,
-  Provider,
-  RetryStrategy,
-  RetryStrategyV2,
-} from "@smithy/types";
+import type { AwsCredentialIdentityProvider, BuildHandlerArguments, Provider } from "@smithy/types";
+import { RETRY_MODES } from "@smithy/util-retry";
 
 /**
  * @internal
@@ -20,7 +15,7 @@ import type {
 type PreviouslyResolved = Partial<{
   credentials?: AwsCredentialIdentityProvider;
   accountIdEndpointMode?: Provider<AccountIdEndpointMode>;
-  retryStrategy?: Provider<RetryStrategy | RetryStrategyV2>;
+  retryStrategy?: Provider<{ mode?: string }>;
 }>;
 
 /**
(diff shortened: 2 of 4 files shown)


packages-internal/middleware-user-agent/src/check-features.ts
```diff
@@ -46,14 +40,10 @@ export async function checkFeatures(
...
-        setFeature(context, "RETRY_MODE_ADAPTIVE", "F");
-      } else {
-        setFeature(context, "RETRY_MODE_STANDARD", "E");
-      }
+    if (typeof retryStrategy.mode === "string" && retryStrategy.mode.includes("adaptive")) {
+      setFeature(context, "RETRY_MODE_ADAPTIVE", "F");
     } else {
-      setFeature(context, "RETRY_MODE_LEGACY", "D");
+      setFeature(context, "RETRY_MODE_STANDARD", "E");
     }
```
trivikr: We should do exact match for values from util-retry, and skip logging otherwise.
```suggestion
    if (typeof retryStrategy.mode === "string") {
      switch (retryStrategy.mode) {
        case RETRY_MODES.ADAPTIVE:
          setFeature(context, "RETRY_MODE_ADAPTIVE", "F");
          break;
        case RETRY_MODES.STANDARD:
          setFeature(context, "RETRY_MODE_STANDARD", "E");
          break;
      }
    }
```

where
```js
import { RETRY_MODES } from "@smithy/util-retry";
```

TrevorBurnham: Updated!

packages-internal/middleware-user-agent/package.json
```diff
@@ -31,6 +31,7 @@
     "@smithy/core": "^3.23.8",
     "@smithy/protocol-http": "^5.3.11",
     "@smithy/types": "^4.13.0",
+    "@smithy/util-retry": "^4.2.10",
```
trivikr: The version was updated this morning.
```suggestion
    "@smithy/util-retry": "^4.2.11",
```

Can you rebase from main and also run `yarn` to update lockfile?
And push with `--force-with-lease`?

TrevorBurnham: Done!

PR #7832, opened 2026-03-09
fix(util-endpoints): cache compiled RegExp in partition lookup

TrevorBurnham:
## Problem

The `partition()` function in `@aws-sdk/util-endpoints` is called during endpoint resolution on every API request. On each call, it constructs a new `RegExp` object from the `regionRegex` string for every partition entry:

```ts
if (new RegExp(regionRegex).test(value)) {
```

Since the partition data (and therefore the regex strings) are static between `setPartitionInfo()` calls, this repeated compilation is unnecessary work on the hot path.

## Solution

Introduce a simple `Record<string, RegExp>` cache that stores compiled `RegExp` objects keyed by their source string. A helper function `getRegionRegex()` returns the cached regex or compiles and caches it on first encounter.

```ts
const regionRegexCache: Record<string, RegExp> = {};

const getRegionRegex = (regionRegex: string): RegExp => {
  if (!regionRegexCache[regionRegex]) {
    regionRegexCache[regionRegex] = new RegExp(regionRegex);
  }
  return regionRegexCache[regionRegex];
};
```

The cache is keyed by the regex string itself, so it naturally handles `setPartitionInfo()` swapping in different partition data. The cache is cleared inside `setPartitionInfo()` so stale entries from a previous partition configuration don't accumulate as dead weight.

---

### DIFF
--- a/packages-internal/util-endpoints/src/lib/aws/partition.ts
+++ b/packages-internal/util-endpoints/src/lib/aws/partition.ts
@@ -35,6 +35,19 @@ let selectedPartitionsInfo: PartitionsInfo = partitionsInfo;
 let selectedUserAgentPrefix = "";
 
 /**
+ * Compiled regex cache for regionRegex strings.
+ * Avoids re-creating RegExp objects on every partition() call.
+ */
+const regionRegexCache: Record<string, RegExp> = {};
+
+const getRegionRegex = (regionRegex: string): RegExp => {
+  if (!regionRegexCache[regionRegex]) {
+    regionRegexCache[regionRegex] = new RegExp(regionRegex);
+  }
+  return regionRegexCache[regionRegex];
+};
+
+/**
  * Evaluates a single string argument value as a region, and matches the
  * string value to an AWS partition.
  * The matcher MUST always return a successful object describing the partition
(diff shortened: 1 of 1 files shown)


packages-internal/util-endpoints/src/lib/aws/partition.ts
```diff
@@ -34,6 +34,19 @@ let selectedPartitionsInfo: PartitionsInfo = partitionsInfo;
...
+ * Avoids re-creating RegExp objects on every partition() call.
+ */
+const regionRegexCache: Record<string, RegExp> = {};
+
+const getRegionRegex = (regionRegex: string): RegExp => {
+  if (!regionRegexCache[regionRegex]) {
+    regionRegexCache[regionRegex] = new RegExp(regionRegex);
+  }
+  return regionRegexCache[regionRegex];
+};
```
kuhe, line 48: isn't this covered by the endpoint cache at the client level? why do we want a cache within a cache?

it may be too much complexity for no gain

TrevorBurnham: You're right. There's very little juice to be squeezed here.

PR #7834, opened 2026-03-09
chore(endpoint-cache): reduce redundant LRU cache lookups in EndpointCache.get()

TrevorBurnham:
## Problem

`EndpointCache.get()` performs 3 underlying LRU cache accesses for a single lookup:

1. `this.has(key)` → calls `this.cache.has(key)` (hash lookup)
2. Inside `has()` → calls `this.cache.peek(key)` (hash lookup + array read, no LRU promotion)
3. Back in `get()` → calls `this.cache.get(key)` (hash lookup + array read + splay/LRU promotion)

The `has()` check exists to handle soft-deleted entries (mnemonist's LRUCache lacks a `delete` method, so deletion is emulated by setting the value to `[]`). But `get()` can handle this directly.

## Solution

Replace the `has()` guard + separate `get()` with a single `this.cache.get(key)` call, followed by a `!value || value.length === 0` check that covers both the missing-key and soft-deleted cases.

Before:
```ts
get(key: string) {
  if (!this.has(key)) {   // cache.has() + cache.peek()
    return;
  }
  const value = this.cache.get(key);  // 3rd lookup
  if (!value) {
    return;
  }
  // ... filter expired endpoints
}
```

After:
```ts
get(key: string) {
  const value = this.cache.get(key);  // single lookup
  if (!value || value.length === 0) {
    return;
  }
  // ... filter expired endpoints
}
```

## Behavioral notes

- `has()` is unchanged and still available for external callers.
- LRU recency change: the old code used `peek()` (via `has()`) to detect soft-deleted entries without promoting them in the LRU. The new code calls `cache.get()` unconditionally, which does promote tombstones (empty arrays from `delete()`). This means looked-up tombstones will be kept alive slightly longer and could displace real entries in a full cache. In practice this is negligible — tombstones are overwritten by future `set()` calls, and the cache capacity would need to be very tight for displacement to matter.
- Tests were updated to reflect that `get()` no longer delegates to `has()`/`peek()`.

### Checklist
  - [x] It's not a feature.
  - [x] No streams here.

---

### DIFF
--- a/packages-internal/endpoint-cache/src/EndpointCache.spec.ts
+++ b/packages-internal/endpoint-cache/src/EndpointCache.spec.ts
@@ -54,44 +54,27 @@ describe(EndpointCache.name, () => {
 
   describe("get", () => {
     beforeEach(() => {
-      has.mockReturnValue(true);
       const endpointsWithExpiry = getEndpointsWithExpiry(mockEndpoints);
-      peek.mockReturnValue(endpointsWithExpiry);
       get.mockReturnValue(endpointsWithExpiry);
       vi.spyOn(Date, "now").mockImplementation(() => now);
     });
 
-    const verifyHasAndGetCalls = () => {
-      expect(has).toHaveBeenCalledTimes(1);
-      expect(has).toHaveBeenCalledWith(key);
+    const verifyGetCalls = () => {
       expect(get).toHaveBeenCalledTimes(1);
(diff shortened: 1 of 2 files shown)


kuhe:
this is a confusing naming collision, but this is not the EndpointCache that runs in the code path of every client

TrevorBurnham:
Right you are. 👍

PR #7837, opened 2026-03-10
fix(core/protocols): restore error fallback to UnknownError for protocols

siddsriv:
### Issue
Internal JS-6638

### Description
restore error fallback for all protocols to "UnknownError" instead of "Unknown"
related fix: https://github.com/aws/aws-sdk-js-v3/pull/7717

### Testing
- local script to get error response and `core` package tests pass
- snapshot tests (diffs attached)

### Checklist
  - [x] It's not a feature.
  - [x] No streams here.

---

### DIFF
--- a/packages-internal/core/src/submodules/protocols/xml/AwsRestXmlProtocol.ts
+++ b/packages-internal/core/src/submodules/protocols/xml/AwsRestXmlProtocol.ts
@@ -132,7 +132,11 @@ export class AwsRestXmlProtocol extends HttpBindingProtocol {
 
     const ns = NormalizedSchema.of(errorSchema);
     const message =
-      dataObject.Error?.message ?? dataObject.Error?.Message ?? dataObject.message ?? dataObject.Message ?? "Unknown";
+      dataObject.Error?.message ??
+      dataObject.Error?.Message ??
+      dataObject.message ??
+      dataObject.Message ??
+      "UnknownError";
     const ErrorCtor = TypeRegistry.for(errorSchema[1]).getErrorCtor(errorSchema) ?? Error;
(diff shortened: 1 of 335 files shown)


packages-internal/core/src/submodules/protocols/xml/AwsRestXmlProtocol.ts
```diff
@@ -132,7 +132,11 @@ export class AwsRestXmlProtocol extends HttpBindingProtocol {
 
     const ns = NormalizedSchema.of(errorSchema);
     const message =
-      dataObject.Error?.message ?? dataObject.Error?.Message ?? dataObject.message ?? dataObject.Message ?? "Unknown";
+      dataObject.Error?.message ??
+      dataObject.Error?.Message ??
+      dataObject.message ??
+      dataObject.Message ??
+      "UnknownError";
```
kuhe, line 139: are there differences in the default error message for other protocols that were introduced in Nov '25?

siddsriv: yes, should we change all `Unknown`s to `UnknownError`s?

siddsriv: i restored uniformity for this message for all protocols (`smithy-client` [src here, what i think was the old behavior](https://github.com/smithy-lang/smithy-typescript/blob/6a0fd69fff1fe145d3ae7b1d4911f34e01b9c467/packages/smithy-client/src/default-error-handler.ts#L15)) 

separate commits for each protocol, for ease of review

kuhe:
this was done in another PR

PR #7839, opened 2026-03-11
docs(clients): generate readme block about protocols

kuhe:
### Issue
JS-6426

### Description
This generates a block in each client readme if the client supports more than one protocol.

The readme template has also been edited for minor issues.

### Testing
visual check

### Checklist
  - [x] It's not a feature.
  - [x] No streams here.

---

### DIFF
--- a/clients/client-cloudwatch/README.md
+++ b/clients/client-cloudwatch/README.md
@@ -21,8 +21,9 @@ visibility into resource utilization, application performance, and operational
 health.</p>
 
 ## Installing
-To install this package, simply type add or install @aws-sdk/client-cloudwatch
-using your favorite package manager:
+
+To install this package, use the CLI of your favorite package manager:
+
 - `npm install @aws-sdk/client-cloudwatch`
 - `yarn add @aws-sdk/client-cloudwatch`
 - `pnpm add @aws-sdk/client-cloudwatch`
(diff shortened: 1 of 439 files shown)


clients/client-cloudwatch/README.md
```diff
@@ -47,24 +48,74 @@ import { CloudWatchClient, ListMetricsCommand } from "@aws-sdk/client-cloudwatch
 
 ### Usage
 
-To send a request, you:
+To send a request:
 
-- Initiate client with configuration (e.g. credentials, region).
-- Initiate command with input parameters.
-- Call `send` operation on client with command object as input.
-- If you are using a custom http handler, you may call `destroy()` to close open connections.
```
kuhe, line 55: removed destroy call because it's unclear to customers when would need this.

for beginners and basic usage, the answer is "never".

clients/client-cloudwatch/README.md
```diff
@@ -47,24 +48,74 @@ import { CloudWatchClient, ListMetricsCommand } from "@aws-sdk/client-cloudwatch
...
 
 ```js
-// a client can be shared by different commands.
 const client = new CloudWatchClient({ region: "REGION" });
 
 const params = { /** input parameters */ };
 const command = new ListMetricsCommand(params);
 ```
 
+#### Supported Message Protocols
```
kuhe, line 66: This block appears only for clients supporting multiple protocols.

For clients with one protocol, no mention is made of this option.

clients/client-cloudwatch/README.md
```diff
@@ -79,26 +130,9 @@ try {
...
-  (data) => {
-    // process data.
-  },
-  (error) => {
-    // error handling.
-  }
-);
-```
-
-Promises can also be called using `.catch()` and `.finally()` as follows:
```
kuhe, line 101: I removed the `.then((res, err) => {})` syntax because it is redundant with the `.then(res => {}).catch(err => {})` example below.

We don't need to go into this much detail about how to use Promises anymore.

clients/client-cloudwatch/README.md
```diff
@@ -114,27 +148,17 @@ client
...
-
-We do not recommend using callbacks because of [callback hell](http://callbackhell.com/),
-but they are supported by the send operation.
-
-```js
-// callbacks.
-client.send(command, (err, data) => {
-  // process err and data.
-});
-```
```
kuhe, line 127: I removed the callback block because, if we don't recommend it, we don't need to highlight it.

clients/client-cloudwatch/README.md
```diff
@@ -114,27 +148,17 @@ client
...
+#### Aggregate client
 
-#### v2 compatible style
+The aggregate version of the client is exported from the same package, but without the "-Client" suffix.
 
-The client can also send requests using v2 compatible style.
-However, it results in a bigger bundle size and may be dropped in next major version. More details in the blog post
-on [modular packages in AWS SDK for JavaScript](https://aws.amazon.com/blogs/developer/modular-packages-in-aws-sdk-for-javascript/)
```
kuhe, line 133: I changed this to say that the aggregate client is available and resembles v2. 

I removed the remark about it being dropped in next major version. Even if we did create a new major version, dynamic creation of the aggregate client would almost certainly be possible and included for compatibility/ease-of-use.

clients/client-cloudwatch/README.md
```diff
@@ -154,7 +178,7 @@ client
     // error handling.
   });
 
-// callbacks.
+// callbacks (not recommended).
 client.listMetrics(params, (err, data) => {
   // process err and data.
 });
```
kuhe, line 188: I leave this callback example, because v2 migration users would be more likely to want to use it. It was removed above in the Command syntax block.

clients/client-cloudwatch/README.md
```diff
@@ -114,27 +148,17 @@ client
...
-We do not recommend using callbacks because of [callback hell](http://callbackhell.com/),
-but they are supported by the send operation.
-
-```js
-// callbacks.
-client.send(command, (err, data) => {
-  // process err and data.
-});
-```
+#### Aggregate client
```
trivikr:
```suggestion
#### Aggregated client
```

clients/client-cloudwatch/README.md
```diff
@@ -114,27 +148,17 @@ client
...
 
-#### v2 compatible style
+The aggregate version of the client is exported from the same package, but without the "-Client" suffix.
```
trivikr:
```suggestion
The aggregated client is exported from the same package, but without the "Client" suffix.
```

clients/client-cloudwatch/README.md
```diff
@@ -114,27 +148,17 @@ client
...
 
-The client can also send requests using v2 compatible style.
-However, it results in a bigger bundle size and may be dropped in next major version. More details in the blog post
-on [modular packages in AWS SDK for JavaScript](https://aws.amazon.com/blogs/developer/modular-packages-in-aws-sdk-for-javascript/)
+This version of the client supports all operations, waiters, and paginators as methods.
+This style may be familiar to you from the AWS SDK for JavaScript v2 previous major version.
```
trivikr:
```suggestion
This style may be familiar to you from the AWS SDK for JavaScript v2.
```

clients/client-cloudwatch/README.md
trivikr: Decided to retain this para, except the "and may be dropped in next major version" part.

PR #7848, opened 2026-03-13
test(canary): use vitest 4.0.17 in canary tests

trivikr:
### Issue
Internal JS-6701

### Description
Use `4.0.x` in canary tests

### Testing
CI

### Checklist
  - [x] It's not a feature.
  - [x] No streams here.

---

### DIFF
--- a/tests/canary/canary-runner.js
+++ b/tests/canary/canary-runner.js
@@ -49,7 +49,7 @@ const packageJson = {
     ...Object.fromEntries([...dependencies].sort().map((dep) => [dep, "latest"])),
     "@aws-sdk/config": "latest",
     "happy-dom": "20.8.3",
-    vitest: "~4.0.17",
+    vitest: "4.0.17",
   },
 };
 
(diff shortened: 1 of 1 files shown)


kuhe (approved):

trivikr:
No longer required, as internal issue JS-6701 is resolved with a retry.
A backlog is created in JS-6703 for investigation though.

tests/canary/canary-runner.js
```diff
@@ -49,7 +49,7 @@ const packageJson = {
     ...Object.fromEntries([...dependencies].sort().map((dep) => [dep, "latest"])),
     "@aws-sdk/config": "latest",
     "happy-dom": "20.8.3",
-    vitest: "^4.0.17",
+    vitest: "~4.0.17",
```
kuhe:
```suggestion
    vitest: "4.0.17",
```

since there is no lockfile here, I think we should pin the version

tests/canary/canary-runner.js
kuhe, line 46:
```suggestion
    canary: "npx vitest run --retry=4 --test-timeout=60000",
```

trivikr: This may not matter since we run canary tests using `npm run canary` which means npm CLI finds where vitest is. Adding `npx` is redundant, since it'll just tell npm CLI to find where npx is, which in turn will find where vitest is.

Issue #7830, opened 2026-03-06
lib-dynamodb: Expose Transaction Put / Update / Delete types.


dblackwell-speedline:
### Describe the feature

In `lib/lib-dynamodb/src/commands/TransactWriteCommand.ts`, you expose the type `TransactWriteCommandInput`, which accepts transaction items in the form of slightly-altered `Put`, `Update`, and `Delete` items. However, these items do not have their types exposed. We would like them to be.

### Use Case

We have written a TransactionBuilder class that allows for readable shorthand for putting, updating, and deleting items in a transaction. However, we've had to create the types above, which does not look clean.

### Proposed Solution

Split the `Put`, `Update`, and `Delete` children into their own exposed types, and reference these types in the already-exposed `TransactWriteCommandInput`.

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### SDK version used

3.1003.0

### Environment details (OS name and version, etc.)

Node 24, AWS Lambda (or locally on Win 10)

obones:
This would indeed be very convenient.

PR #7852, opened 2026-03-17
feat(lib-dynamodb): expose transact write item types

lucianlature-endava:
### Issue
#7830 

### Description
Exposes document-client types for TransactWriteItems so consumers can type transaction items instead of using inline shapes.

Changes:
- Codegen: Updated DocumentClientCommandGenerator.java to handle the TransactWriteItems operation.
- New types: TransactWriteConditionCheck, TransactWritePut, TransactWriteDelete, TransactWriteUpdate, and TransactWriteItem.
- Input type: TransactWriteCommandInput now uses TransactWriteItem[] for TransactItems instead of an inline type.
- These types use NativeAttributeValue instead of AttributeValue, matching the document client’s native JS API.

### Testing
1. Type-check: `cd lib/lib-dynamodb && yarn build:types`
2. Unit tests: `cd lib/lib-dynamodb && yarn test`

### Checklist
- [x] It's not a feature.
- [x] No streams here.

---


lib/lib-dynamodb/src/commands/BatchExecuteStatementCommand.ts
```diff
@@ -15,50 +20,37 @@ export { DynamoDBDocumentClientCommand, $Command };
...
-            NativeAttributeValue
-          >
-           | undefined
-        ;
-      }
-    )[]
-     | undefined
-  ;
+export type BatchExecuteStatementCommandInput = Omit<__BatchExecuteStatementCommandInput, "Statements"> & {
+  Statements: BatchExecuteStatementBatchStatementRequest[] | undefined;
```
kuhe, line 46: why not just use the type `BatchExecuteStatementCommandInput["Statements"]`?

lucianlature-endava: That would hit a self-reference: `BatchExecuteStatementCommandInput["Statements"]` will expand to "the type of the Statements property on BatchExecuteStatementCommandInput". But `BatchExecuteStatementCommandInput` itself is currently being defined, and its Statements property’s type is exactly what we’re trying to specify, so we'll end up with a circular definition, and TypeScript generally can't resolve that as a meaningful alias.

A non-circular DRY version would look like:
```typescript
export type BatchExecuteStatementStatements = BatchExecuteStatementBatchStatementRequest[] | undefined;
export type BatchExecuteStatementCommandInput = Omit<__BatchExecuteStatementCommandInput, "Statements"> & {
  Statements: BatchExecuteStatementStatements;
```

PR #7864, opened 2026-03-18
test(performance-tests): create performance test package

siddsriv:
### Issue
Internal JS-6417

### Description
Prepare a new performance tests package
- First metric: we start with refactoring the existing bundle size measurement infra in the SDK, to be used in the new `performance-tests` package for json outputs to stdout.
- Markdown reports should still be generated in this monorepo.

### Testing
`test.sh` serves as the entry point for running performance tests, and will output results to stdout.

### Checklist
  - [x] It's not a feature.
  - [x] No streams here.

---


tests/performance-tests/test.sh
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env bash
+set -euo pipefail
```
siddsriv, line 2: https://gist.github.com/mohanpedala/1e2ff5661761d3abd0385e8223e16425?permalink_comment_id=3799230

package.json
siddsriv, line 7: ordered alphabetically, can isolate this for a separate PR if you'd like

trivikr (approved):

tests/performance-tests/runner/BundleSizeReportGenerator.mjs
kuhe:
```suggestion
const REPORT_PATH = path.join(__dirname, "..", "..", "..", "benchmark", "size", "report-bundlers.md");
```

kuhe: wrap `{}` for control flow blocks for readability.

These numbers aren't going to align well because 0.123 and 0.000 aren't the same width when trailing zeroes are truncated. Is this for a table where that matters?

siddsriv: it's to keep parity with the table we publish currently, since I removed the tests/bundlers dir, this is just to create and publish the same report in the public repo

tests/performance-tests/runner/prepare-artifact.mjs
kuhe: what is this doing? isn't recursive copying natively supported by fs?

kuhe: RUNTIME_SCOPED_PACKAGES

kuhe: no types needed?

kuhe: console.info?

tests/performance-tests/runner/run.mjs
kuhe: use path join without separators

kuhe: console.warn error? or too noisy?

kuhe, line 68: console.log or info

tests/performance-tests/package.json
kuhe: why not pin these? can installing latest create a build failure?

package.json
kuhe, line 52: does test:size now run every perf test?

kuhe (approved):

tests/performance-tests/runner/run.mjs
siddsriv: supposed to go to stdout by spec

tests/performance-tests/package.json
siddsriv: making them pinned, but i thought testing them against the latest versions of the bundlers would also help us see if there are regressions on the bundlers' end

package.json
siddsriv: test:size ran all bundler applications before (in tests/bundlers) and still does now. The behavior is the same, just moved to the new path. Might rename this as I add more perf-tests

tests/performance-tests/runner/prepare-artifact.mjs
siddsriv: added

tests/performance-tests/runner/run.mjs
siddsriv: added

Issue #7867, opened 2026-03-18
Entity Expansion Exceeded


aphex:
### Checkboxes for prior research

- [x] I've gone through [Developer Guide](https://docs.aws.amazon.com/sdk-for-javascript/v3/developer-guide) and [API reference](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest)
- [x] I've checked [AWS Forums](https://forums.aws.amazon.com) and [StackOverflow](https://stackoverflow.com/questions/tagged/aws-sdk-js).
- [x] I've searched for [previous similar issues](https://github.com/aws/aws-sdk-js-v3/issues) and didn't find any solution.

### Describe the bug

We are seeing a new issue after the latest update 

`Error: Entity expansion limit exceeded: 1598 > 1000` when trying to deploy our code SST project on seed.run. Pretty sure its connected to this

https://github.com/aws/aws-sdk-js-v3/issues/7853

and this

https://github.com/NaturalIntelligence/fast-xml-parser/commit/bd26122c838e6a55e7d7ac49b4ccc01a49999a01

### Regression Issue

- [x] Select this option if this issue appears to be a regression.

### SDK version number

@aws-sdk/package-name@version, ...

### Which JavaScript Runtime is this issue in?

Node.js

### Details of the browser/Node.js/ReactNative version

unknown (seed.run)

### Reproduction Steps

Do not have steps yet, still trying to sort through. We have a very large project so I assume it has something to do with scale. 

### Observed Behavior

```
Error: Entity expansion limit exceeded: 1598 > 1000

  Deserialization error: to see the raw response, inspect the hidden field {error}.$response on this object.

    at B.K [as replaceEntitiesValue] (/tmp/seed/source/node_modules/.pnpm/fast-xml-parser@5.5.6/node_modules/fast-xml-parser/lib/fxp.cjs:1:27694)

    at B.Y [as parseTextData] (/tmp/seed/source/node_modules/.pnpm/fast-xml-parser@5.5.6/node_modules/fast-xml-parser/lib/fxp.cjs:1:20545)
```

### Expected Behavior

CI/CD can deploy our stacks

### Possible Solution

Likely need to look into `maxTotalExpansions` as it seems to be limited here
[...]

PR #7868, opened 2026-03-18
fix(xml-builder): configure maxTotalExpansions on fast-xml-parser

kuhe:
### Issue
https://github.com/aws/aws-sdk-js-v3/issues/7867

### Description
Avoid `Error: Entity expansion limit exceeded`

### Testing
How was this change tested?

### Checklist
  - [x] It's not a feature.
  - [x] No streams here.

---

### DIFF
--- a/packages-internal/core/package.json
+++ b/packages-internal/core/package.json
@@ -82,7 +82,7 @@
   "license": "Apache-2.0",
   "dependencies": {
     "@aws-sdk/types": "workspace:^3.973.6",
-    "@aws-sdk/xml-builder": "workspace:^3.972.12",
+    "@aws-sdk/xml-builder": "workspace:^3.972.13",
     "@smithy/core": "^3.23.12",
     "@smithy/node-config-provider": "^4.3.12",
     "@smithy/property-provider": "^4.2.12",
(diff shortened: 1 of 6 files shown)


packages-internal/core/package.json
```diff
@@ -82,7 +82,7 @@
   "license": "Apache-2.0",
   "dependencies": {
     "@aws-sdk/types": "workspace:^3.973.6",
-    "@aws-sdk/xml-builder": "workspace:^3.972.12",
+    "@aws-sdk/xml-builder": "workspace:^3.972.13",
```
smilkuri, line 85: missing lock file

packages-internal/xml-builder/src/xml-parser.ts
```diff
@@ -1,14 +1,24 @@
...
+    maxTotalExpansions: Infinity,
+  },
   htmlEntities: true,
   ignoreAttributes: false,
   ignoreDeclaration: true,
   parseTagValue: false,
   trimValues: false,
   tagValueProcessor: (_: any, val: any) => (val.trim() === "" && val.includes("\n") ? "" : undefined),
-  maxNestedTags: 1024,
+  maxNestedTags: Infinity,
```
smilkuri, line 21: Can this cause any consequences like stack overflow?

kuhe: we've removed client side validation when it becomes a blocker because the XML parser instance is used against trusted XML only

PR #7878, opened 2026-03-23
fix(scripts): add escape character in changelog, message field

siddsriv:
### Issue
Internal JS-6720

### Description
The changelog script builds JSON by interpolating the git commit subject directly into a string template like `{"commit":"...","message":"<subject>"},`. If the subject contains double quotes, the resulting string is invalid JSON and JSON.parse throws.

The fix adds a step that escapes any `"` inside the message field to `\"` before parsing, so the JSON stays valid regardless of what's in the commit subject.

### Checklist
  - [x] It's not a feature.
  - [x] No streams here.

---


kuhe (approved):

siddsriv:
closing in favor of https://github.com/aws/aws-sdk-js-v3/pull/7879

PR #7880, opened 2026-03-24
chore(lib-storage): change default AbortController in Upload to the global implementation

SukkaW:
### Issue

N/A

### Description

`@aws-sdk/lib-storage` already targets Node.js >= 20.0.0, where Node.js introduced the native `AbortController` back in 15.4.0, which means we do not need to polyfill anymore and should prefer the native version.

In fact, even `@smithy/abort-controller` says itself is deprecated and you should prefer the native version.

To be backward compatible, I still import `AbortController` and `AbortSignal` types from `@smithy/types` package instead, so TypeScript won't scream at existing polyfilled external abort signal usage.

Do notice that `@smithy/abort-controller` is not completely removed, but moved to devDeps, because we still need to test against it to ensure backward compatibility.

I can start applying the same change to other `@smithy/abort-controller` usage in this repo once this PR is merged.

### Testing

To be backward compatible, I have updated `Upload.spec.ts` to test against both native and polyfilled `AbortSignal`.

### Checklist
  - [x] It's not a feature.
  - [x] No streams here.

---

### DIFF
--- a/lib/lib-storage/package.json
+++ b/lib/lib-storage/package.json
@@ -30,7 +30,6 @@
   },
   "license": "Apache-2.0",
   "dependencies": {
-    "@smithy/abort-controller": "^4.2.12",
     "@smithy/middleware-endpoint": "^4.4.27",
     "@smithy/protocol-http": "^5.3.12",
     "@smithy/smithy-client": "^4.12.7",
--- a/lib/lib-storage/src/Upload.spec.ts
+++ b/lib/lib-storage/src/Upload.spec.ts
@@ -917,7 +917,7 @@ describe(Upload.name, () => {
         client: new S3({}),
       });
 
-      expect((upload as any).abortController).toBeInstanceOf(AbortControllerPolyfill);
+      expect((upload as any).abortController).toBeInstanceOf(AbortController);
     });
 
     it("should calculate expectedPartsCount correctly when totalBytes is known", () => {
--- a/lib/lib-storage/src/Upload.spec.ts
+++ b/lib/lib-storage/src/Upload.spec.ts
@@ -8,7 +8,7 @@ import {
   S3Client,
   UploadPartCommand,
 } from "@aws-sdk/client-s3";
-import { AbortController } from "@smithy/abort-controller";
+import { AbortController as AbortControllerPolyfill } from "@smithy/abort-controller";
 import { EventEmitter, Readable } from "node:stream";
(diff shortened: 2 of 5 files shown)


lib/lib-storage/src/types.ts
```diff
@@ -6,7 +6,7 @@ import type {
   Tag,
   UploadPartCommandInput,
 } from "@aws-sdk/client-s3";
-import type { AbortController } from "@smithy/types";
+import type { AbortController as AbortControllerPolyfill } from "@smithy/types";
```
kuhe, line 9: a type import must be listed as a dependency

kuhe (changes requested):

lib/lib-storage/src/types.ts
SukkaW: Fixed in [`851934b` (this PR, after rebase)](https://github.com/aws/aws-sdk-js-v3/pull/7880/commits/851934b26ad699f05322516ce18368427d2ea64a)

Though I don't understand. Previously, we were importing `Endpoint` type from `@smithy/types` as well, yet it is still listed as a devDeps.

kuhe: It's most likely an error that continues to go unnoticed because other packages bring in `@smithy/types` at compile time. 

I'll check and update our validation scripts.

kuhe: related PR to fix validation: https://github.com/aws/aws-sdk-js-v3/pull/7884

lib/lib-storage/package.json
```diff
@@ -43,7 +43,7 @@
   },
   "devDependencies": {
     "@aws-sdk/client-s3": "workspace:3.1016.0",
-    "@smithy/types": "^4.13.1",
+    "@smithy/abort-controller": "^4.2.12",
```
kuhe:
```suggestion
```

lib/lib-storage/src/Upload.spec.ts
```diff
@@ -899,7 +917,7 @@ describe(Upload.name, () => {
         client: new S3({}),
       });
 
-      expect((upload as any).abortController).toBeInstanceOf(AbortController);
+      expect((upload as any).abortController).toBeInstanceOf(AbortControllerPolyfill);
```
kuhe: bad replace here

kuhe (approved):

PR #7881, opened 2026-03-24
docs(client-cognito-identity-provider): fix "ornull" typo in field docs

mquesada02:
**Issue**

N/A

**Description**

Fixes a minor typo in the `client-cognito-identity-provider` package documentation where "ornull" was written as a single word instead of "or null".

**Testing**

No testing required — this is a documentation-only change with no impact on runtime behavior.

**Checklist**

  - [x] It's not a feature.
- [ ] I added access level annotations e.g. `@public`, `@internal` tags and enabled doc generation on the package.
- [ ] Streams - how do they work??
  - [x] No streams here.


kuhe:
these files are generated from models outside this repository so the fix must be made by the service team

cc @aBurmeseDev

PR #7884, opened 2026-03-25
chore(lib): add dependency validation coverage for lib folder

kuhe:
### Issue
n/a, noticed in https://github.com/aws/aws-sdk-js-v3/pull/7880

### Description
add dependency validation for the `lib` folder.

### Testing
by running the validation

### Checklist
  - [x] It's not a feature.
  - [x] No streams here.

---

### DIFF
--- a/lib/lib-storage/package.json
+++ b/lib/lib-storage/package.json
@@ -32,7 +32,9 @@
   "dependencies": {
     "@smithy/abort-controller": "^4.2.12",
     "@smithy/middleware-endpoint": "^4.4.27",
+    "@smithy/protocol-http": "^5.3.12",
     "@smithy/smithy-client": "^4.12.7",
+    "@smithy/types": "^4.13.1",
     "buffer": "5.6.0",
     "events": "3.3.0",
     "stream-browserify": "3.0.0",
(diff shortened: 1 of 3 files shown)


lib/lib-dynamodb/package.json
```diff
@@ -41,7 +41,6 @@
     "@aws-sdk/client-dynamodb": "workspace:^3.1016.0"
   },
   "devDependencies": {
-    "@aws-sdk/client-dynamodb": "workspace:3.1016.0",
```
kuhe, line 44: it's ok to remove test dependencies from individual package.jsons within the monorepo, because the monorepo is the only supported development environment for this package, and all monorepo packages are globally resolvable within the monorepo.

lib/lib-storage/package.json
```diff
@@ -32,7 +32,9 @@
   "dependencies": {
     "@smithy/abort-controller": "^4.2.12",
     "@smithy/middleware-endpoint": "^4.4.27",
+    "@smithy/protocol-http": "^5.3.12",
     "@smithy/smithy-client": "^4.12.7",
+    "@smithy/types": "^4.13.1",
```
kuhe, line 37: types and protocol-http were missing as type imports

this was masked by compilation environments using `lib-storage` likely always having other packages that imported these, such as `client-s3`.

smilkuri (approved):

siddsriv (approved):

lib/lib-dynamodb/package.json
kuhe: no longer applicable to PR

PR #7897, opened 2026-03-27
chore(codegen): differentiate type imports

kuhe:
### Issue
codegen application for https://github.com/smithy-lang/smithy-typescript/commit/3a71b5741abe88724d1af4397ab0146ef1c2a2ba

### Description
changes unnecessary runtime imports to type imports

### Testing
CI, and this is covered by index tests (`make test-indices`).

### Checklist
  - [x] It's not a feature.
  - [x] No streams here.

---


clients/client-s3/src/S3Client.ts
```diff
@@ -60,29 +60,28 @@ import {
...
   Provider,
+  Provider as __Provider,
+  SdkStreamMixinInjector as __SdkStreamMixinInjector,
+  StreamCollector as __StreamCollector,
   StreamHasher as __StreamHasher,
+  UrlParser as __UrlParser,
   UserAgent as __UserAgent,
 } from "@smithy/types";
-import { Readable } from "stream";
+import type { Readable as Readable } from "node:stream";
```
kuhe, line 82: this is a workaround only affecting checksum clients

```java
writer.addImport("type Readable", "Readable", "node:stream")
```

The addTypeImport API does not support arbitrary strings, since the string-based module import is deprecated (too easy to write an import of a module that isn't declared in the dependency list). 

However, node packages don't need to be added to dependencies.

codegen/smithy-aws-typescript-codegen/src/main/java/software/amazon/smithy/aws/typescript/codegen/auth/http/integration/AwsSdkCustomizeEndpointRuleSetHttpAuthSchemeProvider.java
```diff
@@ -165,8 +165,7 @@ public void write(
                         );
                     Map<String, HttpAuthSchemeParameter> httpAuthSchemeParameters =
                         AuthUtils.collectHttpAuthSchemeParameters(effectiveHttpAuthSchemes.values());
-                    w.addDependency(TypeScriptDependency.SMITHY_TYPES);
```
kuhe, line 168: most addDependency calls can be removed because `add[Type]Import` when given a dependency object will automatically register it.

trivikr (approved):

PR #7901, opened 2026-03-29
chore(script): add script to find client changes since particular commit

smilkuri:
### Description
add script to generate client tarballs for changed clients since a given commit

### Checklist
  - [x] It's not a feature.
  - [x] No streams here.

---

### DIFF
--- a/scripts/generate-client-tarball/generate-client-tarball-since.mjs
+++ b/scripts/generate-client-tarball/generate-client-tarball-since.mjs
@@ -36,12 +36,12 @@ function getChangedClients(sinceCommit) {
 
   const clientNames = new Set();
   // Matches client package names in commit message of format "client-" like "client-s3".
-  const regex = /client-[a-z0-9-]+/g;
+  const regex = /^\w+\((client-\w+)\):/;
   for (const message of diffOutput.split("\n")) {
     const scope = message.split(":")[0];
     const match = scope.match(regex);
     if (match) {
-      clientNames.add(match[0]);
+      clientNames.add(match[1]);
     }
   }
   return [...clientNames];
@@ -60,7 +60,8 @@ for (const clientName of changedClients) {
       cwd: rootDir,
       stdio: "inherit",
     });
-  } catch {
+  } catch (e) {
+    console.error(`Failed to generate tarball for ${clientName}`, e);
     failed.push(clientName);
   }
 }
--- a/scripts/generate-client-tarball/generate-client-tarball-since.mjs
+++ b/scripts/generate-client-tarball/generate-client-tarball-since.mjs
@@ -25,25 +25,33 @@ if (!commitHash) {
 }
 
 function getChangedClients(sinceCommit) {
-  const diffOutput = execSync(`git diff --name-only ${sinceCommit}..HEAD`, {
+  const diffOutput = execSync(`git log ${sinceCommit}..HEAD --format=%s`, {
     cwd: rootDir,
     encoding: "utf-8",
   }).trim();
 
-  if (!diffOutput) return [];
+  if (!diffOutput) {
+    return [];
+  }
 
   const clientNames = new Set();
-  for (const file of diffOutput.split("\n")) {
-    const match = file.match(/^clients\/(client-[^/]+)\//);
-    if (match) clientNames.add(match[1]);
+  // Matches client package names in commit message of format "client-" like "client-s3".
+  const regex = /client-[a-z0-9-]+/g;
+  for (const message of diffOutput.split("\n")) {
+    const scope = message.split(":")[0];
+    const match = scope.match(regex);
+    if (match) {
+      clientNames.add(match[0]);
+    }
   }
   return [...clientNames];
 }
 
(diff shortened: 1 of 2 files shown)


scripts/generate-client-tarball/generate-client-tarball-since.mjs
```diff
@@ -0,0 +1,60 @@
...
+  const diffOutput = execSync(`git diff --name-only ${sinceCommit}..HEAD`, {
+    cwd: rootDir,
+    encoding: "utf-8",
+  }).trim();
+
+  if (!diffOutput) return [];
+
+  const clientNames = new Set();
+  for (const file of diffOutput.split("\n")) {
+    const match = file.match(/^clients\/(client-[^/]+)\//);
```
kuhe: describe regex with comment

scripts/generate-client-tarball/generate-client-tarball-since.mjs
```diff
@@ -0,0 +1,60 @@
...
+function getChangedClients(sinceCommit) {
+  const diffOutput = execSync(`git diff --name-only ${sinceCommit}..HEAD`, {
+    cwd: rootDir,
+    encoding: "utf-8",
+  }).trim();
+
+  if (!diffOutput) return [];
+
+  const clientNames = new Set();
+  for (const file of diffOutput.split("\n")) {
```
kuhe: why is diffOutput split named `file`? isn't this `commitMessage`?

scripts/generate-client-tarball/generate-client-tarball-since.mjs
```diff
@@ -0,0 +1,60 @@
...
+  process.exit(1);
+}
+
+function getChangedClients(sinceCommit) {
+  const diffOutput = execSync(`git diff --name-only ${sinceCommit}..HEAD`, {
+    cwd: rootDir,
+    encoding: "utf-8",
+  }).trim();
+
+  if (!diffOutput) return [];
```
kuhe: brace `{}` all control flow statements

scripts/generate-client-tarball/generate-client-tarball-since.mjs
kuhe: what is the purpose of `^clients\/` in the regex?

kuhe (changes requested):

scripts/generate-client-tarball/generate-client-tarball-since.mjs
smilkuri: No, it's not commit message, it's the file paths.

scripts/generate-client-tarball/generate-client-tarball-since.mjs
smilkuri: This `^clients\/` ensure we only match files in clients directory.

scripts/generate-client-tarball/generate-client-tarball-since.mjs
smilkuri: We should only take commit message here because taking file paths would give all paths of client packages containing version bump in package.json and changelog

kuhe: use commit messages only, because we often update every client at once with non-feature changes and that would cause a build to create 400+ tarballs

scripts/generate-client-tarball/generate-client-tarball-since.mjs
```diff
@@ -25,25 +25,32 @@ if (!commitHash) {
...
+  if (!diffOutput) {
+    return [];
+  }
 
   const clientNames = new Set();
-  for (const file of diffOutput.split("\n")) {
-    const match = file.match(/^clients\/(client-[^/]+)\//);
-    if (match) clientNames.add(match[1]);
+  // Matches client package names in commit message of format "client-" like "client-s3".
+  const regex = /client-[a-z0-9-]+/g;
```
kuhe: it has to be at the start of a line within the `()` parens. 

```js
/^\w+\(client-\w+\):
```
$1

scripts/generate-client-tarball/generate-client-tarball-since.mjs
kuhe, line 74: log error. this is running in a remote build and I don't want to guess what happened

kuhe (approved):

scripts/generate-client-tarball/generate-client-tarball-since.mjs
```diff
@@ -0,0 +1,80 @@
...
+  if (!diffOutput) {
+    console.info(`No output for: git log ${sinceCommit}..HEAD --format=%s`);
+    return [];
+  }
+
+  const clientNames = new Set();
+
+  // the "client-s3" part of a commit message like
+  // feat(client-s3): add some feature.
+  const regex = /^\w+\((client-[\w-]+)\):(.*?)/;
```
siddsriv: the `(.*?)` might be removable. what is it matching? it seems like `match[1]` will get the client name

siddsriv (approved):

scripts/generate-client-tarball/generate-client-tarball-since.mjs
kuhe: yes, it can be removed

Issue #7949, opened 2026-04-22
CVE-2026-41650 Vulnerability on fast-xml-parser


josejimenez-swipejobs:
### Checkboxes for prior research

- [x] I've gone through [Developer Guide](https://docs.aws.amazon.com/sdk-for-javascript/v3/developer-guide) and [API reference](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest)
- [x] I've checked [AWS Forums](https://forums.aws.amazon.com) and [StackOverflow](https://stackoverflow.com/questions/tagged/aws-sdk-js).
- [x] I've searched for [previous similar issues](https://github.com/aws/aws-sdk-js-v3/issues) and didn't find any solution.

### Describe the bug

Hello,

We are currently using `@aws-sdk/client-s3` version `3.1035.0` (latest), which depends on `@aws-sdk/xml-builder` internally for version `3.972.18`. This library imports fast-xml-parser version 5.5.8, which contains a known vulnerability listed here: [GHSA-gh4j-gqv2-49f6](https://github.com/advisories/GHSA-gh4j-gqv2-49f6). This issue is fixed in version `5.7.0`.

Would it be possible to update to this newer version to address the vulnerability?

### Regression Issue

- [ ] Select this option if this issue appears to be a regression.

### SDK version number

@aws-sdk/client-s3@3.1035.0 @aws-sdk/xml-builder@3.972.18

### Which JavaScript Runtime is this issue in?

Node.js

### Details of the browser/Node.js/ReactNative version

22.12.0

### Reproduction Steps

Run `npm audit` and getting this:

```
fast-xml-parser  <5.7.0
Severity: moderate
fast-xml-parser XMLBuilder: XML Comment and CDATA Injection via Unescaped Delimiters - https://github.com/advisories/GHSA-gh4j-gqv2-49f6
fix available via `npm audit fix --force`
Will install @aws-sdk/client-s3@3.893.0, which is a breaking change
node_modules/fast-xml-parser
  @aws-sdk/xml-builder  >=3.894.0
  Depends on vulnerable versions of fast-xml-parser
  node_modules/@aws-sdk/xml-builder
    @aws-sdk/core  >=3.894.0
    Depends on vulnerable versions of @aws-sdk/xml-builder
    node_modules/@aws-sdk/core
      @aws-sdk/client-s3  >=3.894.0
      Depends on vulnerable versions of @aws-sdk/core
[...]

PR #7964, opened 2026-04-27
fix(xml-builder): use xml 1.1 parsing behavior for entities

kuhe:
### Issue
https://github.com/aws/aws-sdk-js-v3/issues/7949

### Description
Parse XML 1.1 C0 control chars that can appear in S3 XML responses (such as Object keys) that otherwise declare XML 1.0.

### Testing
added unit, e2e tests

### Checklist
  - [ ] It's not a feature.
  - [x] No streams here.

---

### DIFF
--- a/packages-internal/xml-builder/src/xml-parser.ts
+++ b/packages-internal/xml-builder/src/xml-parser.ts
@@ -1,4 +1,28 @@
+// todo: package types are incorrect wrt exports.
+import type EntityDecoder from "@nodable/entities";
+import { COMMON_HTML, CURRENCY, XML } from "@nodable/entities";
 import { XMLParser } from "fast-xml-parser";
+
+// todo: package types are incorrect wrt exports.
+const { EntityDecoder: EntityDecoderImpl } = require("@nodable/entities");
+
+/**
+ * Custom entity decoder that preserves C0 control characters (e.g. U+0015)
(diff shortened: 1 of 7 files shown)


trivikr (approved):

packages-internal/xml-builder/src/xml-parser.ts
```diff
@@ -1,5 +1,29 @@
+// todo: package types are incorrect wrt exports.
+import type EntityDecoder from "@nodable/entities";
+import { COMMON_HTML, CURRENCY, XML } from "@nodable/entities";
 import { XMLParser } from "fast-xml-parser";
 
+// todo: package types are incorrect wrt exports.
+const { EntityDecoder: EntityDecoderImpl } = require("@nodable/entities");
```
akloss-terion, line 7: FWIW, jest does not like this. I can open an issue if you're interested. Let me know, thanks!

ajw015: Please do! I do see this use of `require` causing issues in ESM infrastructure

akloss-terion: https://github.com/aws/aws-sdk-js-v3/issues/7967

curtdept: this should be rolled back, its a pretty bad breakage

PR #7975, opened 2026-04-29
fix(cloudfront-signer): preserve URL encoding of resource path (#7571)

mohanrajvenkatesan23-04:
### Issue
#7571

### Description
PR #7237 replaced `new URL(baseUrl)` with raw string concatenation to stop the `URL` object from normalizing path segments such as `.` and `..`. As a side effect, the `URL` constructor's automatic percent-encoding of unsafe path characters (notably spaces -> `%20`) was lost, so `getSignedUrl({ url: "https://.../My File.pdf" })` returned a URL with a literal space in the path, breaking the signature once a client encoded the space before sending.

This PR adds two internal helpers in `packages/cloudfront-signer/src/sign.ts`:

- `encodeBaseUrlPath` - splits scheme+authority from path and only encodes the path portion.
- `encodePathPreservingEncoded` - percent-encodes characters outside the RFC 3986 path-allowed set while leaving existing `%XX` sequences untouched (no double-encoding).

The existing `encodeBaseUrlQuery` helper was hoisted out of `getSignedUrl`, extended to also encode the host-and-path, and applied once at the top of `getSignedUrl` so the policy resource and the final signed URL stay in sync. `getSignedCookies` and the `policy`-extracted `baseUrl` path are routed through the same encoder. `.` and `..` segments remain in the allowed set, preserving the original PR #7237 behavior.

Files changed:
- `packages/cloudfront-signer/src/sign.ts`
- `packages/cloudfront-signer/src/sign.spec.ts`

_generated by AI tools, and reviewed by Mohanraj Venkatesan_

### Testing
Three regression tests added in `packages/cloudfront-signer/src/sign.spec.ts`:

1. `encodes a raw space in the resource path as %20` - reproduces the bug; asserts the result contains `/My%20File.pdf`.
2. `does not double-encode a path that is already percent-encoded` - input `/My%20File.pdf` stays `%20` (not `%2520`).
3. `encodes the path and produces a signature that matches a policy whose resource uses the encoded path` - verifies the RSA-SHA1 signature against the encoded resource.

Result: `Tests 31 passed (31)` (28 pre-existing + 3 new). All previously-existing tests continue to pass, including those covering `.`/`..` path segments from PR #7237.

### Checklist
  - [x] It's not a feature.
  - [x] No streams here.

---

### DIFF
--- a/packages/cloudfront-signer/src/sign.spec.ts
+++ b/packages/cloudfront-signer/src/sign.spec.ts
@@ -383,6 +383,57 @@ describe("getSignedUrl", () => {
       expect(verifySignature(signatureQueryParam, policy)).toBeTruthy();
     });
   });
+
+  describe("should URL-encode special characters in the resource path (regression for #7571)", () => {
+    it("encodes a raw space in the resource path as %20", () => {
+      const urlWithSpace = "https://d111111abcdef8.cloudfront.net/My File.pdf";
+      const result = getSignedUrl({
+        url: urlWithSpace,
+        keyPairId,
+        dateLessThan,
+        privateKey,
+        passphrase,
+      });
+      expect(result).toContain("/My%20File.pdf");
+      expect(result).not.toContain("/My File.pdf");
+    });
+
+    it("does not double-encode a path that is already percent-encoded", () => {
+      const urlAlreadyEncoded = "https://d111111abcdef8.cloudfront.net/My%20File.pdf";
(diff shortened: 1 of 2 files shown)


trivikr:
Thank you, @mohanrajvenkatesan2304-hash, for submitting pull request.
We merged another pull request which was already open, which was fixing the same bug.

PR #7976, opened 2026-04-29
test(client-dynamodb): fix table cleanup criteria

kuhe:
Avoid accidentally deleting a fresh table from another test.

### DIFF
--- a/clients/client-dynamodb/test/dynamodb-features.e2e.spec.ts
+++ b/clients/client-dynamodb/test/dynamodb-features.e2e.spec.ts
@@ -73,7 +73,7 @@ describe(DynamoDB.name, () => {
         }
         try {
           const desc = await client.describeTable({ TableName: name });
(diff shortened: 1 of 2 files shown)


clients/client-dynamodb/test/dynamodb-features.e2e.spec.ts
```diff
@@ -73,7 +73,7 @@ describe(DynamoDB.name, () => {
         }
         try {
           const desc = await client.describeTable({ TableName: name });
-          if (desc?.Table?.CreationDateTime?.getTime?.() ?? -Infinity < eightHoursAgo) {
```
kuhe, line 76: change to positive Inf: if we don't know the creation time of a table, we don't delete it

the other fix is to wrap the `a ?? b` in parens

trivikr (approved):

Issue #7913, opened 2026-04-03
fromHttp credential provider leaks sockets when EKS Pod Identity token expires — new http.Agent created per invocation, never destroyed


tomaszhrem:
### Checkboxes for prior research

- [x] I've gone through [Developer Guide](https://docs.aws.amazon.com/sdk-for-javascript/v3/developer-guide) and [API reference](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest)
- [x] I've checked [AWS Forums](https://forums.aws.amazon.com) and [StackOverflow](https://stackoverflow.com/questions/tagged/aws-sdk-js).
- [x] I've searched for [previous similar issues](https://github.com/aws/aws-sdk-js-v3/issues) and didn't find any solution.

### Describe the bug

When using EKS Pod Identity with the default credential chain, the `fromHttp` credential provider in `@aws-sdk/credential-provider-http` leaks TCP sockets to the Pod Identity Agent (`169.254.170.23:80`). Under sustained credential refresh failures (caused by an expired Pod Identity token), this leads to tens of thousands of ESTABLISHED sockets accumulating within minutes, eventually exhausting file descriptors and crashing the process.

### Root cause (in our case)

There are two compounding issues:

1. **Token expiration trigger (EKS/infra side):** The EKS Pod Identity token (`/var/run/secrets/pods.eks.amazonaws.com/serviceaccount/eks-pod-identity-token`) has a 1-hour TTL. A race condition between token expiry and kubelet rotation causes the Pod Identity Agent to reject credential requests, triggering rapid SDK retry cycles.

2. **Socket leak in `fromHttp` (SDK side):** The default credential provider chain appears to call `fromHttp()` repeatedly (not just the returned provider function). Each call to `fromHttp()` creates a **new `NodeHttpHandler`** via `NodeHttpHandler.create()`, which internally creates a new `http.Agent({ keepAlive: true, maxSockets: 50 })`. The previous handler's `destroy()` is never called, so its sockets remain in ESTABLISHED state indefinitely.

### Relevant source (`fromHttp.js`):

```javascript
const requestHandler = node_http_handler_1.NodeHttpHandler.create({
    requestTimeout: options.timeout ?? 1000,
    connectionTimeout: options.timeout ?? 1000,
});
return retryWrapper(async () => {
    const result = await requestHandler.handle(request);
    return getCredentials(result.response);
}, options.maxRetries ?? 3, options.timeout ?? 1000);
```

[...]

PR #7995, opened 2026-05-06
fix(credential-provider-http): close sockets after connection attempts

siddsriv:
### Issue
#7913 

### Description
Ensures the `NodeHttpHandler` created by `fromHttp()` is destroyed after each credential fetch (success or failure), preventing socket accumulation when credential refresh fails repeatedly.

### Testing
new unit test
 ```console
 RUN  v4.0.17 /../aws-sdk-js-v3/packages-internal/credential-provider-http

 ✓ src/fromHttp/checkUrl.spec.ts (8 tests) 4ms
 ✓ src/fromHttp/requestHelpers.spec.ts (4 tests) 7ms
 ✓ src/fromHttp/fromHttp.spec.ts (7 tests) 3012ms
     ✓ destroys the request handler even when the request fails  3005ms

 Test Files  3 passed (3)
      Tests  19 passed (19)
```

new integ test:
```console
 RUN  v4.0.17 /../aws-sdk-js-v3/packages-internal/credential-provider-http

 ✓ src/fromHttp/fromHttp.integ.spec.ts (2 tests) 2644ms
   ✓ fromHttp socket cleanup (2)
     ✓ destroys sockets after each credential fetch 75ms
     ✓ destroys sockets even when requests fail  2563ms

 Test Files  1 passed (1)
      Tests  2 passed (2)
```

### Checklist
  - [x] It's not a feature.
  - [x] No streams here.

---

### DIFF
--- a/packages-internal/credential-provider-http/src/fromHttp/fromHttp.ts
+++ b/packages-internal/credential-provider-http/src/fromHttp/fromHttp.ts
@@ -16,12 +16,6 @@ const AWS_CONTAINER_AUTHORIZATION_TOKEN_FILE = "AWS_CONTAINER_AUTHORIZATION_TOKE
 const AWS_CONTAINER_AUTHORIZATION_TOKEN = "AWS_CONTAINER_AUTHORIZATION_TOKEN";
 
 /**
- * Cached request handler shared across all fromHttp provider instances.
- * This prevents socket leaks from creating a new NodeHttpHandler (and http.Agent) per credential refresh.
- */
-let cachedHandler: InstanceType<typeof NodeHttpHandler> | undefined;
-
-/**
  * Creates a provider that gets credentials via HTTP request.
  */
 export const fromHttp = (options: FromHttpOptions = {}): AwsCredentialIdentityProvider => {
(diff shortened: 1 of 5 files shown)


trivikr (approved):

kuhe (changes requested):
this creates asymmetry that is unclear why the memoized chain, which is already cached, further caches an intermediate function, but only one of them

fix should be in the code that manages the lifecycle of the request handler and its sockets

packages-internal/credential-provider-http/src/fromHttp/fromHttp.ts
```diff
@@ -66,10 +72,13 @@ Set AWS_CONTAINER_CREDENTIALS_FULL_URI or AWS_CONTAINER_CREDENTIALS_RELATIVE_URI
...
-  });
+  if (!cachedHandler) {
+    cachedHandler = NodeHttpHandler.create({ connectionTimeout: options.timeout ?? 1000 }) as InstanceType<
+      typeof NodeHttpHandler
+    >;
+  }
+  const requestHandler = cachedHandler;
+  const requestTimeout = options.timeout ?? 1000;
 
   return retryWrapper(
```
kuhe: since this isn't a high volume request, I think the simpler fix is to call destroy on the handler when the retryWrapper finishes.

packages-internal/credential-provider-http/src/fromHttp/fromHttp.ts
kuhe: why `InstanceType<typeof NodeHttpHandler>` instead of the inferred type which should be `NodeHttpHandler`?

kuhe (approved):

packages-internal/credential-provider-http/src/fromHttp/fromHttp.ts
siddsriv: no difference, will change it

kuhe (approved):

PR #8003, opened 2026-05-11
chore(signature-v4-multi-region): invert dependency on middleware-sdk-s3

kuhe:
### Issue

### Description
This inverts the dependency between sigv4-multi-region and middleware-sdk-s3.

Because multi-region currently depends on the S3Express signing behavior, but was also the general provider for sigv4a within the SDK, and now that STS uses sigv4a and STS has always itself been used in the default provider chain, all clients ended up importing middleware-sdk-s3.

This PR doesn't change the behavior of the S3Express signer, but puts a generic label on it, "SignatureV4SignWithCredentials". Then, middleware-sdk-s3's SignatureV4S3Express now extends this class rather than providing an implementation to multi-region.

### Testing
CI

### Checklist
  - [x] It's not a feature.
  - [x] No streams here.

---


packages-internal/signature-v4-multi-region/src/SignatureV4SignWithCredentials.ts
```diff
@@ -0,0 +1,82 @@
...
+    secretAccessKey: credentials.secretAccessKey,
+    expiration: credentials.expiration,
+  };
+}
+
+function setSingleOverride(
+  privateAccess: { credentialProvider: Provider<AwsCredentialIdentity> },
+  credentialsWithoutSessionToken: Omit<AwsCredentialIdentity, "sessionToken">
+) {
+  const currentCredentialProvider = privateAccess.credentialProvider;
```
smilkuri, line 77: Any reason for removing the `setTimeout` ?

kuhe: it throws in a way that isn't catchable, and the thing it was defending against can be asserted via unit tests instead

smilkuri (approved):

Issue #8004, opened 2026-05-11
Package consolidation


kuhe:
Follows https://github.com/smithy-lang/smithy-typescript/issues/1980

See planning document from https://github.com/aws/aws-sdk-js-v3/pull/8002.

This is a PR aggregation issue for the project to consolidate the number of NPM packages down to a smaller set of dependencies.

kuhe:
This is complete for now. I don't see a need to consolidate the Glacier group, since the client is going to become unused.

PR #8006, opened 2026-05-12
chore(scripts): add submodule variant api surface parity linting

kuhe:
### Issue

### Description
Adds/fixes some validation scripts.

- packages-internal now linted by the package-json rules
- updated dist-cjs Inliner to build submodule variants (none exist yet) and updated to the same validations as in smithy-ts

### Testing
ci

### Checklist
  - [x] It's not a feature.
  - [x] No streams here.

---


packages-internal/nested-clients/src/submodules/sts/defaultStsRoleAssumers.ts
```diff
@@ -222,26 +222,6 @@ export const getDefaultRoleAssumerWithWebIdentity = (
...
 
-/**
- * The default credential providers depend STS client to assume role with desired API: sts:assumeRole,
- * sts:assumeRoleWithWebIdentity, etc. This function decorates the default credential provider with role assumers which
- * encapsulates the process of calling STS commands. This can only be imported by AWS client packages to avoid circular
- * dependencies.
- *
- * @internal
- */
-export const decorateDefaultCredentialProvider =
```
kuhe, line 233: function deletion - this function, unlike the function with the same name in the adjacent defaultRoleAssumers.ts file, is not exported from the package, and can be deleted.

siddsriv (approved):

Issue #8005, opened 2026-05-12
RequestTimeTooSkewed not auto-corrected for parallel requests


shadow-light:
### Checkboxes for prior research

- [x] I've gone through [Developer Guide](https://docs.aws.amazon.com/sdk-for-javascript/v3/developer-guide) and [API reference](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest)
- [x] I've checked [AWS Forums](https://forums.aws.amazon.com) and [StackOverflow](https://stackoverflow.com/questions/tagged/aws-sdk-js).
- [x] I've searched for [previous similar issues](https://github.com/aws/aws-sdk-js-v3/issues) and didn't find any solution.

### Describe the bug

When the system clock is out-of-sync, the SDK includes auto-retry with an offset applied to correct the time. This works. However, when running simultaneous requests, only one of them will be retried, likely the first to finish. It would appear that when requests have been made prior to the offset being calculated, when they return with RequestTimeTooSkewed the client believes it has already attempted to fix the offset and failed and so never retries them.

### Regression Issue

- [x] Select this option if this issue appears to be a regression.

### SDK version number

@aws-sdk/client-s3@3.1006.0

### Which JavaScript Runtime is this issue in?

Browser

### Details of the browser/Node.js/ReactNative version

Chrome/146.0.7680.80 Electron/41.0.3

### Reproduction Steps

Execute at least three S3 PutObject requests simultaneously on a system that has a clock wrong by at least 1 hour.

### Observed Behavior

All requests initially fail due to time skew. One request is retried and succeeds after the SDK corrects the time, but the other requests are never retried.

### Expected Behavior

All requests are retried with the corrected time.

### Possible Solution

I added this middleware that fixes the issue:

```js
client.middlewareStack.add((next, context) => {
    return async (args) => {
        try {
            return await next(args)
        } catch (error) {

            // If not a clock stew error, simply rethrow
            if ((error as AwsError)?.name !== 'RequestTimeTooSkewed') {
                throw error
            }

            // AWS SDK has built-in retry with clock correction offset.
            // Since RequestTimeTooSkewed was raised and not auto-corrected, it's likely this
            // was an initial parallel request that fired before the correction could happen.
            // When it returns the client thinks it tried with the corrected time but actually
            // didn't have access to the correction when it was first sent.
            // So force a retry, likely with a time offset already now applied
            return await next(args)
        }
    }
}, {
    name: 'clock_skew_handler',
[...]

Issue #7605, opened 2025-12-24
Intermittent InvalidSignatureException: Signature expired error in AWS Lambda


minh-le-q:
### Checkboxes for prior research

- [x] I've gone through [Developer Guide](https://docs.aws.amazon.com/sdk-for-javascript/v3/developer-guide) and [API reference](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest)
- [x] I've checked [AWS Forums](https://forums.aws.amazon.com) and [StackOverflow](https://stackoverflow.com/questions/tagged/aws-sdk-js).
- [x] I've searched for [previous similar issues](https://github.com/aws/aws-sdk-js-v3/issues) and didn't find any solution.

### Describe the bug

When using AWS SDK for JavaScript v3 from an AWS Lambda, we see intermittent InvalidSignatureException errors with messages like:

```
Signature expired: ... is now earlier than ... - 5 min.
```

From CloudWatch logs, we see these occur on the first DynamoDB request after the Lambda has been idle (frozen) for several minutes, and subsequent requests in the same warm environment succeed.
This happens despite AWS SDK v3’s documented automatic clock skew detection and correction under the `correctClockSkew` section [here](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/migrating/notable-changes/).

### Regression Issue

- [ ] Select this option if this issue appears to be a regression.

### SDK version number

"@aws-sdk/client-dynamodb": "^3.614.0", "@aws-sdk/lib-dynamodb": "^3.614.0"

### Which JavaScript Runtime is this issue in?

Node.js

### Details of the browser/Node.js/ReactNative version

Node.js 20.x on AWS Lambda constructed by the NodejsFunction construct via "aws-cdk-lib": "^2.158.0". The Lambda function is not set up with any preserved concurrency.

### Reproduction Steps

Here is an example Lambda code with the `handler` function as the Lambda entrypoint to reproduce the issue

```typescript
import { DynamoDBClient } from "@aws-sdk/client-dynamodb";
import { DynamoDBDocumentClient, GetCommand } from "@aws-sdk/lib-dynamodb";

// Intentionally skew the clock to simulate freeze/thaw or bad offset
const ddb = new DynamoDBClient({
  region: "us-east-1",
  systemClockOffset: -3 * 60 * 60 * 1000, // -3 hours
});

const doc = DynamoDBDocumentClient.from(ddb);

export const handler = async () => {
  await doc.send(
[...]

PR #8009, opened 2026-05-12
fix(core): fix clock-skew retry for concurrent requests and JSON-protocol services

Zelys-DFKH:
## Problem

Two bugs in `AwsSdkSigV4Signer.errorHandler`.

**Concurrent correction false negative.** When multiple requests hit clock-skew errors simultaneously, only the first reaches `getUpdatedSystemClockOffset` with a stale offset. By the time the rest arrive, a concurrent request has already corrected `config.systemClockOffset`. Their offsets don't change, so `clockSkewCorrected` never gets set, so they don't retry.

**DynamoDB (and other JSON-protocol services) never correct at all.** XML services like S3 include a `ServerTime` field in the error body. DynamoDB doesn't. The HTTP `Date` header isn't guaranteed on error responses either. With no server time available, `errorHandler` skips the correction block entirely. The client gets an `InvalidSignatureException` back with no retry signal.

## Fix

**Pre-request snapshot.** `sign()` now saves `config.systemClockOffset` into `signingProperties._preRequestSystemClockOffset` before calling the actual signer. The error handler compares against that value, not against zero, so it can tell whether the offset changed since this specific request started, regardless of what concurrent requests did in the meantime. `AwsSdkSigV4ASigner` overrides `sign()` but inherits `errorHandler`, so it gets the same treatment.

**Message-based server time fallback.** AWS always embeds the server time in the error message itself: `Signature expired: ... (20251224T185913Z - 5 min. ahead of your clock)`. `getServerTimeFromSkewMessage()` parses this as a third fallback after `ServerTime` and the `Date` header. The regex requires the timestamp to appear inside a `(` to avoid false matches on the expired-signature timestamp earlier in the same message.

The `isKnownClockSkewError` guard (`!!(errorException.ServerTime ?? messageServerTime)`) keeps the `Date`-header path conservative: `clockSkewCorrected` only gets set via the concurrent-correction branch when a reliable clock-skew indicator is present, not just any HTTP error that happened to include a date header.

## Tests

Seven tests total. Four are new, one is updated:

- Offset changes when `systemClockOffset` is non-zero (pre-existing)
- `ServerTime` concurrent correction — adds `_preRequestSystemClockOffset: 0` to reflect real runtime state (updated)
- DynamoDB: message parsing, fresh start — no `ServerTime`, no `Date` header, message has compact timestamp (new)
- DynamoDB: message parsing, concurrent correction — offset already corrected by a concurrent request (new)
- Message with no compact timestamp — `clockSkewCorrected` stays `undefined` (new)
- Bare non-parenthesized timestamp in message — must not match (new)
- Throttling error with `Date` header but no `ServerTime` — `clockSkewCorrected` stays `undefined` (pre-existing)

Reported in #7605 and #8005.

---

*Developed by Zelys with assistance from Claude.*

### DIFF
--- a/packages-internal/core/src/submodules/httpAuthSchemes/aws_sdk/AwsSdkSigV4Signer.ts
+++ b/packages-internal/core/src/submodules/httpAuthSchemes/aws_sdk/AwsSdkSigV4Signer.ts
@@ -44,20 +44,6 @@ interface AwsSdkSigV4AuthSigningProperties {
 
 /**
  * @internal
- * Parses the compact AWS server timestamp from a clock-skew error message.
- * AWS embeds the server time as e.g. "(20251224T185913Z - 5 min.)" in the message
- * for services that omit a Date response header and ServerTime body field.
- */
-const getServerTimeFromSkewMessage = (message: string | undefined): string | undefined => {
-  if (!message) return undefined;
-  const match = message.match(/\((\d{4})(\d{2})(\d{2})T(\d{2})(\d{2})(\d{2})Z/);
-  if (!match) return undefined;
-  const [, year, month, day, hour, min, sec] = match;
-  return `${year}-${month}-${day}T${hour}:${min}:${sec}Z`;
-};
-
-/**
- * @internal
  */
 interface AwsSdkSigV4Exception extends ServiceException {
   ServerTime?: string;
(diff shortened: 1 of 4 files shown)


packages-internal/core/src/submodules/httpAuthSchemes/aws_sdk/AwsSdkSigV4Signer.ts
```diff
@@ -132,7 +132,9 @@ export class AwsSdkSigV4Signer implements HttpSigner {
         const config = throwSigningPropertyError("config", signingProperties.config as AwsSdkSigV4Config | undefined);
         const initialSystemClockOffset = config.systemClockOffset;
         config.systemClockOffset = getUpdatedSystemClockOffset(serverTime, config.systemClockOffset);
-        const clockSkewCorrected = config.systemClockOffset !== initialSystemClockOffset;
+        const clockSkewCorrected =
+          config.systemClockOffset !== initialSystemClockOffset ||
+          (!!(error as AwsSdkSigV4Exception).ServerTime && config.systemClockOffset !== 0);
```
kuhe: I think `config.systemClockOffset !== 0` is going to be brittle, since a non-zero offset doesn't imply much about the local error that we are handling in this function.

Is there instead a way to get `initialSystemClockOffset` prior to the request going out? Then, the check for `clockSkewCorrected` should detect whether a parallel request adjusted the skew between the request start and this error handler running.

kuhe (dismissed):

packages-internal/core/src/submodules/httpAuthSchemes/aws_sdk/AwsSdkSigV4Signer.ts
Zelys-DFKH: Good catch. `!== 0` was too broad. I've replaced it with a pre-request snapshot: `sign()` saves `config.systemClockOffset` into `signingProperties` before calling `signer.sign()`, and the error handler now compares against that rather than zero. Answers the actual question: did the offset change since this request started? `AwsSdkSigV4ASigner` gets the same treatment since it overrides `sign()` but inherits `errorHandler`.

While I was in here, I noticed that JSON-protocol services like DynamoDB can never benefit from clock-skew correction at all: no `ServerTime` in the body, and AWS doesn't guarantee a `Date` header on error responses. The error message itself always has the server time though: `(20251224T185913Z - 5 min.)`. Added `getServerTimeFromSkewMessage()` as a third fallback.

kuhe:
moved to https://github.com/aws/aws-sdk-js-v3/pull/8078

Issue #7578, opened 2025-12-15
AWS SES requires template subject despite being marked optional


acarotin:
### Checkboxes for prior research

- [x] I've gone through [Developer Guide](https://docs.aws.amazon.com/sdk-for-javascript/v3/developer-guide) and [API reference](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest)
- [x] I've checked [AWS Forums](https://forums.aws.amazon.com) and [StackOverflow](https://stackoverflow.com/questions/tagged/aws-sdk-js).
- [x] I've searched for [previous similar issues](https://github.com/aws/aws-sdk-js-v3/issues) and didn't find any solution.

### Describe the bug

According to the [EmailTemplateContent](https://github.com/aws/aws-sdk-js-v3/blob/7e8a34ea9e759e533defdb14c8b69bc46a2c4225/clients/client-sesv2/src/models/models_0.ts#L568) type definition, the subject field is marked as optional.
However, when creating an AWS SES template without providing a subject, SES returns an error and rejects the request.

### Regression Issue

- [ ] Select this option if this issue appears to be a regression.

### SDK version number

@aws-sdk/client-sesv2@3.946.0

### Which JavaScript Runtime is this issue in?

Node.js

### Details of the browser/Node.js/ReactNative version

v22.21.1

### Reproduction Steps

```ts
await client.send(new CreateEmailTemplateCommand({
	TemplateName: 'Name',
	TemplateContent: {
		Text: 'text',
	},
}));
```

### Observed Behavior

AWS SES rejects templates without a subject, returning an error
```json
"errorType": "BadRequestException",
"errorMessage": "The subject must be specified.",
```

### Expected Behavior

Templates should be accepted when the subject field is omitted, as indicated by the optional type definition.

### Possible Solution

Either the subject field is truly optional (which I believe it should be), and the API call should succeed without providing one, or the type definition should be updated to mark it as required.

aBurmeseDev:
Hi @acarotin - thanks for reporting this. I was able to confirm that the subject appears to be required by service API to create the email template and that error is thrown without it. Since service model change has to be done by upstream service team, I need to verify the correct behavior of this part of the model with SES team internally.

Will circle back here when service team responds. (ref _V2038667761_)

PR #8011, opened 2026-05-13
fix(sesv2): mark EmailTemplateContent.Subject as required

Zelys-DFKH:
### Issue
#7578

### Description

`EmailTemplateContent.Subject` is typed as `Subject?: string | undefined` (optional) in the generated TypeScript, but the SES v2 API always rejects `CreateEmailTemplate` requests without it:

```
BadRequestException: The subject must be specified.
```

The reporter in #7578 confirmed this, and maintainer @aBurmeseDev verified it as well. The mismatch tells TypeScript users the field is optional when the API won't accept that.

### Fix

Two-file change:

1. `codegen/sdk-codegen/aws-models/sesv2.json`: adds `smithy.api#required` to the `Subject` member of `com.amazonaws.sesv2#EmailTemplateContent`
2. `clients/client-sesv2/src/models/models_0.ts`: removes the `?` optional marker from `Subject` (the generated output of that model change)

The diff is 3 lines across 2 files.

### Testing

No runtime logic changes. The generated TypeScript now correctly reflects the API contract that has always been enforced server-side. Existing callers that already pass `Subject` are unaffected; callers that omit it will get a compile-time error instead of a runtime rejection.

### Checklist
- [x] It's not a feature.
- [x] No streams here.

Developed by Zelys with assistance from Claude.

### DIFF
--- a/clients/client-sesv2/src/models/models_0.ts
+++ b/clients/client-sesv2/src/models/models_0.ts
@@ -570,7 +570,7 @@ export interface EmailTemplateContent {
    * <p>The subject line of the email.</p>
    * @public
    */
-  Subject?: string | undefined;
+  Subject: string | undefined;
 
   /**
    * <p>The email body that will be visible to recipients whose email clients do not display
--- a/codegen/sdk-codegen/aws-models/sesv2.json
+++ b/codegen/sdk-codegen/aws-models/sesv2.json
@@ -4231,7 +4231,8 @@
         "Subject": {
           "target": "com.amazonaws.sesv2#EmailTemplateSubject",
           "traits": {
-            "smithy.api#documentation": "<p>The subject line of the email.</p>"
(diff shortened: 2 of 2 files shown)


aBurmeseDev:
Hey @Zelys-DFKH - thanks for the PR! However, this change needs to be upstreamed by service team as I mentioned in my [comment](https://github.com/aws/aws-sdk-js-v3/issues/7578#issuecomment-3657944568) previously. 

Upon checking internally, it looks like the patch was deployed recently. Are you able to verify it with latest version?

Zelys-DFKH:
Thanks for checking internally, @aBurmeseDev. I pulled down 3.1045.0 and `Subject` still shows as `Subject?: string | undefined` there. Did the fix make it required in the TypeScript types, or did the SES team update the API to accept requests without Subject? Happy to verify once I know which direction to look.

PR #8025, opened 2026-05-15
fix(client-sts): update imports to new module locations

kuhe:
### Issue
https://github.com/aws/aws-sdk-js-v3/issues/8027

### Description
- regenerate the imports for client-sts
- clean up imports in the `./private` folder's non-code-generated packages.

### Testing
CI

### Checklist
  - [x] It's not a feature.
  - [x] No streams here.

---


clients/client-sts/src/defaultStsRoleAssumers.ts
```diff
@@ -1,8 +1,7 @@
 // smithy-typescript generated code
 // Please do not touch this file. It's generated from template in:
 // https://github.com/aws/aws-sdk-js-v3/blob/main/codegen/smithy-aws-typescript-codegen/src/main/resources/software/amazon/smithy/aws/typescript/codegen/sts-client-defaultStsRoleAssumers.ts
-import { setCredentialFeature } from "@aws-sdk/core/client";
-import { stsRegionDefaultResolver } from "@aws-sdk/region-config-resolver";
+import { setCredentialFeature, stsRegionDefaultResolver } from "@aws-sdk/core/client";
```
kuhe, line 4: The template was already updated

kuhe: it's possible that changes to how often the Java codegen is recompiled caused the resource file which contains the import path not to be picked up.

In the future I'd like to remove the resource file and use Java string templates instead. The resource file doesn't integrate well with TS tools in its current location anyway.

trivikr (approved):

siddsriv (approved):

Issue #8020, opened 2026-05-14
Upload misses implementation for a few headers


chrisguttandin:
### Checkboxes for prior research

- [x] I've gone through [Developer Guide](https://docs.aws.amazon.com/sdk-for-javascript/v3/developer-guide) and [API reference](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest)
- [x] I've checked [AWS Forums](https://forums.aws.amazon.com) and [StackOverflow](https://stackoverflow.com/questions/tagged/aws-sdk-js).
- [x] I've searched for [previous similar issues](https://github.com/aws/aws-sdk-js-v3/issues) and didn't find any solution.

### Describe the bug

I'm trying to use `Upload` from `lib-storage` with the `IfNoneMatch` param. It throws an error saying that it is not yet implemented.

### Regression Issue

- [ ] Select this option if this issue appears to be a regression.

### SDK version number

@aws-sdk/lib-storage@3.1044.0

### Which JavaScript Runtime is this issue in?

Node.js

### Details of the browser/Node.js/ReactNative version

v22.22.2

### Reproduction Steps

This will trigger the error:

```js
import { S3Client } from '@aws-sdk/client-s3';
import { Upload } from '@aws-sdk/lib-storage';

const s3Client = new S3Client(/* ... */);
const upload = new Upload({ client: s3Client, params: { Body, Bucket: '...', IfNoneMatch: '*', Key: '...' } });

await upload.done();
```

### Observed Behavior

It will throw an error.

### Expected Behavior

I think this could be implemented.

### Possible Solution

It looks like the `params` are internally handled already. The `Body` for example gets replaced ...

https://github.com/aws/aws-sdk-js-v3/blob/a66433550dc5bfd0b3228222ad096bfb25d042c0/lib/lib-storage/src/Upload.ts#L146

... or set to undefined ...

https://github.com/aws/aws-sdk-js-v3/blob/a66433550dc5bfd0b3228222ad096bfb25d042c0/lib/lib-storage/src/Upload.ts#L227

... if necessary. I think a similar thing could be done for the params that should only be set for the `PutObjectCommand` and `CompleteMultipartUploadCommand` respectively.

### Additional Information/Context

I'm happy to provide PR for this if it helps.

aBurmeseDev:
Hi @chrisguttandin, thanks for checking in! I gave this a shot but couldn't actually repro it.

[...]

PR #8029, opened 2026-05-16
fix(lib-storage): replace UploadPart spread with field allowlist

Zelys-DFKH:
### Issue
#8020

### Description
`Upload.__doConcurrentUpload` spreads `...this.params` (typed as `PutObjectCommandInput`) into every `UploadPartCommand`. `PutObjectCommandInput` includes object-creation-only fields — `IfNoneMatch`, `ContentType`, `ContentMD5`, `Metadata`, `ACL`, `StorageClass`, `ObjectLock*`, and others — and S3 rejects them on the UploadPart API with `NotImplemented` (501) or `MalformedXML`.

This PR replaces the spread with an explicit allowlist of the fields `UploadPartRequest` actually accepts: `Bucket`, `Key`, the full `Checksum*` family, `SSECustomerAlgorithm/Key/KeyMD5`, `RequestPayer`, and `ExpectedBucketOwner`. `Body`, `PartNumber`, `UploadId`, and `ContentLength` are set explicitly as before.

Two intentional omissions:
- `ContentMD5` — `this.params.ContentMD5` is the whole-object MD5. Forwarding it to each part would be wrong. The `UploadPartRequest.ContentMD5` field exists for per-part hashes, which callers can supply via request interceptors.
- `ContentLength` — already explicitly set to `undefined` in the existing code; no change.

`CompleteMultipartUploadCommand` continues to receive `...this.params` at the end of the flow, so `IfNoneMatch` and other `CompleteMultipartUpload`-valid fields still work correctly there.

**Note on PR #7994:** That PR strips `ContentMD5` via a denylist. This PR uses a full allowlist instead — more defensive against future `PutObjectCommandInput` additions silently leaking through. Happy to coordinate if the team prefers the denylist approach.

### Testing
Added a regression test in the `large buffers` describe block that:
- forces the multipart path
- confirms `IfNoneMatch`, `ContentType`, and `ContentMD5` don't reach `UploadPartCommand`
- confirms `SSECustomerAlgorithm` (valid for UploadPart) flows through
- confirms `CompleteMultipartUploadCommand` still receives `IfNoneMatch`

All 53 existing tests pass.

### Checklist
  - [x] It's not a feature.
  - [x] No streams here.

### DIFF
--- a/lib/lib-storage/src/Upload.spec.ts
+++ b/lib/lib-storage/src/Upload.spec.ts
@@ -506,6 +506,69 @@ describe(Upload.name, () => {
         expect(PutObjectCommand).toHaveBeenCalledTimes(0);
       });
     });
+
+    it("should not forward PUT-only or object-level Checksum* params to UploadPart", async () => {
+      const largeBuffer = Buffer.from("#".repeat(MOCK_PART_SIZE + 10));
+      const upload = new Upload({
+        params: {
+          ...params,
+          Body: largeBuffer,
+          IfNoneMatch: "*",
+          ContentType: "application/json",
+          ContentMD5: "object-level-md5-base64==",
+          ChecksumSHA256: "object-level-sha256-base64==",
+          SSECustomerAlgorithm: "AES256",
+        },
+        partSize: MOCK_PART_SIZE,
+        client: new S3({}),
+      });
+
+      await upload.done();
+
+      expect(UploadPartCommand).toHaveBeenCalledTimes(2);
+
+      // PUT-only fields S3 rejects on UploadPart (501 NotImplemented).
+      expect(UploadPartCommand).not.toHaveBeenCalledWith(expect.objectContaining({ IfNoneMatch: "*" }));
+      expect(UploadPartCommand).not.toHaveBeenCalledWith(expect.objectContaining({ ContentType: "application/json" }));
+
+      // Object-level checksums: forwarding them as per-part values causes BadDigest (#6742).
+      expect(UploadPartCommand).not.toHaveBeenCalledWith(
+        expect.objectContaining({ ContentMD5: "object-level-md5-base64==" })
+      );
+      expect(UploadPartCommand).not.toHaveBeenCalledWith(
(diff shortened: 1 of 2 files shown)


danyalahmed1995:
I think this still has the same root issue in a slightly narrower form.

The PR removes the broad `...this.params` spread into `UploadPartCommand`, which fixes the reported `IfNoneMatch` failure. But the new allowlist still forwards top-level checksum fields from `Upload` params into every part upload.

I tested this locally with a temporary regression test using:

```ts
ChecksumSHA256: "object-level-sha256-base64=="
```

on the `Upload` params. The focused test fails because both mocked `UploadPartCommand` calls receive that same `ChecksumSHA256` value.

That looks like the same class of bug as `ContentMD5`: the value can be object-level at the `Upload` API boundary, but `UploadPartCommand` interprets checksum fields as per-part checksums. So the current allowlist can still create invalid multipart requests for users passing object-level checksum values.

Because of that, I don’t think the safe boundary is “fields accepted by `UploadPartCommand`.” Some fields are accepted by `UploadPartCommand`, but should only be passed if lib-storage computed them for that specific part. For user-supplied top-level params, the allowlist probably needs to exclude concrete `Checksum*` values too, not just `ContentMD5`.

Zelys-DFKH:
Good catch. You're right that the allowlist still leaks object-level checksums into UploadPart. `ChecksumSHA256` at the `Upload` boundary is a whole-object hash; sending it as a per-part value causes `BadDigest`. Same class of bug as `ContentMD5`.

Pushed `22fc2bc`: dropped the concrete `Checksum*` values from the allowlist, kept only `ChecksumAlgorithm`. That one is a directive (e.g. `"SHA256"`), not a precomputed hash, so the SDK middleware can still compute a per-part value from it. Extended the regression test to assert `ChecksumSHA256` doesn't reach `UploadPart`, plus a single-part PUT path test confirming object-level checksums are still preserved when multipart isn't used. 54/54 passing locally.

This complements #7990 (fail-fast on full-object `Checksum*` with multipart). Either this or your #8042 closes the bug; happy to defer to whichever the team prefers.

PR #8042, opened 2026-05-24
fix(lib-storage): avoid forwarding object-level params to UploadPart

### LINKED ISSUE #8020: Upload misses implementation for a few headers

danyalahmed1995:
### Issue
Fixes #8020

### Description
This narrows the params passed to `UploadPartCommand` in multipart `Upload` so object-level `Upload` params are not forwarded into every part upload.

The multipart path now avoids copying fields such as `IfNoneMatch`, `ContentType`, `ContentMD5`, and concrete user-supplied `Checksum*` values into `UploadPartCommand`.

`ChecksumAlgorithm` is still passed through so the SDK checksum middleware can compute checksums for the actual part body.

This also keeps the single-part `PutObjectCommand` path unchanged, so object-level params are still preserved when the upload does not use multipart.

Related to #8029. This follows the same `UploadPartCommand` boundary issue, but also excludes concrete user-supplied `Checksum*` values from multipart part requests.

### Testing
Added regression coverage for:
- single-part `PutObjectCommand` preserving object-level params
- multipart `UploadPartCommand` not receiving object-level headers/checksum values
- valid part-level fields still reaching `UploadPartCommand`
- `CompleteMultipartUploadCommand` still receiving `IfNoneMatch`

Ran:

```powershell
yarn exec vitest run lib/lib-storage/src/Upload.spec.ts
yarn exec vitest run -c lib/lib-storage/vitest.config.mts lib/lib-storage/src
```

Also verified locally with a custom request handler using the real `S3Client` serializer path. Multipart `UploadPart` requests no longer carried the object-level conditional/checksum headers, while the single `PutObject` path still preserved them.

### Checklist
  - It’s not a feature.
- [x] My E2E tests are resilient to concurrent I/O.
  - I didn’t write any E2E tests.
  - I didn’t add any public functions.
- [x] Streams - how do they work?? My WebStream readers/locks are properly lifecycled. Node.js stream backpressure is handled. Error handling.
  - No stream lifecycle changes.

### DIFF
--- a/lib/lib-storage/src/Upload.spec.ts
+++ b/lib/lib-storage/src/Upload.spec.ts
@@ -201,6 +201,29 @@ describe(Upload.name, () => {
     expect(PutObjectTaggingCommand).toHaveBeenCalledTimes(0);
   });
 
+  it("should preserve object-level params when uploading using PUT", async () => {
+    const actionParams = {
+      ...params,
+      IfNoneMatch: "*",
+      ContentType: "application/json",
+      ContentMD5: "object-level-md5-base64==",
+      ChecksumSHA256: "object-level-sha256-base64==",
+    };
+    const upload = new Upload({
+      params: actionParams,
+      client: new S3({}),
+    });
+
+    await upload.done();
+
+    expect(PutObjectCommand).toHaveBeenCalledTimes(1);
+    expect(PutObjectCommand).toHaveBeenCalledWith({
+      ...actionParams,
+      Body: Buffer.from(params.Body),
+    });
+    expect(UploadPartCommand).toHaveBeenCalledTimes(0);
+  });
+
   it("should upload using PUT when parts are smaller than one part stream", async () => {
     const streamBody = Readable.from(
       (function* () {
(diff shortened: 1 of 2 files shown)


danyalahmed1995:
@aBurmeseDev If the original issue is resolved should i close this PR or you would like to do it ?

danyalahmed1995:
Closing as the original issue #8020 was resolved.

PR #8044, opened 2026-05-25
ci: declare workflow-level contents: read on 2 lint workflows

arpitjain099:
Pins the default `GITHUB_TOKEN` to `contents: read` at the workflow level. The workflow runs checks only; no GitHub API writes.

Same post-CVE-2025-30066 (`tj-actions/changed-files`) supply-chain hardening pattern. YAML validated locally with `yaml.safe_load`.


arpitjain099:
This duplicates #8013 from earlier in the batch. Closing in favor of that one, sorry for the dup.

PR #8074, opened 2026-06-04
feat(lib-transfer-manager): add @aws-sdk/lib-transfer-manager as private package

smilkuri:
### Description
This is a work in progress high level library to interact with S3 which provides support for multipart upload using worker threads.

### Testing
```
$ yarn test:e2e

 Test Files  1 passed (1)
      Tests  26 passed (26)
   Start at  16:16:36
   Duration  277.44s (transform 188ms, setup 0ms, import 370ms, tests 276.94s, environment 0ms)
```

### Checklist
  - [ ] It's not a feature.
- [x] Streams - how do they work?? My WebStream readers/locks are properly lifecycled. Node.js stream backpressure is handled. Error handling.
  - [ ] No streams here.

---

### DIFF
--- a/lib/lib-transfer-manager/package.json
+++ b/lib/lib-transfer-manager/package.json
@@ -5,9 +5,35 @@
   "description": "S3 Transfer Manager for AWS SDK JS v3",
   "main": "./dist-cjs/index.js",
   "module": "./dist-es/index.js",
+  "browser": {
+    "./dist-es/runtimeConfig": "./dist-es/runtimeConfig.browser",
+    "./dist-es/s3-transfer-manager/join-streams": "./dist-es/s3-transfer-manager/join-streams.browser",
+    "./dist-es/s3-transfer-manager/worker-http-handler": "./dist-es/s3-transfer-manager/worker-http-handler.browser",
+    "fs": false
+  },
+  "react-native": {
+    "./dist-es/runtimeConfig": "./dist-es/runtimeConfig.native"
+  },
   "types": "./dist-types/index.d.ts",
+  "exports": {
+    ".": {
+      "types": "./dist-types/index.d.ts",
+      "module": "./dist-es/index.js",
+      "node": "./dist-cjs/index.js",
+      "import": "./dist-es/index.js",
+      "require": "./dist-cjs/index.js"
+    },
+    "./worker": {
+      "types": "./dist-types/submodules/worker/index.d.ts",
+      "module": "./dist-es/submodules/worker/index.js",
+      "node": "./dist-cjs/submodules/worker/index.js",
+      "import": "./dist-es/submodules/worker/index.js",
+      "require": "./dist-cjs/submodules/worker/index.js"
+    }
+  },
   "scripts": {
     "build": "concurrently 'yarn:build:types' 'yarn:build:es' && yarn build:cjs",
+    "lint": "node ../../scripts/validation/submodules-linter.js --pkg lib-transfer-manager",
     "build:cjs": "node ../../scripts/compilation/inline lib-transfer-manager",
     "build:es": "tsc -p tsconfig.es.json",
     "build:include:deps": "yarn g:turbo run build -F=\"$npm_package_name\"",
@@ -30,18 +56,16 @@
   },
   "license": "Apache-2.0",
   "dependencies": {
-    "@aws-sdk/config": "workspace:3.1062.0",
-    "@aws-sdk/lib-storage": "workspace:3.1062.0",
     "@smithy/core": "^3.24.6",
     "@smithy/node-http-handler": "^4.7.6",
     "@smithy/types": "^4.14.3",
     "tslib": "^2.6.2"
   },
   "peerDependencies": {
-    "@aws-sdk/client-s3": "^3.0.0"
+    "@aws-sdk/client-s3": "workspace:^3.1063.0"
   },
   "devDependencies": {
-    "@aws-sdk/client-s3": "workspace:3.1062.0",
+    "@aws-sdk/client-s3": "workspace:3.1063.0",
     "@tsconfig/recommended": "1.0.1",
     "@types/node": "^20.14.8",
     "concurrently": "7.0.0",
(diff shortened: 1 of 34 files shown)


lib/lib-transfer-manager/package.json
```diff
@@ -0,0 +1,81 @@
...
+  },
+  "browser": {
+    "./dist-es/runtimeConfig": "./dist-es/runtimeConfig.browser",
+    "./dist-es/s3-transfer-manager/join-streams": "./dist-es/s3-transfer-manager/join-streams.browser",
+    "./dist-es/s3-transfer-manager/worker-http-handler": "./dist-es/s3-transfer-manager/worker-http-handler.browser",
+    "fs": false,
+    "stream": "stream-browserify"
+  },
+  "standaloneFiles": [
+    "s3-transfer-manager/http-request-worker.js"
```
kuhe: why is this needed?

lib/lib-transfer-manager/package.json
```diff
@@ -0,0 +1,81 @@
...
+    "node": ">=20.0.0"
+  },
+  "author": {
+    "name": "AWS SDK for JavaScript Team",
+    "url": "https://aws.amazon.com/javascript/"
+  },
+  "license": "Apache-2.0",
+  "dependencies": {
+    "@aws-sdk/config": "workspace:3.1062.0",
+    "@aws-sdk/lib-storage": "workspace:3.1062.0",
```
kuhe: don't take a dependency on another HLL.

copy the function for now, move it into an upstream package later

lib/lib-transfer-manager/package.json
```diff
@@ -0,0 +1,81 @@
...
+  "engines": {
+    "node": ">=20.0.0"
+  },
+  "author": {
+    "name": "AWS SDK for JavaScript Team",
+    "url": "https://aws.amazon.com/javascript/"
+  },
+  "license": "Apache-2.0",
+  "dependencies": {
+    "@aws-sdk/config": "workspace:3.1062.0",
```
kuhe: don't take a direct dep on config, use the internal package directly

lib/lib-transfer-manager/package.json
```diff
@@ -0,0 +1,81 @@
...
+    "@aws-sdk/lib-storage": "workspace:3.1062.0",
+    "@smithy/core": "^3.24.6",
+    "@smithy/node-http-handler": "^4.7.6",
+    "@smithy/types": "^4.14.3",
+    "tslib": "^2.6.2"
+  },
+  "peerDependencies": {
+    "@aws-sdk/client-s3": "^3.0.0"
```
kuhe: "@aws-sdk/client-s3": "workspace:^3.1062.0",

peer must be at minimum the tested version. It simply doesn't work with v3.0.0

lib/lib-transfer-manager/package.json
```diff
@@ -0,0 +1,81 @@
...
+        "dist-types/ts3.4/*"
+      ]
+    }
+  },
+  "browser": {
+    "./dist-es/runtimeConfig": "./dist-es/runtimeConfig.browser",
+    "./dist-es/s3-transfer-manager/join-streams": "./dist-es/s3-transfer-manager/join-streams.browser",
+    "./dist-es/s3-transfer-manager/worker-http-handler": "./dist-es/s3-transfer-manager/worker-http-handler.browser",
+    "fs": false,
+    "stream": "stream-browserify"
```
kuhe: don't use this, just use ReadableStream in browsers

lib/lib-transfer-manager/package.json
```diff
@@ -0,0 +1,81 @@
...
+    "<4.0": {
+      "dist-types/*": [
+        "dist-types/ts3.4/*"
+      ]
+    }
+  },
+  "browser": {
+    "./dist-es/runtimeConfig": "./dist-es/runtimeConfig.browser",
+    "./dist-es/s3-transfer-manager/join-streams": "./dist-es/s3-transfer-manager/join-streams.browser",
+    "./dist-es/s3-transfer-manager/worker-http-handler": "./dist-es/s3-transfer-manager/worker-http-handler.browser",
```
kuhe: rather than file level replacements, create a separate index.browser.ts file.

```
main: dist-cjs/index.js
browser: dist-es/index.browser.js
module: dist-es/index.browser.js
react-native: dist-es/index.native.js
```

scripts/compilation/Inliner.js
```diff
@@ -22,6 +22,7 @@ module.exports = class Inliner {
     this.isPackage = !this.isInternalPackage && fs.existsSync(path.join(root, "packages", pkg));
     this.isLib = fs.existsSync(path.join(root, "lib", pkg));
     this.reExportStubs = false;
+    this.standaloneFiles = [];
```
kuhe: no changes should be needed in this file

scripts/validation/api.json
```diff
@@ -533,13 +533,19 @@
   },
   "@aws-sdk/lib-storage": {
     "BodyDataTypes": "type(union)",
+    "byteLength": "function",
     "Configuration": "type(interface)",
+    "getChunk": "function",
```
kuhe: no API additions to lib-storage should be made while we develop this new lib. Copy them over.

lib/lib-transfer-manager/package.json
smilkuri: To keep this file as is and not to inline it into `dist-cjs/index.js` during bundling. Because this `http-request-worker` runs inside the worker thread and [Worker](https://github.com/aws/aws-sdk-js-v3/pull/8074/changes#diff-5398d198c7d947f5cea871110292ec95779c3f767e3eff38ba70bae788344f68R201) requires a [file](https://github.com/aws/aws-sdk-js-v3/pull/8074/changes#diff-5398d198c7d947f5cea871110292ec95779c3f767e3eff38ba70bae788344f68R172-R181) path. 
https://nodejs.org/api/worker_threads.html#new-workerfilename-options

kuhe: make the package a submodule carrier, and move the request worker file into its own submodule.

this way the file will exist at `dist-cjs/submodules/worker/index.js` and be separate from the main bundle at `dist-cjs/index.js`.

To do so you need to create an `exports` field in package.json and create the `src/submodules/worker` folder. Submodule linter will take care of the rest, but `scripts/validation/submodules-linter.js` needs to be updated to look for submodules in `lib` too.

For this package, use 

```
"lint": "node ../../scripts/validation/submodules-linter.js --pkg lib-transfer-manager",
```

in package.json scripts

smilkuri: Okay thanks. Will do that.

lib/lib-transfer-manager/package.json
smilkuri: For this we need to duplicate `S3TransferManager.ts` into `S3TransferManager.browser.ts` and replace join-streams and worker-http-handler browser variants in the imports. Is it worth duplicating the code?

kuhe: use dependency injection binding like in https://github.com/smithy-lang/smithy-typescript/blob/main/packages/core/src/submodules/serde/index.browser.ts

```ts
export class Uint8ArrayBlobAdapter extends bindUint8ArrayBlobAdapter(toUtf8, fromUtf8, toBase64, fromBase64) {}
```

lib/lib-transfer-manager/src/submodules/worker/index.ts
```diff
@@ -0,0 +1,219 @@
+/**
```
kuhe: no impl code in index, export from another file

lib/lib-transfer-manager/package.json
kuhe: we can do this in a later PR

kuhe (approved):

PR #8086, opened 2026-06-08
fix(core): export legacy root symbols from submodules instead of relative paths

kuhe:
### Issue
V2236964587

### Description
This is a fix for older clients that import from `@aws-sdk/core` root level. The root should export submodules by name so they can re-resolve through the exports mappings rather than directly file-reference a submodule index. 

Using relative paths prevents the conditional browser submodule exports from working.

### Testing
CI

### Checklist
  - [x] It's not a feature.
  - [x] No streams here.

---

### DIFF
--- a/packages-internal/core/src/index.ts
+++ b/packages-internal/core/src/index.ts
@@ -11,7 +11,6 @@
  * No additional exports are allowed.
  */
 
-// --- @aws-sdk/core/client (legacy subset) ---
 export {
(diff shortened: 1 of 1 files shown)


kuhe:
rebase, restart CI after https://github.com/aws/aws-sdk-js-v3/pull/8087

trivikr (approved):

PR #8087, opened 2026-06-08
chore(lib-transfer-manager): sort scripts alphabetically

trivikr:
### Issue
https://github.com/aws/aws-sdk-js-v3/actions/runs/27156466760/job/80160519079

### Description
Sort scripts alphabetically for lib-transfer-manager.
This was missed as https://github.com/aws/aws-sdk-js-v3/pull/8074 wasn't rebased before merging

### Testing
CI

### Checklist
- [ ] If the PR is a feature, add integration tests (`*.integ.spec.ts`) or E2E tests.
  - [x] It's not a feature.
- [ ] My E2E tests are resilient to concurrent i/o.
  - [x] I didn't write any E2E tests.
- [ ] I added access level annotations e.g. `@public`, `@internal` tags and enabled doc generation on the package. Remember that access level annotations go below the description, not above.
  - [x] I didn't add any public functions.
- [ ] Streams - how do they work?? My WebStream readers/locks are properly lifecycled. Node.js stream backpressure is handled. Error handling.
  - [x] No streams here.

---
By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #8088, opened 2026-06-08
test(middleware-sdk-s3): use crypto.randomUUID for E2E test bucket names

trivikr:
### Issue
Internal JS-6960

### Description

Replace 2-character random ID with crypto.randomUUID to avoid
BucketAlreadyExists errors from S3's global bucket namespace collisions
during concurrent CI runs. Also remove unnecessary STS dependency,
redundant deleteBucket pre-cleanup, and headBucket sanity check.

### Testing
CI

### Checklist
  - [x] It's not a feature.
  - [x] No streams here.

---

### DIFF
--- a/packages-internal/middleware-sdk-s3/src/submodules/s3/middleware-region-redirect/region-redirect-middleware.e2e.spec.ts
+++ b/packages-internal/middleware-sdk-s3/src/submodules/s3/middleware-region-redirect/region-redirect-middleware.e2e.spec.ts
@@ -1,7 +1,5 @@
 import type { S3ClientConfig } from "@aws-sdk/client-s3";
-import { S3, waitUntilBucketExists, waitUntilBucketNotExists } from "@aws-sdk/client-s3";
-import type { GetCallerIdentityCommandOutput } from "@aws-sdk/client-sts";
-import { STS } from "@aws-sdk/client-sts";
+import { S3, waitUntilBucketExists } from "@aws-sdk/client-s3";
 import { afterAll, beforeAll, describe, expect, onTestFailed, test as it } from "vitest";
 
 const testValue = "Hello S3 global client!";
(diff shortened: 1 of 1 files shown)


trivikr:
This will be marked ready for review post rebase once precommit fix in https://github.com/aws/aws-sdk-js-v3/pull/8087 is merged.

packages-internal/middleware-sdk-s3/src/submodules/s3/middleware-region-redirect/region-redirect-middleware.e2e.spec.ts
```diff
@@ -30,32 +28,18 @@ describe("S3 Global Client Test", () => {
...
-    callerID = await stsClient.getCallerIdentity({});
-    bucketNames = regionConfigs.map((config) => `${callerID.Account}-${randId}-redirect-${config.region}`);
-    await Promise.all(
-      bucketNames.map(async (bucketName, index) => {
-        await deleteBucket(s3Clients[index], bucketName);
-        return waitUntilBucketNotExists({ client: s3Clients[index], maxWaitTime: 60 }, { Bucket: bucketName });
-      })
-    );
+    const uniqueId = crypto.randomUUID();
+    bucketNames = regionConfigs.map((config) => `redirect-${config.region}-${uniqueId}`);
```
kuhe, line 36: use this new S3 feature for the general purpose bucket(s)?

https://docs.aws.amazon.com/AmazonS3/latest/userguide/gpbucketnamespaces.html

kuhe (approved):

packages-internal/middleware-sdk-s3/src/submodules/s3/middleware-region-redirect/region-redirect-middleware.e2e.spec.ts
trivikr: The account regional namespace will avoid collision with other accounts, but wouldn't avoid collision between E2E tests being in parallel. I'll prefer to keep the UUID solution as it fixes collision and also avoids STS import (required for account ID).

kuhe: UUID is sufficiently unique

PR #8092, opened 2026-06-09
docs: suggest UndiciHttpHandler as an alternative request handler

trivikr:
### Issue
N/A

### Description

Document @smithy/undici-http-handler in CLIENTS.md with usage example
showing dispatcher configuration options. Add a note in the parallel
workloads guide pointing to it as a higher-throughput alternative to
the default node:https handler. Share link in performance README too.

### Testing
N/A

### Checklist
- [ ] If the PR is a feature, add integration tests (`*.integ.spec.ts`) or E2E tests.
  - [x] It's not a feature.
- [ ] My E2E tests are resilient to concurrent i/o.
  - [x] I didn't write any E2E tests.
- [ ] I added access level annotations e.g. `@public`, `@internal` tags and enabled doc generation on the package. Remember that access level annotations go below the description, not above.
  - [x] I didn't add any public functions.
- [ ] Streams - how do they work?? My WebStream readers/locks are properly lifecycled. Node.js stream backpressure is handled. Error handling.
  - [x] No streams here.

---
By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #8093, opened 2026-06-09
feat(undici-http-handler): re-export '@smithy/undici-http-handler'

trivikr:
### Issue
Refs: https://github.com/aws/aws-sdk-js-v3/pull/8092#discussion_r3382550195

### Description
Re-export '@smithy/undici-http-handler' as '@aws-sdk/undici-http-handler'

### Testing
CI

### Checklist
  - [x] It's not a feature.
  - [x] No streams here.

---

### DIFF
--- a/packages-internal/undici-http-handler/package.json
+++ b/packages-internal/undici-http-handler/package.json
@@ -11,7 +11,9 @@
     "build:es": "tsc -p tsconfig.es.json",
     "build:types": "tsc -p tsconfig.types.json",
     "build:types:downlevel": "downlevel-dts dist-types dist-types/ts3.4",
-    "clean": "premove dist-cjs dist-es dist-types tsconfig.cjs.tsbuildinfo tsconfig.es.tsbuildinfo tsconfig.types.tsbuildinfo"
+    "clean": "premove dist-cjs dist-es dist-types tsconfig.cjs.tsbuildinfo tsconfig.es.tsbuildinfo tsconfig.types.tsbuildinfo",
+    "test": "yarn g:vitest run",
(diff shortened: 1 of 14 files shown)


packages-internal/undici-http-handler/package.json
```diff
@@ -0,0 +1,56 @@
+{
+  "name": "@aws-sdk/undici-http-handler",
+  "version": "2.0.0",
```
kuhe, line 3: is this intended to be imported directly by customers? 

if so, it should go in `/packages` (public-facing), and it will have the 3.x version scheme. 

If you want it specifically to use a non-3.x versioning scheme, then it can live in `packages-internal`, but still be referred to under this name in documentation.

kuhe: it looks like the plan is to match the major version of `@smithy/undici-http-handler`, but have separate minor/patch versions

kuhe (approved):

packages-internal/undici-http-handler/package.json
trivikr: Yes. Since UndiciHttpHandler is tightly coupled with undici, we follow it's node version support.
And we've to release major versions of UndiciHttpHandler whenever undici updates their node version support in its major version

trivikr: Since `@aws-sdk/undici-http-handler` is just a re-export which uses the latest from specific major version of `@smithy/undici-http-handler`, we're likely just going to bump it only when new major version of latter is published. We can consider bumping new versions of former if there's a critical bug fix in latter.

samuellagrene90-cpu (approved):

PR #8102, opened 2026-06-15
test(client-iam): increase hook timeout to 240s

smilkuri:
### Issue
Internal P452562039

### Description
increase hook timeout to 240s

### Testing
CI

### Checklist
  - [x] It's not a feature.
  - [x] No streams here.

---

### DIFF
--- a/clients/client-iam/test/iam-features.e2e.spec.ts
+++ b/clients/client-iam/test/iam-features.e2e.spec.ts
@@ -28,7 +28,7 @@ describe(IAM.name, () => {
         console.warn(`Failed to delete role ${roleName}:`, error);
       }
     }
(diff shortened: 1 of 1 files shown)


clients/client-iam/test/iam-features.e2e.spec.ts
```diff
@@ -28,7 +28,7 @@ describe(IAM.name, () => {
         console.warn(`Failed to delete role ${roleName}:`, error);
       }
     }
-  }, 120_000);
+  }, 240_000);
```
kuhe, line 31: we're not creating a lot of roles/users. The root cause is not likely to be an insufficient timeout.

PR #8121, opened 2026-06-23
feat(lib-transfer-manager): add uploadDirectory functionality

smilkuri:
### Issue
Internal JS-6110

### Description
Adds `uploadDirectory` functionality to S3TransferManager for uploading files from a local directory to an S3 bucket.

### Testing
```
yarn test:e2e

Test Files  1 passed (1)
      Tests  37 passed (37)
   Start at  18:36:24
   Duration  273.35s (transform 256ms, setup 0ms, import 448ms, tests 272.77s, environment 0ms)
```

### Checklist
  - [x] It's not a feature.
  - [x] No streams here.

---

### DIFF
--- a/lib/lib-transfer-manager/src/submodules/transfer-manager/S3TransferManager.ts
+++ b/lib/lib-transfer-manager/src/submodules/transfer-manager/S3TransferManager.ts
@@ -520,10 +520,14 @@ export class S3TransferManager implements IS3TransferManager {
       discoveredFiles++;
       discoveredBytes += fileStat.size;
 
-      await semaphore.acquire();
+      const count = fileStat.size >= this.multipartUploadThresholdBytes
+        ? Math.min(Math.ceil(fileStat.size / this.targetPartSizeBytes), this.maxConcurrentUploads)
+        : 1;
+
+      await semaphore.acquire(count);
 
       if (terminated || abortController.signal.aborted) {
-        semaphore.release();
+        semaphore.release(count);
         break;
       }
 
@@ -577,7 +581,7 @@ export class S3TransferManager implements IS3TransferManager {
           }
           objectsFailed++;
         } finally {
-          semaphore.release();
+          semaphore.release(count);
         }
       })();
 
--- a/lib/lib-transfer-manager/src/submodules/transfer-manager/S3TransferManager.ts
+++ b/lib/lib-transfer-manager/src/submodules/transfer-manager/S3TransferManager.ts
@@ -478,6 +478,11 @@ export class S3TransferManager implements IS3TransferManager {
 
     this.checkAborted(transferOptions);
 
+    const userSignal = transferOptions?.abortSignal as AbortSignal | undefined;
+    if (userSignal) {
+      userSignal.addEventListener("abort", () => abortController.abort(), { once: true });
+    }
+
     for await (const filePath of this.traverseDirectory(absoluteSource, {
       recursive: request.recursive ?? false,
       followSymbolicLinks: request.followSymbolicLinks ?? false,
@@ -510,9 +515,13 @@ export class S3TransferManager implements IS3TransferManager {
             putRequest = request.uploadObjectRequestModifier(putRequest);
           }
 
-          await this.upload(putRequest, transferOptions);
+          await this.upload(putRequest, { ...transferOptions, abortSignal: abortController.signal });
           objectsUploaded++;
         } catch (error) {
+          if (terminated || abortController.signal.aborted) {
+            objectsFailed++;
+            return;
+          }
           const context: DirectoryTransferFailureContext = {
             request,
             objectRequest: { Bucket: request.bucket, Key: this.deriveS3Key(absoluteSource, filePath, request.s3Prefix) },
@@ -521,10 +530,11 @@ export class S3TransferManager implements IS3TransferManager {
           const result = await handleFailure(failurePolicy, context, abortController);
           if (result === "terminate") {
             terminated = true;
-            terminationError = error;
-          } else {
-            objectsFailed++;
+            if (!terminationError) {
+              terminationError = error;
+            }
           }
+          objectsFailed++;
         } finally {
           semaphore.release();
         }
--- a/lib/lib-transfer-manager/src/submodules/transfer-manager/S3TransferManager.ts
+++ b/lib/lib-transfer-manager/src/submodules/transfer-manager/S3TransferManager.ts
@@ -477,11 +477,32 @@ export class S3TransferManager implements IS3TransferManager {
     let terminationError: unknown;
 
     this.checkAborted(transferOptions);
+    this.addEventListeners(transferOptions?.eventListeners);
+
+    let transferredBytes = 0;
+    let transferredFiles = 0;
+    let totalFiles: number | undefined = undefined;
+    let totalBytes: number | undefined = undefined;
+    let traversalComplete = false;
+
+    const removeLocalEventListeners = () => {
+      this.removeEventListeners(transferOptions?.eventListeners);
+    };
+
+    this.dispatchEvent(
+      Object.assign(new Event("transferInitiated"), {
+        request,
+        snapshot: { transferredBytes: 0, totalBytes: undefined, transferredFiles: 0, totalFiles: undefined },
+      })
+    );
 
     const userSignal = transferOptions?.abortSignal as AbortSignal | undefined;
     if (userSignal) {
       userSignal.addEventListener("abort", () => abortController.abort(), { once: true });
     }
+
+    let discoveredFiles = 0;
+    let discoveredBytes = 0;
 
     for await (const filePath of this.traverseDirectory(absoluteSource, {
       recursive: request.recursive ?? false,
@@ -492,6 +513,10 @@ export class S3TransferManager implements IS3TransferManager {
 
       if (request.filter && !request.filter(filePath)) continue;
 
+      const fileStat = await stat(filePath);
+      discoveredFiles++;
+      discoveredBytes += fileStat.size;
+
       await semaphore.acquire();
 
       if (terminated || abortController.signal.aborted) {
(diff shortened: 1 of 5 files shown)


lib/lib-transfer-manager/src/submodules/transfer-manager/directory-transfer-utils.ts
```diff
@@ -0,0 +1,93 @@
...
+ * Simple counting semaphore for bounding concurrency.
+ * @internal
+ */
+export class Semaphore {
+  private slots: number;
+  private queue: Array<() => void> = [];
+
+  constructor(slots: number) {
+    this.slots = slots;
+  }
```
kuhe: can be simplified to

```ts
public constructor(private slots: number) {}
```

lib/lib-transfer-manager/src/submodules/transfer-manager/directory-transfer-utils.ts
```diff
@@ -0,0 +1,93 @@
...
+
+  async acquire(): Promise<void> {
+    if (this.slots > 0) {
+      this.slots--;
+      return;
+    }
+    return new Promise((resolve) => this.queue.push(resolve));
+  }
+
+  release(): void {
```
kuhe: add access level words

lib/lib-transfer-manager/src/submodules/transfer-manager/directory-transfer-utils.ts
```diff
@@ -0,0 +1,93 @@
...
+ *
+ * @internal
+ */
+export async function validateDirectory(dirPath: string): Promise<string> {
+  const absolutePath = resolve(dirPath);
+  let stats: Stats;
+  try {
+    stats = await stat(absolutePath);
+  } catch (error: any) {
+    throw new Error(`Directory does not exist at: ${absolutePath}`);
```
kuhe: what errors does `stat` throw? 

do they all imply the directory doesn't exist?

lib/lib-transfer-manager/src/submodules/transfer-manager/types.ts
```diff
@@ -363,3 +322,179 @@ export interface JoinStreamIterationEvents {
...
+/**
+ * Failure policy for directory transfers.
+ * Either a canned policy enum value, or a custom callback invoked per failure.
+ * Custom callbacks return "continue" to skip the failure, or "terminate" to abort.
+ *
+ * @alpha
+ */
+export type FailurePolicy =
+  | CannedFailurePolicy
+  | ((context: DirectoryTransferFailureContext) => Promise<"continue" | "terminate">);
```
kuhe: `Promise<CannedFailurePolicy>`?

lib/lib-transfer-manager/src/submodules/transfer-manager/types.ts
```diff
@@ -363,3 +322,179 @@ export interface JoinStreamIterationEvents {
...
   onStreamConsumed?: (index: number) => void;
 }
+
+/**
+ * Canned failure policies for directory transfers.
+ * Used to control behavior when an individual object fails during a multi-object transfer.
+ *
+ * @alpha
+ */
+export enum CannedFailurePolicy {
```
kuhe: we mostly don't use TS enums. Use a string union enum?

lib/lib-transfer-manager/src/submodules/transfer-manager/types.ts
```diff
@@ -363,3 +322,179 @@ export interface JoinStreamIterationEvents {
...
+  /**
+   * Whether to upload directories recursively.
+   * Default: false.
+   */
+  recursive?: boolean;
+  /**
+   * S3 key prefix prepended to each object key.
+   * By default, if no prefix is specified, objects are uploaded to the root of the bucket.
+   */
+  s3Prefix?: string;
```
kuhe, line 397: why does this one property get named with "s3" as a prefix?

everything in here is S3. This is more like `keyPrefix`.

lib/lib-transfer-manager/src/submodules/transfer-manager/types.ts
```diff
@@ -363,3 +322,179 @@ export interface JoinStreamIterationEvents {
...
+  /**
+   * A callback to allow users to filter out unwanted files.
+   * Return true to include the file for upload, false to skip.
+   * By default, if no filter is specified, all files will be uploaded.
+   */
+  filter?: (filePath: string) => boolean;
```
kuhe: also accept a regex?

lib/lib-transfer-manager/src/submodules/transfer-manager/types.ts
```diff
@@ -363,3 +322,179 @@ export interface JoinStreamIterationEvents {
...
+  /**
+   * Failure policy for handling individual upload failures.
+   * Default: Terminate.
+   */
+  failurePolicy?: FailurePolicy;
+  /**
+   * Max concurrent file uploads for this directory operation.
+   * Default: 100.
+   */
+  maxConcurrency?: number;
```
kuhe, line 419: why is this 100? the default MPU doesn't even have 100 concurrency

lib/lib-transfer-manager/src/submodules/transfer-manager/directory-transfer-utils.ts
smilkuri: No, it throws other errors too like permission denied, path is not a directory.... will make the error message generic.

lib/lib-transfer-manager/src/submodules/transfer-manager/types.ts
smilkuri: Named it according to the specification.

lib/lib-transfer-manager/src/submodules/transfer-manager/types.ts
smilkuri: That was the default specified in the SEP. For file level concurreny, not part-level.But for directories with many large files the memory could be high.

lib/lib-transfer-manager/src/submodules/transfer-manager/types.ts
kuhe: refer to SEP as "spec" or specification in public

lib/lib-transfer-manager/src/submodules/transfer-manager/types.ts
kuhe: for small files this will call PutObject with file stream input rather than MPU, right?

The memory usage in that case is mostly stream chunk buffers? I think if MPU is triggered by a large file in the directory, it should count more than 1 against the concurrency limit.

lib/lib-transfer-manager/src/submodules/transfer-manager/directory-transfer-utils.ts
```diff
@@ -0,0 +1,90 @@
...
+import { resolve } from "node:path";
+
+import type { CannedFailurePolicy, DirectoryTransferFailureContext, FailurePolicy } from "./types";
+
+/**
+ * Simple counting semaphore for bounding concurrency.
+ * @internal
+ */
+export class Semaphore {
+  private queue: Array<{ weight: number; resolve: () => void }> = [];
```
smilkuri: Problem I can think of here with FIFO based ordering is if a file needs more slots than available, it will also block the requests behind it which might need a fewer slots. But if we let smaller files go first, it might lead to large files waiting indefinitely.

smilkuri: Updated the semaphore to allow a request to proceed as long as inFlightRequest count is less than concurrency limit.

lib/lib-transfer-manager/src/submodules/transfer-manager/S3TransferManager.ts
```diff
@@ -1522,6 +1661,60 @@ export class S3TransferManager implements IS3TransferManager {
     }
   }
 
+  /**
+   * Async generator that yields file paths from the source directory.
+   * Uses fs.opendir with recursive option (Node 20+).
+   * Handles symlink cycle detection when followSymbolicLinks is true.
+   *
+   * @internal
```
kuhe, line 1669: minor issue - private methods don't need the `@internal` tag.

lib/lib-transfer-manager/src/submodules/transfer-manager/S3TransferManager.ts
```diff
@@ -444,28 +453,168 @@ export class S3TransferManager implements IS3TransferManager {
...
-    recursive?: boolean;
-    s3Prefix?: string;
-    filter?: (filepath: string) => boolean;
-    s3Delimiter?: string;
-    putObjectRequestCallback?: (putObjectRequest: PutObjectCommandInput) => Promise<void>;
-    failurePolicy?: (error?: unknown) => Promise<void>;
-    transferOptions?: TransferOptions;
-  }): Promise<{ objectsUploaded: number; objectsFailed: number }> {
-    throw new Error("Method not implemented.");
+  public async uploadDirectory(request: UploadDirectoryRequest, transferOptions?: TransferOptions): Promise<UploadDirectoryResponse> {
```
kuhe, line 466: this being a method makes TM inherently unusable in browsers, right?

perhaps this method should dispatch to a injectable implementation which throws 'not implemented' in browsers?

kuhe (approved):

lib/lib-transfer-manager/src/submodules/transfer-manager/S3TransferManager.ts
smilkuri: Yes, currently `upload` (`threadUploadInParts`, `threadedUploadInPartsFromFile` paths) and `uploadAll ` doesn't have browser support, considering the less no of customers we have that use the SDK in browsers. But will add it later.

PR #8146, opened 2026-07-02
fix(util-dynamodb): null empty ArrayBuffer and DataView with convertEmptyValues

spokodev:
With `marshallOptions.convertEmptyValues` set to true, an empty binary value is
meant to be converted to a NULL attribute. This works for a typed array such as
`Uint8Array`, but not for an empty `ArrayBuffer` or `DataView`.

`convertToAttr` checks `data.length === 0` inside the binary branch, but
`ArrayBuffer` and `DataView` do not have a `length` property (they use
`byteLength`), so `data.length` is `undefined`, the check is false, and the empty
value is marshalled as `{ B: ... }` instead of `{ NULL: true }`. `isBinary`
already recognizes `ArrayBuffer` and `DataView` as binary, so both reach this
branch.

The fix uses `byteLength`, which is defined for every binary type including typed
arrays, so the empty check is correct for all of them. Added a test for an empty
`ArrayBuffer` and `DataView`.

### DIFF
--- a/packages-internal/util-dynamodb/src/convertToAttr.spec.ts
+++ b/packages-internal/util-dynamodb/src/convertToAttr.spec.ts
@@ -162,6 +162,17 @@ describe("convertToAttr", () => {
           NULL: true,
         });
       });
+
+      it("returns null for an empty ArrayBuffer or DataView when options.convertEmptyValues=true", () => {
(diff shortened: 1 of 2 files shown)


kuhe:
I've reported this issue, and we will apply a different fix https://github.com/aws/aws-sdk-js-v3/issues/8154

PR #8147, opened 2026-07-02
feat: introducing w3c tracing headers propagation

darklight3it:
### Issue

N/A — W3C trace context propagation

### Description

Adds a build-step middleware that propagates W3C trace context headers (`traceparent`, `tracestate`, `baggage`) from the Lambda `InvokeStore` onto outbound AWS SDK requests.

**Why:** When a Lambda function receives a request with trace context, downstream calls made via the AWS SDK should carry that context forward. This enables distributed tracing across service boundaries on the AWS SDK.

**How it works:**
- The middleware reads trace context from `@aws/lambda-invoke-store` (which the Lambda runtime populates per-invocation from inbound headers)
- If `traceparent` is present and the outgoing request doesn't already have one, it injects `traceparent`, `tracestate`, and `baggage` headers
- Existing trace headers are respected (never overwritten) and normalized to lowercase
- Browser and React Native get no-op implementations (InvokeStore is Node-only)

**Design decisions:**
- Registered as a plugin (`getTraceContextPropagationPlugin`) — not auto-added to generated clients.
- `traceparent` is the gate: `tracestate` and `baggage` are only propagated when `traceparent` is present (per W3C spec).
- `InvokeStore.getInstanceAsync()` is memoized at module level — subsequent calls on the hot path just await a resolved promise.

### Testing

- 7 unit tests covering: full propagation, partial headers, missing traceparent, no InvokeStore context, existing headers preserved, case normalization, non-HTTP request no-op
- 1 integration test exercising a real S3 client middleware stack with `requireRequestsFrom` (no network calls)
- All tests pass locally. Fork CI cannot authenticate with upstream's AWS OIDC role (expected for fork PRs).

### Checklist
  - [ ] It's not a feature.
  - [x] No streams here.

---

### DIFF
--- a/packages-internal/core/src/submodules/client/index.browser.ts
+++ b/packages-internal/core/src/submodules/client/index.browser.ts
@@ -25,6 +25,11 @@ export { recursionDetectionMiddlewareOptions } from "./middleware-recursion-dete
 export { getRecursionDetectionPlugin } from "./middleware-recursion-detection/getRecursionDetectionPlugin.browser";
 export { recursionDetectionMiddleware } from "./middleware-recursion-detection/recursionDetectionMiddleware.browser";
 
+// middleware-trace-context-propagation
+export { traceContextPropagationMiddlewareOptions } from "./middleware-trace-context-propagation/configuration";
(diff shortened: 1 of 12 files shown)


kuhe:
moved to https://github.com/aws/aws-sdk-js-v3/pull/8156

PR #8152, opened 2026-07-05
fix(util-user-agent-node): use direct process.env access for bundler tree-shaking

### LINKED ISSUE #8149: Unable to remove TypeScript package checks for user-agent telemetry during bundling

### Describe the bug

Since our lambda code is being bundled before use, the checks done in https://github.com/aws/aws-sdk-js-v3/blob/74dc940b3bc837ef038d97b43b05a2afa164e5cc/packages-internal/core/src/submodules/client/util-user-agent-node/getTypeScriptUserAgentPair.ts are kind of useless, as no `node_modules` folder exists.

However, there's no way for me to remove this code during the bundler step, as it accesses `process.env` in a non-typical fashion:

https://github.com/aws/aws-sdk-js-v3/blob/74dc940b3bc837ef038d97b43b05a2afa164e5cc/packages-internal/core/src/submodules/client/util-user-agent-node/getTypeScriptUserAgentPair.ts#L30

If the code would use regular `process.env.ENV_VARIABLE_HERE`, its value could be inline during the bundling process.

### Regression Issue

### SDK version number

3.972.8

Node.js

Node v24.15

### Reproduction Steps

1. Use a bundler like esbuid/rolldown/rollup
2. e.g. for rolldown, add `process.env.AWS_SDK_JS_TYPESCRIPT_DETECTION_DISABLED` to `transform.define`
3. Bundle the lambda
4. The obsolete code has not been removed.

### Observed Behavior

Nothing got removed.

### Expected Behavior

The bundler should see that the value is always truthy, and as such, can remove the rest of the function (which would also remove all extra imported functions, etc..)

### Possible Solution

Replace the call with something like:
```ts
process.env.AWS_SDK_JS_TYPESCRIPT_DETECTION_DISABLED == true || process.env.AWS_SDK_JS_TYPESCRIPT_DETECTION_DISABLED == 1
```

_No response_

tsushanth:
## Summary

Closes #8149

`getTypeScriptUserAgentPair` calls `booleanSelector(process.env, "AWS_SDK_JS_TYPESCRIPT_DETECTION_DISABLED", SelectorType.ENV)` to read the disable flag. Because `booleanSelector` receives `process.env` as a dynamic argument, bundlers (esbuild, rolldown, rollup, etc.) cannot statically resolve the value and cannot tree-shake the entire TypeScript-detection block even when the env var is defined in the bundler's `define` / `transform.define` config.

**Root cause:** indirect property access via `booleanSelector(record, key)` is opaque to bundlers. Only a literal `process.env.KEY` access can be inlined at bundle time.

**Fix:** replace the `booleanSelector` call with a direct `process.env.AWS_SDK_JS_TYPESCRIPT_DETECTION_DISABLED` property read and inline boolean coercion (`=== "true" || === "1"`). This preserves the existing semantics exactly:

| env value | before | after |
|-----------|--------|-------|
| `"true"` | skip TS detection | skip TS detection |
| `"1"` | skip TS detection | skip TS detection |
| `"false"` / `"0"` / unset | proceed | proceed |
| unrecognized value | `booleanSelector` throws → caught → proceed | `=== "true"` false → proceed |

The `@smithy/core/config` import is now unused in this file and has been removed.

## Test changes

- Removed `vi.mock("@smithy/core/config")` (no longer needed)
- Replaced `vi.mocked(booleanSelector).mockReturnValue(...)` stubs with direct `process.env[ENV_VAR]` assignment/deletion
- Added explicit test for `"1"` truthy value and `"false"` non-truthy value

## Verification (manual)

With esbuild `define: { "process.env.AWS_SDK_JS_TYPESCRIPT_DETECTION_DISABLED": JSON.stringify("true") }` the bundled output now contains zero references to `getTypeScriptUserAgentPair`, `getNodeModulesParentDirs`, `getSanitizedTypeScriptVersion`, etc.

### DIFF
--- a/packages-internal/core/src/submodules/client/util-user-agent-node/getTypeScriptUserAgentPair.ts
+++ b/packages-internal/core/src/submodules/client/util-user-agent-node/getTypeScriptUserAgentPair.ts
@@ -25,7 +25,7 @@ export const getTypeScriptUserAgentPair = async (): Promise<UserAgentPair | unde
   // Disable TypeScript detection if environment variable is set.
   // Direct property access (rather than booleanSelector) allows bundlers to statically replace
   // process.env.AWS_SDK_JS_TYPESCRIPT_DETECTION_DISABLED and tree-shake this entire function.
-  const _tsDetectionDisabledEnv = process.env.AWS_SDK_JS_TYPESCRIPT_DETECTION_DISABLED;
+  const _tsDetectionDisabledEnv = process.env?.AWS_SDK_JS_TYPESCRIPT_DETECTION_DISABLED;
   if (_tsDetectionDisabledEnv === "true" || _tsDetectionDisabledEnv === "1") {
     tscVersion = null;
     return undefined;
--- a/packages-internal/core/src/submodules/client/util-user-agent-node/getTypeScriptUserAgentPair.ts
+++ b/packages-internal/core/src/submodules/client/util-user-agent-node/getTypeScriptUserAgentPair.ts
@@ -1,4 +1,3 @@
-import { booleanSelector, SelectorType } from "@smithy/core/config";
 import type { UserAgentPair } from "@smithy/types";
 import { readFile } from "node:fs/promises";
 import { join } from "node:path";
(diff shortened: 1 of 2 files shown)


packages-internal/core/src/submodules/client/util-user-agent-node/getTypeScriptUserAgentPair.ts
```diff
@@ -24,12 +23,10 @@ export const getTypeScriptUserAgentPair = async (): Promise<UserAgentPair | unde
...
   // Disable TypeScript detection if environment variable is set.
-  let isTypeScriptDetectionDisabled = false;
-  try {
-    isTypeScriptDetectionDisabled =
-      booleanSelector(process.env, "AWS_SDK_JS_TYPESCRIPT_DETECTION_DISABLED", SelectorType.ENV) || false;
-  } catch {}
-  if (isTypeScriptDetectionDisabled) {
+  // Direct property access (rather than booleanSelector) allows bundlers to statically replace
+  // process.env.AWS_SDK_JS_TYPESCRIPT_DETECTION_DISABLED and tree-shake this entire function.
+  const _tsDetectionDisabledEnv = process.env.AWS_SDK_JS_TYPESCRIPT_DETECTION_DISABLED;
```
kuhe: needs nullchecks

kuhe (changes requested):

packages-internal/core/src/submodules/client/util-user-agent-node/getTypeScriptUserAgentPair.ts
tsushanth: Added optional chaining: `process.env?.AWS_SDK_JS_TYPESCRIPT_DETECTION_DISABLED`.

trivikr:
Thank you @tsushanth for looking into the issue, and posting a PR.

This is replaced by https://github.com/aws/aws-sdk-js-v3/pull/8177, as AWS SDK for JavaScript team decided to remove the TypeScript version user-agent metric.

PR #8156, opened 2026-07-06
feat(core/client): propagate w3c trace headers

kuhe:
### Issue
modified from https://github.com/aws/aws-sdk-js-v3/pull/8147

### Description
propagate `traceparent` `tracestate` `baggage` headers from Lambda invoke store.

### Testing
CI, new unit tests

### Checklist
  - [ ] It's not a feature.
  - [x] No streams here.

---

### DIFF
--- a/packages-internal/core/src/submodules/client/middleware-recursion-detection/recursionDetectionMiddleware.ts
+++ b/packages-internal/core/src/submodules/client/middleware-recursion-detection/recursionDetectionMiddleware.ts
@@ -1,20 +1,31 @@
 // @ts-ignore
-import { InvokeStore } from "@aws/lambda-invoke-store";
+import { type InvokeStoreBase, InvokeStore } from "@aws/lambda-invoke-store";
 import { HttpRequest } from "@smithy/core/protocols";
 import type {
   BuildHandler,
   BuildHandlerArguments,
   BuildHandlerOutput,
   BuildMiddleware,
+  HeaderBag,
   MetadataBearer,
 } from "@smithy/types";
 
-const TRACE_ID_HEADER_NAME = "X-Amzn-Trace-Id";
-const ENV_LAMBDA_FUNCTION_NAME = "AWS_LAMBDA_FUNCTION_NAME";
-const ENV_TRACE_ID = "_X_AMZN_TRACE_ID";
+// env
+const AWS_LAMBDA_FUNCTION_NAME = "AWS_LAMBDA_FUNCTION_NAME";
+const _X_AMZN_TRACE_ID = "_X_AMZN_TRACE_ID";
+
+// headers
+const X_AMZN_TRACE_ID = "X-Amzn-Trace-Id";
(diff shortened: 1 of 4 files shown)


packages-internal/core/src/submodules/client/middleware-recursion-detection/recursionDetectionMiddleware.ts
```diff
@@ -25,26 +36,65 @@ export const recursionDetectionMiddleware =
...
+        }
+      }
     }
-    const functionName = process.env[ENV_LAMBDA_FUNCTION_NAME];
 
-    const traceIdFromEnv = process.env[ENV_TRACE_ID];
-    const invokeStore = await InvokeStore.getInstanceAsync();
-    const traceIdFromInvokeStore = invokeStore?.getXRayTraceId();
-    const traceId = traceIdFromInvokeStore ?? traceIdFromEnv;
+    {
```
maxday, line 63: When traceparent is already set in request headers (ie: by user), the current logic still unconditionally injects tracestate and baggage from InvokeStore. Do we want this? From the spec, I understand it differently 
```
Section 3.4: "If the value of the traceparent field wasn't changed before propagation, tracestate MUST NOT be modified as well."
```
WDYT?

kuhe: I didn't read the spec initially, only copied the code function. 

I think this is a good catch, I've updated this in the 2nd commit to only take the secondary headers from invokeStore if the traceparent was from the invokeStore.

maxday (approved):
Thanks!

packages-internal/core/src/submodules/client/middleware-recursion-detection/recursionDetectionMiddleware.ts
darklight3it: I agree. that's what i was doing in the original code

https://github.com/aws/aws-sdk-js-v3/pull/8147/changes#diff-81a677b66509984f1991c2bc235a7b66d993c5e8bbc20f50dfcfe78f89098defR47-R49

darklight3it: As a note. Sanitization is the only thing we should do regardless.

```
3.2.1 Header Name

Header name: traceparent

In order to increase interoperability across multiple protocols and encourage successful integration, by default vendors SHOULD keep the header name lowercase. The header name is a single word without any delimiters, for example, a hyphen (-).

Vendors MUST expect the header name in any case (upper, lower, mixed), and SHOULD send the header name in lowercase.
```

kuhe: @darklight3it oh, I see. I converted the code incorrectly, thanks.

trivikr (approved):

PR #8164, opened 2026-07-07
chore: use isolatedModules for types

kuhe:
### Issue
https://github.com/smithy-lang/smithy-typescript/pull/2126

### Description
enable isolatedModules compilation check to allow e.g. oxc single file transform in the future.

This PR only covers handwritten packages. A later followup will cover codegen compatibility with isolatedModuels.

### Testing
CI

### Checklist
  - [x] It's not a feature.
  - [x] No streams here.

---


lib/lib-dynamodb/tsconfig.types.json
```diff
@@ -5,7 +5,8 @@
     "rootDir": "src",
     "paths": {
       "@smithy/core/*": ["../../node_modules/@smithy/core/dist-types/submodules/*/index.d.ts"]
-    }
+    },
+    "isolatedModules": false
```
smilkuri, line 9: Why are we disabling for `lib-dynamodb` package?

smilkuri (approved):

lib/lib-dynamodb/tsconfig.types.json
kuhe: it will have codegen changes as a followup

PR #8189, opened 2026-07-13
feat(core): update clock skew detection threshold and configs

siddsriv:
### Issue
Internal JS-7004

### Description
Some updates for clock skew/retry behavior:
- detection threshold changes to 4 mins from 5 mins
- NTP midpoint formula for calculation of candidate skew
- Age header guard: discard skew candidate if response has an Age header
- Elapsed guard: discard skew candidate if round-trip exceeds 15 minutes
- new config option `disableClockSkewCorrection` 

Note that `disableClockSkewCorrection` config resolver lives in `httpAuthSchemes/aws_sdk/` (alongside the signer that consumes it) rather than a new submodule, following the `NODE_AUTH_SCHEME_PREFERENCE_OPTIONS` pattern, and because I want to keep this PR focused on the update. Adding a new submodule for this config would have some extra overhead.

smithy-ts Change: https://github.com/smithy-lang/smithy-typescript/pull/2145

### Testing
unit/integ testing

### Checklist
  - [ ] It's not a feature.
  - [x] No streams here.

---


codegen/smithy-aws-typescript-codegen/src/main/java/software/amazon/smithy/aws/typescript/codegen/AddDisableClockSkewCorrectionRuntimeConfig.java
```diff
@@ -0,0 +1,121 @@
...
+            switch (target) {
+                case BROWSER:
+                    runtimeConfigs.put("disableClockSkewCorrection", writer -> {
+                        writer.addImportSubmodule(
+                            "DEFAULT_DISABLE_CLOCK_SKEW_CORRECTION",
+                            null,
+                            AwsDependency.AWS_SDK_CORE,
+                            "/httpAuthSchemes"
+                        );
+                        writer.write("(() => Promise.resolve(DEFAULT_DISABLE_CLOCK_SKEW_CORRECTION))");
```
kuhe: just write `false` or `DEFAULT_DISABLE_CLOCK_SKEW_CORRECTION` as a value.

kuhe: now that you are importing core/httpAuthSchemes into a browser config, the browser submodule requires an index.browser.ts file. Look at how submodules/client is laid out.

Anything that has a browser implementation should export it, and anything that doesn't have a corresponding browser implementation should export as `const no = Symbol.for("node-only");`.

packages-internal/core/src/submodules/httpAuthSchemes/aws_sdk/AwsSdkSigV4ASigner.ts
```diff
@@ -29,9 +29,17 @@ export class AwsSdkSigV4ASigner extends AwsSdkSigV4Signer {
       signingRegionSet ?? [signingRegion]
     ).join(",");
 
-    signingProperties._preRequestSystemClockOffset = config.systemClockOffset;
+    const disabled = (await config.disableClockSkewCorrection?.()) === true;
```
kuhe:
```suggestion
    const noSkewCorrection = (await config.disableClockSkewCorrection?.()) === true;
```

packages-internal/core/src/submodules/httpAuthSchemes/aws_sdk/resolveDisableClockSkewCorrectionConfig.ts
```diff
@@ -0,0 +1,41 @@
+import { normalizeProvider } from "@smithy/core/client";
+import type { Provider } from "@smithy/types";
+
+import { DEFAULT_DISABLE_CLOCK_SKEW_CORRECTION } from "./NODE_DISABLE_CLOCK_SKEW_CORRECTION_CONFIG_OPTIONS";
```
kuhe: this should be imported from a different file. It ties this file to Node.js if left like this.

packages-internal/core/src/submodules/httpAuthSchemes/aws_sdk/clock-skew-node-config.ts
```diff
@@ -0,0 +1,26 @@
+import { type LoadedConfigSelectors, booleanSelector, SelectorType } from "@smithy/core/config";
```
kuhe, line 1: name this file normally like clock-skew-config.ts

packages-internal/core/src/submodules/httpAuthSchemes/utils/getUpdatedSystemClockOffset.ts
```diff
@@ -1,18 +1,52 @@
-import { isClockSkewed } from "./isClockSkewed";
-
 /**
  * @internal
```
kuhe: move access level tsdoc to after description

packages-internal/core/src/submodules/httpAuthSchemes/aws_sdk/AwsSdkSigV4Signer.ts
```diff
@@ -116,10 +118,18 @@ export class AwsSdkSigV4Signer implements HttpSigner {
       }
     }
 
-    // Capture the clock offset before signing so errorHandler can detect concurrent corrections.
-    signingProperties._preRequestSystemClockOffset = config.systemClockOffset;
+    const disabled = (await config.disableClockSkewCorrection?.()) === true;
```
kuhe:
```suggestion
    const noSkewCorrection = (await config.disableClockSkewCorrection?.()) === true;
```

packages-internal/core/src/submodules/httpAuthSchemes/aws_sdk/AwsSdkSigV4Signer.ts
```diff
@@ -129,30 +139,53 @@ export class AwsSdkSigV4Signer implements HttpSigner {
...
+          const ageHeader = getAgeHeader(errorException.$response);
+          const newOffset = getUpdatedSystemClockOffset(
+            serverTime,
+            config.systemClockOffset,
+            timeRequestSent,
+            ageHeader
+          );
+
+          // Always update ClientSkew
+          config.systemClockOffset = newOffset;
```
kuhe, line 158: why update the offset if the request does not qualify as having been skewed?

packages-internal/core/src/submodules/httpAuthSchemes/aws_sdk/resolveDisableClockSkewCorrectionConfig.ts
```diff
@@ -0,0 +1,41 @@
...
+
+/**
+ * @public
+ */
+export interface DisableClockSkewCorrectionInputConfig {
```
kuhe: This config stack should be merged into resolveAwsSdkSigV4Config. The only thing concerned with clock skew is sigv4, I believe.

That means the input config type, resolved config type, and resolver function.

kuhe (changes requested):

packages-internal/core/src/submodules/httpAuthSchemes/aws_sdk/resolveDisableClockSkewCorrectionConfig.ts
siddsriv: merged, to note, i also changed the predicate in the codegen file, this change will now only apply according to `isAwsService` since non-AWS SigV4 services (like weather) don't go through `AwsSdkCustomizeSigV4Auth` and never get `AwsSdkSigV4AuthInputConfig` added to their type. The predicate now matches the same set of services that actually resolve this config through `resolveAwsSdkSigV4Config`.

packages-internal/core/src/submodules/httpAuthSchemes/aws_sdk/AwsSdkSigV4Signer.ts
kuhe: comment should explain why: heal the clock offset over time even if the errors are not breaching the threshold

packages-internal/core/src/submodules/httpAuthSchemes/aws_sdk/resolveAwsSdkSigV4Config.ts
```diff
@@ -21,6 +21,8 @@ import type {
   RequestSigner,
 } from "@smithy/types";
 
+import { DEFAULT_DISABLE_CLOCK_SKEW_CORRECTION } from "./clock-skew-config";
```
kuhe: needs to be imported from a file that doesn't have Node.js imports

kuhe: 🤔 actually, booleanSelector might just be `Symbol.for('node-only')` so this could be fine. the bundler tests cover this, so if they pass it's ok.

siddsriv: might just move the default to a new file to avoid overcrowding the sigv4 resolver file

packages-internal/core/src/submodules/httpAuthSchemes/index.ts
```diff
@@ -4,6 +4,10 @@ export {
   validateSigningProperties,
   AwsSdkSigV4ASigner,
   NODE_AUTH_SCHEME_PREFERENCE_OPTIONS,
+  DEFAULT_DISABLE_CLOCK_SKEW_CORRECTION,
+  ENV_DISABLE_CLOCK_SKEW_CORRECTION,
+  CONFIG_DISABLE_CLOCK_SKEW_CORRECTION,
+  NODE_DISABLE_CLOCK_SKEW_CORRECTION_CONFIG_OPTIONS,
```
kuhe: these don't need exporting, do they?

private/aws-protocoltests-ec2-schema/src/runtimeConfig.browser.ts
```diff
@@ -35,6 +35,7 @@ export const getRuntimeConfig = (config: EC2ProtocolClientConfig) => {
     bodyLengthChecker: config?.bodyLengthChecker ?? calculateBodyLength,
     credentialDefaultProvider: config?.credentialDefaultProvider ?? ((_: unknown) => () => Promise.reject(new Error("Credential is missing"))),
     defaultUserAgentProvider: config?.defaultUserAgentProvider ?? createDefaultUserAgentProvider({serviceId: clientSharedValues.serviceId, clientVersion: packageInfo.version}),
+    disableClockSkewCorrection: config?.disableClockSkewCorrection ?? false,
```
kuhe: don't need to codegen this since the sigv4 resolver sets the default value.

avoids codegen diff for the 400 clients

private/aws-protocoltests-ec2-schema/src/runtimeConfig.ts
```diff
@@ -55,6 +58,7 @@ export const getRuntimeConfig = (config: EC2ProtocolClientConfig) => {
     bodyLengthChecker: config?.bodyLengthChecker ?? calculateBodyLength,
     credentialDefaultProvider: config?.credentialDefaultProvider ?? credentialDefaultProvider,
     defaultUserAgentProvider: config?.defaultUserAgentProvider ?? createDefaultUserAgentProvider({serviceId: clientSharedValues.serviceId, clientVersion: packageInfo.version}),
+    disableClockSkewCorrection: config?.disableClockSkewCorrection ?? loadNodeConfig(NODE_DISABLE_CLOCK_SKEW_CORRECTION_CONFIG_OPTIONS, loaderConfig),
```
kuhe: can you get the sigv4 config resolver to run this (Node.js only) so we don't end up with hundreds of these?

packages-internal/core/src/submodules/httpAuthSchemes/index.ts
siddsriv: revised exports here

packages-internal/core/src/submodules/httpAuthSchemes/aws_sdk/clock-skew-defaults.browser.ts
```diff
@@ -0,0 +1,5 @@
+/**
+ * @internal
```
kuhe: `@internal` goes below description

packages-internal/core/src/submodules/httpAuthSchemes/aws_sdk/clock-skew-defaults.ts
```diff
@@ -0,0 +1,10 @@
+import { loadConfig } from "@smithy/core/config";
+
+import { NODE_DISABLE_CLOCK_SKEW_CORRECTION_CONFIG_OPTIONS } from "./clock-skew-node-config";
+
+/**
+ * @internal
```
kuhe: `@internal` goes below description

packages-internal/core/src/submodules/httpAuthSchemes/aws_sdk/resolveAwsSdkSigV4Config.ts
```diff
@@ -21,6 +21,8 @@ import type {
   RequestSigner,
 } from "@smithy/types";
 
+import { DEFAULT_DISABLE_CLOCK_SKEW_CORRECTION } from "./clock-skew-defaults";
```
kuhe: you have to bind this in the export index

static impl (original)
```ts
import dep;

export function impl() {
  dep();
}
```

dependency binding implementation
```ts
export function bindImpl(dep) {
  return function impl() {
    dep();
  }
}
```

index for node
```ts
import dep from 'x.browser.ts';

export const impl = bindImpl(dep);
```

index for browser
```ts
import dep from 'x.ts';

export const impl = bindImpl(dep);
```

packages-internal/core/package.json
```diff
@@ -33,7 +33,9 @@
   "main": "./dist-cjs/index.js",
   "module": "./dist-es/index.js",
   "browser": {
-    "./dist-es/submodules/client/index.js": "./dist-es/submodules/client/index.browser.js"
+    "./dist-es/submodules/client/index.js": "./dist-es/submodules/client/index.browser.js",
+    "./dist-es/submodules/httpAuthSchemes/aws_sdk/clock-skew-defaults.js": "./dist-es/submodules/httpAuthSchemes/aws_sdk/clock-skew-defaults.browser.js",
```
kuhe: this probably works, but since we have the browser index, we do it with the `bindImpl` pattern, which is less reliant on the bundler recognizing this metadata.

Because, in the `exports` package.json field this information isn't available, and who knows if a particular bundler is going to know to merge these two pieces of information.

packages-internal/core/src/submodules/httpAuthSchemes/aws_sdk/resolveAwsSdkSigV4Config.spec.ts
```diff
@@ -1,7 +1,9 @@
 import type { AwsCredentialIdentityProvider } from "@smithy/types";
 import { describe, expect, test as it, vi } from "vitest";
 
-import { resolveAwsSdkSigV4Config } from "./resolveAwsSdkSigV4Config";
+import { bindResolveAwsSdkSigV4Config } from "./resolveAwsSdkSigV4Config";
```
kuhe: you can import `resolveAwsSdkSigV4Config` from index

packages-internal/core/src/submodules/httpAuthSchemes/index.browser.ts
```diff
@@ -0,0 +1,32 @@
+const no = Symbol.for("node-only");
+
+export {
+  AwsSdkSigV4Signer,
+  AWSSDKSigV4Signer,
+  validateSigningProperties,
+  AwsSdkSigV4ASigner,
+  resolveAwsSdkSigV4AConfig,
+  resolveAWSSDKSigV4Config,
+  bindResolveAwsSdkSigV4Config,
```
kuhe: do not export this

packages-internal/core/src/submodules/httpAuthSchemes/index.ts
```diff
@@ -6,7 +6,7 @@ export {
   NODE_AUTH_SCHEME_PREFERENCE_OPTIONS,
   resolveAwsSdkSigV4AConfig,
   NODE_SIGV4A_CONFIG_OPTIONS,
-  resolveAwsSdkSigV4Config,
+  bindResolveAwsSdkSigV4Config,
```
kuhe: don't export this

scripts/validation/api.json
```diff
@@ -336,6 +336,7 @@
     "AWSSDKSigV4PreviouslyResolved": "type(interface)",
     "AwsSdkSigV4Signer": "function",
     "AWSSDKSigV4Signer": "function",
+    "bindResolveAwsSdkSigV4Config": "function",
```
kuhe: don't export the binding function

packages-internal/core/src/submodules/httpAuthSchemes/aws_sdk/AwsSdkSigV4Signer.ts
```diff
@@ -129,30 +139,54 @@ export class AwsSdkSigV4Signer implements HttpSigner {
...
+      const ageHeader = getAgeHeader(httpResponse);
+      config.systemClockOffset = getUpdatedSystemClockOffset(
+        dateHeader,
+        config.systemClockOffset,
+        timeRequestSent,
+        ageHeader
+      );
     }
   }
 }
```
kuhe, line 192: add more test cases to `systemClockOffset.integ.spec.ts`. They may be redundant to the unit tests but integ coverage is needed for features.

packages-internal/core/src/submodules/httpAuthSchemes/aws_sdk/resolveAwsSdkSigV4Config.ts
```diff
@@ -290,21 +309,22 @@ export interface AWSSDKSigV4PreviouslyResolved extends AwsSdkSigV4PreviouslyReso
 export interface AWSSDKSigV4AuthResolvedConfig extends AwsSdkSigV4AuthResolvedConfig {}
 
 /**
+ * @deprecated renamed to {@link bindResolveAwsSdkSigV4Config}
+ *
  * @internal
- * @deprecated renamed to {@link resolveAwsSdkSigV4Config}
  */
-export const resolveAWSSDKSigV4Config = resolveAwsSdkSigV4Config;
+export const resolveAWSSDKSigV4Config = bindResolveAwsSdkSigV4Config;
```
kuhe: this changes the signature entirely. This is simply a naming alias and can be exported from the new indexes

packages-internal/core/src/submodules/httpAuthSchemes/aws_sdk/resolveAwsSdkSigV4Config.ts
```diff
@@ -342,7 +362,7 @@ function normalizeCredentialProvider(
  * Uses a state marker on the function to avoid doing this more than once.
  */
 function bindCallerConfig(
-  config: Parameters<typeof resolveAwsSdkSigV4Config>[0],
+  config: AwsSdkSigV4AuthInputConfig & AwsSdkSigV4PreviouslyResolved,
```
kuhe: make a local type alias for this intersection type and reuse it, including in bindResolveAwsSdkSigV4Config

packages-internal/core/src/submodules/httpAuthSchemes/aws_sdk/AwsSdkSigV4Signer.ts
siddsriv: added a case or two, between this and `packages-internal/core/integ/clock-skew.integ.spec.ts` i think we should be covered? let me know if you see gaps or have recommendations for more test cases

kuhe (approved):

PR #8193, opened 2026-07-15
chore: remove redundant type assertions

trivikr:
### Issue

N/A

### Description

Drop assertions that are unnecessary because the values are already
narrowed by a preceding type guard in the same scope.

### Testing
CI

### Checklist
  - [x] It's not a feature.
  - [x] No streams here.

---

### DIFF
--- a/packages-internal/eventstream-handler-node/src/EventStreamPayloadHandler.ts
+++ b/packages-internal/eventstream-handler-node/src/EventStreamPayloadHandler.ts
@@ -57,7 +57,7 @@ export class EventStreamPayloadHandler implements IEventStreamPayloadHandler {
       throw new Error("Eventstream payload must be a Readable stream.");
     }
 
-    const payloadStream = payload as Readable;
+    const payloadStream = payload;
     request.body = new PassThrough({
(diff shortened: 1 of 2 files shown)


packages-internal/eventstream-handler-node/src/EventStreamPayloadHandler.ts
```diff
@@ -57,7 +57,7 @@ export class EventStreamPayloadHandler implements IEventStreamPayloadHandler {
       throw new Error("Eventstream payload must be a Readable stream.");
     }
 
-    const payloadStream = payload as Readable;
```
kuhe, line 60: no need to keep the variable then

trivikr: Removed in [`bdba015` (this PR)](https://github.com/aws/aws-sdk-js-v3/pull/8193/changes/bdba0150c32176ea285d563b1552c6f36ec400a8)

siddsriv (approved):

kuhe (approved):

PR #8216, opened 2026-07-23
fix(credential-provider-login): always read token from disk

smilkuri:
### Issue
Internal V2298487436

### Description
updated the credential fetcher to always re-read the token from disk and reload the token from disk to check whether it was already refreshed by another process before initiating a refresh.

### Testing
Added an integ test
```
yarn test: integration
 Test Files  1 passed (1)
      Tests  48 passed | 3 skipped (51)
   Start at  19:19:26
   Duration  2.86s (transform 1.08s, setup 0ms, import 1.50s, tests 883ms, environment 0ms)
```

### Checklist
  - [x] It's not a feature.
  - [x] No streams here.

---

### DIFF
--- a/packages-internal/credential-provider-login/src/LoginCredentialsFetcher.ts
+++ b/packages-internal/credential-provider-login/src/LoginCredentialsFetcher.ts
@@ -1,9 +1,5 @@
 import type { CreateOAuth2TokenCommandInput } from "@aws-sdk/nested-clients/signin";
-import type {
-  AwsCredentialIdentity,
-  AwsIdentityProperties,
-  Logger,
-} from "@aws-sdk/types";
+import type { AwsCredentialIdentity, AwsIdentityProperties, Logger } from "@aws-sdk/types";
 import { CredentialsProviderError } from "@smithy/core/config";
 import { HttpRequest } from "@smithy/core/protocols";
 import type {
@@ -13,12 +9,7 @@ import type {
   IniSection,
   MetadataBearer,
 } from "@smithy/types";
-import {
-  createHash,
-  createPrivateKey,
-  createPublicKey,
-  sign,
-} from "node:crypto";
+import { createHash, createPrivateKey, createPublicKey, sign } from "node:crypto";
 import { promises as fs } from "node:fs";
 import { homedir } from "node:os";
 import { dirname, join } from "node:path";
@@ -35,8 +26,8 @@ export class LoginCredentialsFetcher {
   public constructor(
     private readonly profileData: IniSection,
     private readonly init?: FromLoginCredentialsInit,
-    private readonly callerClientConfig?: AwsIdentityProperties["callerClientConfig"],
-  ) { }
+    private readonly callerClientConfig?: AwsIdentityProperties["callerClientConfig"]
+  ) {}
 
   /**
    * Loads credentials and refreshes if necessary
@@ -47,7 +38,7 @@ export class LoginCredentialsFetcher {
     if (!token) {
       throw new CredentialsProviderError(
         `Failed to load a token for session ${this.loginSession}, please re-authenticate using aws login`,
-        { tryNextLink: false, logger: this.logger },
+        { tryNextLink: false, logger: this.logger }
       );
     }
 
@@ -61,13 +52,7 @@ export class LoginCredentialsFetcher {
       return this.refresh(token);
     }
 
-    return {
-      accessKeyId: accessToken.accessKeyId,
-      secretAccessKey: accessToken.secretAccessKey,
-      sessionToken: accessToken.sessionToken,
-      accountId: accessToken.accountId,
-      expiration: new Date(accessToken.expiresAt),
-    };
+    return this.toCredentials(token.accessToken);
   }
 
   private get logger(): Logger | undefined {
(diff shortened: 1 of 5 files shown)


packages-internal/credential-provider-login/src/LoginCredentialsFetcher.ts
```diff
@@ -26,8 +35,8 @@ export class LoginCredentialsFetcher {
   public constructor(
     private readonly profileData: IniSection,
     private readonly init?: FromLoginCredentialsInit,
-    private readonly callerClientConfig?: AwsIdentityProperties["callerClientConfig"]
-  ) {}
+    private readonly callerClientConfig?: AwsIdentityProperties["callerClientConfig"],
+  ) { }
```
kuhe: what are these formatting changes from? `make format` to run oxfmt

packages-internal/credential-provider-login/src/LoginCredentialsFetcher.ts
```diff
@@ -70,7 +79,22 @@ export class LoginCredentialsFetcher {
...
+    const freshExpiry = new Date(freshToken.accessToken.expiresAt).getTime();
+    if (freshExpiry - Date.now() > LoginCredentialsFetcher.REFRESH_THRESHOLD) {
+      return {
+        accessKeyId: freshToken.accessToken.accessKeyId,
+        secretAccessKey: freshToken.accessToken.secretAccessKey,
+        sessionToken: freshToken.accessToken.sessionToken,
+        accountId: freshToken.accessToken.accountId,
+        expiration: new Date(freshToken.accessToken.expiresAt),
+      };
+    }
```
kuhe, line 87: this object mapping is executed 3 (?) times in this file and should be a function

packages-internal/credential-provider-login/src/LoginCredentialsFetcher.ts
```diff
@@ -70,7 +79,22 @@ export class LoginCredentialsFetcher {
   }
 
   private async refresh(token: LoginToken): Promise<AwsCredentialIdentity> {
```
kuhe, line 76: arg is unused now, but... why overwrite this with the disk token always?

what if the disk token is missing or for some reason is expired when this token is not expired?

smilkuri: You are right. The token arg should be used as fallback if the disk read fails or the API refresh fails, we should fall back to the original token as long as it hasn't expired. I'll update the code to use token as fallback.

packages-internal/credential-provider-login/src/LoginCredentialsFetcher.ts
```diff
@@ -69,7 +63,24 @@ export class LoginCredentialsFetcher {
...
+      secretAccessKey: token.accessToken.secretAccessKey,
+      sessionToken: token.accessToken.sessionToken,
+      accountId: token.accessToken.accountId,
+      expiration: new Date(token.accessToken.expiresAt),
+    };
+  }
+
   private async refresh(token: LoginToken): Promise<AwsCredentialIdentity> {
+    // first reload the token from disk to ensure the token hasn't already been refreshed externally.
+    const freshToken = await this.loadToken().catch(() => token);
```
kuhe: the fresh token should also be discarded if the existing token is unexpired and the fresh token is expired

packages-internal/credential-provider-login/src/LoginCredentialsFetcher.ts
```diff
@@ -136,15 +147,7 @@ export class LoginCredentialsFetcher {
...
       await this.saveToken(updatedToken);
 
-      const newAccessToken = updatedToken.accessToken;
-
-      return {
-        accessKeyId: newAccessToken.accessKeyId,
-        secretAccessKey: newAccessToken.secretAccessKey,
-        sessionToken: newAccessToken.sessionToken,
-        accountId: newAccessToken.accountId,
-        expiration,
```
kuhe, line 146: is this no longer passing expiration?

packages-internal/credential-provider-login/src/LoginCredentialsFetcher.ts
```diff
@@ -69,7 +63,24 @@ export class LoginCredentialsFetcher {
     return this.profileData.login_session!;
   }
 
+  private toCredentials(token: LoginToken): AwsCredentialIdentity {
```
kuhe: this input looks like it should be `token: LoginToken["accessToken"]` or the specific type of it if available.

packages-internal/credential-provider-login/src/LoginCredentialsFetcher.ts
```diff
@@ -180,12 +197,7 @@ export class LoginCredentialsFetcher {
   private async loadToken(): Promise<LoginToken> {
     const tokenFilePath = this.getTokenFilePath();
     try {
-      let tokenData: string;
-      try {
-        tokenData = await readFile(tokenFilePath, { ignoreCache: this.init?.ignoreCache });
-      } catch {
-        tokenData = await fs.readFile(tokenFilePath, "utf8");
-      }
```
kuhe, line 188: what was this try catch for?

packages-internal/credential-provider-login/src/LoginCredentialsFetcher.ts
```diff
@@ -282,9 +294,9 @@ export class LoginCredentialsFetcher {
           if (HttpRequest.isInstance(args.request)) {
             const request = args.request as HttpRequest;
             // Extract the actual endpoint URL after resolution
-            const actualEndpoint = `${request.protocol}//${request.hostname}${request.port ? `:${request.port}` : ""}${
-              request.path
-            }`;
+            const actualEndpoint = `${request.protocol}//${request.hostname}${
+              request.port ? `:${request.port}` : ""
+            }${request.path}`;
```
kuhe, line 300: formatting change?

packages-internal/credential-provider-login/src/LoginCredentialsFetcher.ts
```diff
@@ -69,7 +63,28 @@ export class LoginCredentialsFetcher {
...
+      secretAccessKey: token.secretAccessKey,
+      sessionToken: token.sessionToken,
+      accountId: token.accountId,
+      expiration: new Date(token.expiresAt),
+    };
+  }
+
   private async refresh(token: LoginToken): Promise<AwsCredentialIdentity> {
+    // first reload the token from disk to ensure the token hasn't already been refreshed externally.
+    // Pick whichever token has the later expiry.
```
kuhe: no, it should not take the token with the later expiry.

do this:
1. diskToken unexpired
2. inputToken unexpired
3. diskToken, expired
4. inputToken, expired

packages-internal/credential-provider-login/src/LoginCredentialsFetcher.ts
smilkuri: In the initial implementation we had that, the first readFile was from `@smithy/core/config` which reads from module level cache and the later one was fallback to read from disk. Now we removed it since we always want to read from disk.

packages-internal/credential-provider-login/src/LoginCredentialsFetcher.ts
smilkuri: Yes

packages-internal/credential-provider-login/src/LoginCredentialsFetcher.ts
```diff
@@ -75,11 +75,12 @@ export class LoginCredentialsFetcher {
 
   private async refresh(token: LoginToken): Promise<AwsCredentialIdentity> {
     // first reload the token from disk to ensure the token hasn't already been refreshed externally.
```
kuhe: wording: 

> first reload the token from disk to ensure the token hasn't already been refreshed externally.

This kind of sounds like we want to "ensure the token hasn't been refreshed", although the implied full statement is 

> first reload the token from disk to ensure the token hasn't already been refreshed externally **without us noticing**.

Say instead "first reload the token from disk **in case** the token **has** been refreshed externally"

kuhe (approved):

PR #8222, opened 2026-07-23
fix(core/protocols): avoid Buffer.slice in return ByteJsonShapeSerializer.flush

kuhe:
### Issue
P478244064

### Description
Change the way Uint8Arrays are returned from the ByteJsonShapeSerializer. 
- instead of Buffer.slice returning the value (view mutation risk?), the buffer itself is returned, handing off ownership, and the serializer allocates a new write buffer.

### Testing
CI

### Checklist
  - [x] It's not a feature.
  - [x] No streams here.

---


kuhe:
Encountered https://issues.chromium.org/issues/521080746.

trivikr (approved):

PR #8214, opened 2026-07-22
chore(core/protocols): JSON serde consistency testing and perf tuning

kuhe:
### Issue
n/a

### Description
- uses a `defineKey` call to prep writes to the `__proto__` key across AWS protocols.
- detects a fast path in JSON map ser: use JSON.stringify on the entire map if the map is a simple type like `Map<string, number>`. This covers enough of the gap such that both experimental serde implementations are now faster than the original multi-pass JSON serde.

### Net effects
  
For typical AWS SDK usage, switching to the byte serializer and buffer deserializer will give 30–65% faster serde on common payloads, with no regressions in other cases. The only scenario that doesn't improve is the small widget struct with bigint members (reviver still required). Public AWS SDK models don't use bigint or bigdecimal so as to even need a reviver.

### Testing
added a bit of fuzz testing to the JSON serde

### Checklist
- [ ] If the PR is a feature, add integration tests (`*.integ.spec.ts`) or E2E tests.
  - [x] It's not a feature.
- [ ] My E2E tests are resilient to concurrent i/o.
  - [x] I didn't write any E2E tests.
- [ ] I added access level annotations e.g. `@public`, `@internal` tags and enabled doc generation on the package. Remember that access level annotations go below the description, not above.
  - [x] I didn't add any public functions.
- [ ] Streams - how do they work?? My WebStream readers/locks are properly lifecycled. Node.js stream backpressure is handled. Error handling.
  - [x] No streams here.

```
JSON Shape Serializer Benchmark (isolated processes)
══════════════════════════════════════════════════════════════════════════════════════════

Running multipass serializer... done.
Running byte serializer...      done.

Scenario                       │     Size │    Multipass │         Byte │        Δ
───────────────────────────────┼──────────┼──────────────┼──────────────┼─────────
nested-struct (depth=4)        │    695 B │   17.5 kb/ms │   15.5 kb/ms │   -13.2%
nested-struct (depth=16)       │   2.3 KB │   38.7 kb/ms │   40.0 kb/ms │    +3.4%
nested-struct (depth=64)       │   8.9 KB │   64.0 kb/ms │   65.6 kb/ms │    +2.4%
nested-struct (depth=256)      │  35.4 KB │   73.3 kb/ms │   89.4 kb/ms │   +18.0%
nested-struct (depth=1024)     │ 141.1 KB │   46.6 kb/ms │   88.2 kb/ms │   +47.1%
[...]

PR #8225, opened 2026-07-24
fix(core): parse JSON numbers in exponent notation

### LINKED ISSUE #8224: Deserialization fails on JSON numbers in exponent notation (regression in @aws-sdk/core 3.977.0) — breaks all ecs:DescribeServices calls

- [x] I've gone through [Developer Guide](https://docs.aws.amazon.com/sdk-for-javascript/v3/developer-guide/) and [API reference](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/)
- [x] I've searched for [previous similar issues](https://github.com/aws/aws-sdk-js-v3/issues) and didn't find any solution
- [x] I've read the pinned notice in smithy-lang/smithy-typescript#1562 and am filing here rather than in smithy-typescript

### Describe the bug

Response deserialization throws whenever an `awsJson` response body contains a JSON number written in **exponent notation** (e.g. `1.784316439424E9`). Exponent notation is valid JSON per RFC 8259, and AWS services do emit it — ECS serializes epoch-second timestamps that way, so *every* `ecs:DescribeServices` call now fails.

```
Error: @smithy/core/serde - NumericValue must only contain [0-9], at most one decimal point ".", and an optional negation prefix "-".
  Deserialization error: to see the raw response, inspect the hidden field {error}.$response on this object.
```

The same responses parse correctly with botocore / the AWS CLI, and parsed correctly with the JS SDK before this regression.

### Regression Issue

**Last good:** `@aws-sdk/core@3.976.0`
**First bad:** `@aws-sdk/core@3.977.0` (published 2026-07-24, shipped with `v3.1095.0`)

Suspect: #8214 (`chore(core/protocols): JSON serde consistency testing and perf tuning`), which touches `parseJsonBody.ts`, `needsReviver.ts`, and `JsonShapeDeserializer.ts`.

Note this floats independently of the service client — pinning `@aws-sdk/client-*` or `@smithy/core` does **not** avoid it, because `@aws-sdk/core` is resolved by range.

### SDK version number

`@aws-sdk/core@3.977.0`, reproduced with `@aws-sdk/client-ecs@3.1094.0` and `3.1095.0` (any client using the awsJson protocol should reproduce)

Node.js

Node.js v24.12.0 (Linux and macOS). Requires a runtime with `JSON.parse` source access (Node ≥ 22) — on older Node the reviver's `context` is undefined and the bug does not trigger.

### Reproduction Steps

Fully self-contained: no AWS account or credentials needed, since the HTTP layer is stubbed with a canned response body.

```bash
mkdir repro && cd repro
npm init -y >/dev/null
npm pkg set type=module
npm install @aws-sdk/client-ecs@3.1094.0 @smithy/protocol-http
```

`repro.mjs`:

```js
import { ECS } from "@aws-sdk/client-ecs";
import { HttpResponse } from "@smithy/protocol-http";
import { Readable } from "node:stream";

// A minimal awsJson response. `createdAt` uses exponent notation, exactly as
// real AWS services serialize epoch-second timestamps.
const BODY =
  '{"failures":[],"services":[{"serviceName":"my-service","status":"ACTIVE","createdAt":1.784316439424E9}]}';

const client = new ECS({
  region: "us-east-1",
  credentials: { accessKeyId: "AKIAIOSFODNN7EXAMPLE", secretAccessKey: "secret" },
  requestHandler: {
    async handle() {
      return {
        response: new HttpResponse({
          statusCode: 200,
          headers: { "content-type": "application/x-amz-json-1.1" },
          body: Readable.from([BODY]),
        }),
      };
    },
  },
});

try {
  const out = await client.describeServices({ cluster: "my-cluster", services: ["my-service"] });
  console.log("OK ->", JSON.stringify(out.services));
} catch (e) {
  console.log("FAIL ->", e.message.split("\n")[0]);
}
```

```bash
node repro.mjs
# FAIL -> @smithy/core/serde - NumericValue must only contain [0-9], at most one decimal point ".", ...

# Now pin the previous version of @aws-sdk/core and repeat:
npm pkg set overrides.@aws-sdk/core=3.976.0 && npm install
node repro.mjs
# OK -> [{"serviceName":"my-service","status":"ACTIVE","createdAt":"2026-07-17T19:27:19.424Z"}]
```

Against the live API, `ecs:DescribeServices` reproduces it for any service, because ECS returns `createdAt` / `updatedAt` in exponent form.

### Observed Behavior

`describeServices` (and any awsJson call whose response contains an exponent-notation number) rejects with the `NumericValue` error above. Nothing is returned to the caller, and the error text names neither the field nor the offending value.

### Expected Behavior

The response deserializes, with `createdAt` becoming a `Date` (as it does on `@aws-sdk/core@3.976.0`). Exponent notation is valid JSON and should be accepted wherever a plain decimal is.

### Possible Solution

The reviver in `@aws-sdk/core` (`submodules/protocols/index.js`, `jsonReviver`) decides a token is precision-losing by comparing the raw source text against `String(value)`:

```js
function jsonReviver(key, value, context) {
  if (context?.source) {
    const numericString = context.source;
    if (typeof value === "number") {
      if (value > Number.MAX_SAFE_INTEGER || value < Number.MIN_SAFE_INTEGER || numericString !== String(value)) {
        const isFractional = numericString.includes(".");
        if (isFractional) {
          return new NumericValue(numericString, "bigDecimal");  // throws on "1.78E9"
        } else {
          return BigInt(numericString);                          // throws on "1E9"
        }
      }
    }
  }
  return value;
}
```

An exponent-notation token always fails `numericString !== String(value)` (`"1.784316439424E9" !== "1784316439.424"`), so it is misclassified as precision-losing even though the `double` round-trips exactly. Both branches then reject it:

- fractional → `NumericValue`, whose format is `/^-?\d*(\.\d+)?$/` and forbids `E`
- non-fractional → `BigInt("1E9")` throws `SyntaxError: Cannot convert 1E9 to a BigInt`

Suggested fixes, roughly in order of smallest blast radius:

1. Normalize the exponent away before the comparison / before constructing `NumericValue`, so exponent form is expanded to plain decimal.
2. Treat a token as precision-losing only when it genuinely does not round-trip — compare numerically rather than by string equality (e.g. against `Number(numericString)`), which lets `1.78E9` fall through to the plain `number` path.
3. Widen the `NumericValue` format in `@smithy/core` to accept the full JSON number grammar, including exponents, and make the non-fractional branch use a parser that accepts them.

Whichever direction, a regression test with exponent-notation input on both branches (`1.5E9` and `1E9`) would pin it.

Impact is broad and not opt-in: any code path that reads an ECS service description with the JS SDK fails, which includes common CI/CD deployment tooling built on the SDK. The failure surfaces as an opaque error with no field name, and — because tools often capture the SDK's stdout — can present as an exit code with no message at all, which makes it expensive to diagnose.

`botocore` and the AWS CLI parse the identical response bodies without complaint, so this is specific to the JS SDK's reviver.

MohMaya:
### Issue

Fixes #8224

### Description

`jsonReviver` throws on any JSON number written in exponent notation, e.g. `1.784316439424E9`:

```
Error: @smithy/core/serde - NumericValue must only contain [0-9], at most one decimal point ".", and an optional negation prefix "-".
```

Exponent notation is valid JSON, and AWS services emit it — ECS serializes epoch-second timestamps that way, so every `ecs:DescribeServices` response currently fails to deserialize.

The reviver decides a token is precision-losing by comparing the raw source text against `String(value)`:

```ts
if (value > Number.MAX_SAFE_INTEGER || value < Number.MIN_SAFE_INTEGER || numericString !== String(value)) {
  const isFractional = numericString.includes(".");
  if (isFractional) {
    return new NumericValue(numericString, "bigDecimal");
  } else {
    return BigInt(numericString);
  }
}
```

An exponent-notation token always fails that string comparison (`"1.784316439424E9" !== "1784316439.424"`), so it is treated as precision-losing even when the `double` round-trips exactly. Both branches then reject it:

- fractional → `NumericValue`, whose format is `/^-?\d*(\.\d+)?$/` and forbids `E`
- non-fractional → `BigInt("1E9")` throws `SyntaxError: Cannot convert 1E9 to a BigInt`

This change normalizes exponent notation to plain decimal *before* the classification, so the existing logic sees a form it understands. Values that round-trip (the common case, including ECS timestamps) now fall through to a plain `number`; values that genuinely exceed `double` precision are still preserved as `BigInt` / `NumericValue`.

Expansion is bounded by `MAX_EXPANDED_DIGITS` (1024) so an extreme exponent cannot cause a large string allocation; past that bound the reviver returns the parsed double rather than throwing.

### Behavior notes for reviewers

Two consequences worth a look, both currently exceptions rather than values:

- `1e1000` / `1e-1000` were `SyntaxError` / `NumericValue` throws; they are now preserved exactly (`BigInt` and `NumericValue` respectively) instead of collapsing to `Infinity` / `0`. This follows the reviver's stated purpose, but it is a behavior choice — happy to return the double instead if you'd prefer the narrower change.
- Beyond `MAX_EXPANDED_DIGITS` the parsed double is returned (`Infinity` / `0`), matching no-reviver behavior. The threshold is arbitrary; say the word if you want a different bound.

I did not widen `NumericValue`'s format regex in `smithy-typescript`, since fixing the classification here keeps the change in one repo. If you'd rather accept the full JSON number grammar there, that would also solve it and I'm glad to follow up.

### Testing

Added cases to `jsonReviver.spec.ts`, gated on `contextSourceAvailable` like the existing tests:

- exponent notation parses to plain numbers — including the ECS-style `1.784316439424E9`, integral `1E9`, lowercase `1e9`, negative `-2.5E3`, and fractional `1.23E1`
- precision is still preserved for exponent values that need it — `1.5E-10` → `NumericValue("0.00000000015")`, `1.2345678901234567890E30` → `BigInt`
- extreme exponents fall back to the parsed double instead of throwing

The existing tests in that file (small ints/decimals, 37-digit `BigInt`, 41-digit `bigDecimal`) are unchanged and still pass — those inputs contain no exponent, so they take the original path untouched.

Verified end to end by patching the compiled fix into `@aws-sdk/core@3.977.0` in a local install: `describeServices` goes from the error above to returning `createdAt` as a `Date`, both against a stubbed response and against the live ECS API.

`tsc --strict --noEmit` is clean on the changed file.

### Additional context

Regression range: `@aws-sdk/core@3.976.0` is fine, `3.977.0` fails. Suspect #8214, which enabled the reviver more broadly via `needsReviver` / `parseJsonBody`.

`botocore` parses these identical bodies without complaint, so the divergence is specific to this reviver.

----

*By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.*

### DIFF
--- a/packages-internal/core/src/submodules/protocols/json/jsonReviver.spec.ts
+++ b/packages-internal/core/src/submodules/protocols/json/jsonReviver.spec.ts
@@ -58,4 +58,41 @@ describe(jsonReviver.name, () => {
     expect(parsed.bigint).toBe(1e36);
     expect(parsed.bigDecimal).toEqual(0.12345);
   });
+
+  (contextSourceAvailable ? it : it.skip)("parses numbers in exponent notation", async () => {
+    // AWS services serialize epoch-second timestamps this way, e.g. ECS DescribeServices.
+    const data = `
+    {
+      "epochSeconds": 1.784316439424E9,
+      "integral": 1E9,
+      "lowercase": 1e9,
+      "negative": -2.5E3,
+      "fraction": 1.23E1
+    }`;
+
+    const parsed = JSON.parse(data, jsonReviver);
+    expect(parsed.epochSeconds).toBe(1784316439.424);
+    expect(parsed.integral).toBe(1000000000);
+    expect(parsed.lowercase).toBe(1000000000);
+    expect(parsed.negative).toBe(-2500);
+    expect(parsed.fraction).toBe(12.3);
+  });
+
+  (contextSourceAvailable ? it : it.skip)("preserves precision of exponent notation values", async () => {
+    const data = `
+    {
+      "smallMagnitude": 1.5E-10,
+      "largeMagnitude": 1.2345678901234567890E30
+    }`;
+
+    const parsed = JSON.parse(data, jsonReviver);
+    expect(parsed.smallMagnitude).toEqual(new NumericValue("0.00000000015", "bigDecimal"));
+    expect(parsed.largeMagnitude).toBe(1234567890123456789000000000000n);
+  });
+
+  (contextSourceAvailable ? it : it.skip)("falls back to the parsed double for extreme exponents", async () => {
+    const parsed = JSON.parse(`{ "high": 1e5000, "low": 1e-5000 }`, jsonReviver);
+    expect(parsed.high).toBe(Infinity);
+    expect(parsed.low).toBe(0);
+  });
 });
--- a/packages-internal/core/src/submodules/protocols/json/jsonReviver.ts
+++ b/packages-internal/core/src/submodules/protocols/json/jsonReviver.ts
@@ -1,6 +1,56 @@
 import { NumericValue } from "@smithy/core/serde";
 
 /**
+ * Matches a JSON number in exponent notation, capturing the sign, integral
+ * digits, fractional digits, and exponent.
+ */
+const EXPONENT_NOTATION = /^(-?)(\d+)(?:\.(\d*))?[eE]([+-]?\d+)$/;
+
+/**
+ * Upper bound on the digit count of an expanded number, so that an extreme
+ * exponent cannot cause an enormous string allocation.
+ */
+const MAX_EXPANDED_DIGITS = 1024;
+
+/**
+ * Rewrites a JSON number in exponent notation to plain decimal notation, for
+ * example 1.5e3 to 1500.
+ *
+ * @param source - the original numeric string.
+ *
+ * @internal
+ *
+ * @returns the plain decimal form, or undefined when source is not exponent
+ * notation or when expanding it would exceed MAX_EXPANDED_DIGITS.
+ */
+function expandExponentNotation(source: string): string | undefined {
+  const match = EXPONENT_NOTATION.exec(source);
+  if (!match) {
+    return undefined;
+  }
+  const [, sign, integerDigits, fractionDigits = "", exponentDigits] = match;
+  const exponent = Number(exponentDigits);
+  const digits = integerDigits + fractionDigits;
+  if (digits.length + Math.abs(exponent) > MAX_EXPANDED_DIGITS) {
+    return undefined;
+  }
+  // Where the decimal point lands within `digits` once the exponent is applied.
+  const pointIndex = integerDigits.length + exponent;
+  let expanded: string;
+  if (pointIndex <= 0) {
+    expanded = "0." + "0".repeat(-pointIndex) + digits;
+  } else if (pointIndex >= digits.length) {
+    expanded = digits + "0".repeat(pointIndex - digits.length);
+  } else {
+    expanded = digits.slice(0, pointIndex) + "." + digits.slice(pointIndex);
+  }
+  if (expanded.includes(".")) {
+    expanded = expanded.replace(/0+$/, "").replace(/\.$/, "");
+  }
+  return sign + expanded.replace(/^0+(?=\d)/, "");
+}
+
+/**
  * @param key - JSON object key.
  * @param value - parsed value.
  * @param context - original JSON string for reference. Not available until Node.js 21 and unavailable in Safari as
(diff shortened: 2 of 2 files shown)


MohMaya:
Duplicate of PR #8226

Issue #6174, opened 2024-06-06
Create validation script for api-extractor.json 


kuhe:
### Describe the issue

On our API docs site, some symbols don't link to anything. This makes discovery very difficult unless you read the API through your actual code editor.

The validation should do the following:
- if any code object in a package has the `@public` typedoc annotation, the package should have an `api-extractor.json` metadata so its types can be published to the API doc website.

### Links

https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/Package/-smithy-smithy-client/Class/Command/

kuhe:
Context: we reduced the number of packages published to the API docs website to reduce the chances of users getting lost. 

Not all package exported symbols are considered public API. But those that are should be published to the site.

aBurmeseDev:
addressed in https://github.com/aws/aws-sdk-js-v3/pull/8242.

PR #8242, opened 2026-07-31
chore(scripts): add api-extractor.json validation for packages with public annotations

aBurmeseDev:
### Issue
https://github.com/aws/aws-sdk-js-v3/issues/6174

### Description
adds a validation that checks packages with `@public` typedoc annotations have correctly configured  `api-extractor.json`, so their types are published to the API docs.

scans all packages under `clients/`, `lib/`, `packages/`, and `packages-internal/` for `@public` annotations and for each package, validates:
- `api-extractor.json` exists
- `extends` points to the correct base config for its root
- `mainEntryPointFilePath` is set
- `package.json` has an `extract:docs` script

### Testing
added test case that creates a temp package with a `@public` but no `api-extractor.json` 

### Checklist
  - [ ] It's not a feature.
  - [ ] No streams here.

---

### DIFF
--- a/scripts/validation/api-extractor-config.js
+++ b/scripts/validation/api-extractor-config.js
@@ -1,0 +1,155 @@
+#!/usr/bin/env node
+
+// Usage: node api-extractor-config.js [packageDir ...]
+//
+//
+
+const fs = require("node:fs");
+const path = require("node:path");
+const { getPackageDirs, summarizePackages } = require("./validation-shared");
+
+const EXTENDS_BY_ROOT = {
+  clients: "../../api-extractor.json",
+  lib: "../../api-extractor.lib.json",
+  packages: "../../api-extractor.packages.json",
+  "packages-internal": "../../api-extractor.packages.json",
+};
+
+// Packages with @public symbols that lack api-extractor.json.
+const KNOWN_GAPS = new Set([
+  "@aws-sdk/credential-provider-http",
+  "@aws-sdk/credential-provider-ini",
+  "@aws-sdk/credential-provider-login",
+  "@aws-sdk/credential-provider-node",
+  "@aws-sdk/credential-provider-sso",
+  "@aws-sdk/credential-provider-web-identity",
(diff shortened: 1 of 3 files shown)


scripts/validation/api-extractor-config.js
```diff
@@ -0,0 +1,155 @@
...
+
+const EXTENDS_BY_ROOT = {
+  clients: "../../api-extractor.json",
+  lib: "../../api-extractor.lib.json",
+  packages: "../../api-extractor.packages.json",
+  "packages-internal": "../../api-extractor.packages.json",
+};
+
+// Packages with @public symbols that lack api-extractor.json.
+const KNOWN_GAPS = new Set([
```
aBurmeseDev, line 19: These have `@public` but no `api-extractor.json`. They'd warn instead of failing the build. The staleness check at the bottom ensures entries get removed once the gap is fixed. Open to adding/removing any of these if we feel differently about which packages should publish docs.

scripts/validation/api-extractor-config.js
```diff
@@ -0,0 +1,155 @@
...
+      errors.push(`${pkgJson.name} api-extractor.json is missing mainEntryPointFilePath`);
+    }
+
+    if (!pkgJson.scripts?.["extract:docs"]) {
+      errors.push(`${pkgJson.name} has @public symbols but no extract:docs script in package.json`);
+    }
+  }
+
+  // Only enforce on full scans, positional runs see a subset of packages.
+  if (isFullScan) {
```
aBurmeseDev, line 134: Packages like `ec2-metadata-service` or `signature-v4-multi-region` adds an `api-extractor.json`, it'd no longer show up as a gap during the scan.

kuhe (approved):
apply to smithy-typescript as well

aBurmeseDev:
> apply to smithy-typescript as well

https://github.com/smithy-lang/smithy-typescript/pull/2199

Issue #8246, opened 2026-08-04
Modeled number members deserialize to NumericValue when the response schema contains a Document (regression in @aws-sdk/core 3.977.0)


RasengunShotaro:
### Checkboxes for prior research

- [x] I've gone through [Developer Guide](https://docs.aws.amazon.com/sdk-for-javascript/v3/developer-guide) and [API reference](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest)
- [x] I've checked [AWS Forums](https://forums.aws.amazon.com) and [StackOverflow](https://stackoverflow.com/questions/tagged/aws-sdk-js).
- [x] I've searched for [previous similar issues](https://github.com/aws/aws-sdk-js-v3/issues) and didn't find any solution.

### Describe the bug

Members that are modeled as numbers (generated type `score?: number`) can deserialize to a `NumericValue` object instead of a `number`, whenever the response shape contains **any** `Document` member.

`needsReviver` returns `true` for a response schema that contains a document:

```js
// @aws-sdk/core/dist-cjs/submodules/protocols/index.js
else if (ns.isDocumentSchema()) {
    return true;
}
```

`RetrieveResponse` for `bedrock-agent-runtime` qualifies, because `RetrievalResultMetadata` is `Map<String, Document>`. Enabling the reviver applies it to the **whole body**, so `jsonReviver` also rewrites numeric literals belonging to modeled numeric members, and the schema-aware read does not coerce them back to `number`.

The result is silent data corruption rather than an error. `typeof result.score` is `"object"`, and `result.score + 1` evaluates to the string `"1.01"` instead of `2`. TypeScript cannot catch this, because the generated type declares `score?: number`, so a `typeof x === "number"` guard is reported as redundant and is likely to be removed as dead code.

Only literals that do not round-trip are affected (`1.0`, `0.0`, `1e3`, values beyond `Number.MAX_SAFE_INTEGER`), so this surfaces intermittently — a relevance score of `0.5611` is unaffected while a perfect match of `1.0` is not.

`Document` values deserializing to `NumericValue` seems intended, and is not what this report is about. The problem is that modeled numeric members are affected as collateral.

This is adjacent to #8224 (same origin: #8214, `@aws-sdk/core@3.977.0`) but distinct: #8224 was a thrown error for exponent notation, fixed in `3.977.1`. This one is silent and still reproduces on the latest version.

### Regression Issue

- [x] Select this option if this issue appears to be a regression.

[...]

PR #8247, opened 2026-08-04
fix(core): deserialize numeric members to number when the reviver is active

RasengunShotaro:
### Issue
#8246

### Description

`needsReviver` returns `true` when a response schema contains a bigInteger, bigDecimal or **document** member, which enables the `JSON.parse` reviver for the whole body. `jsonReviver` then rewrites every numeric literal that does not round-trip (`1.0`, `0.0`, `1e3`, anything past `Number.MAX_SAFE_INTEGER`), including literals that belong to plain numeric members, and the shape deserializer passed those values through unchanged.

The result is that a member whose generated type is `number` can deserialize to a `NumericValue`. It fails silently: `typeof result.score` becomes `"object"` and `result.score + 1` evaluates to the string `"1.01"` instead of `2`. TypeScript cannot catch it, because the generated type says `score?: number`, so a `typeof x === "number"` guard looks redundant and is a candidate for removal.

This affects any client whose response shape contains a document member. It was found on `bedrock-agent-runtime` `Retrieve`, where `RetrievalResultMetadata` is `Map<String, Document>`: the knowledge base returns page numbers as `22.0` inside `metadata`, and the relevance `score` in the same response was corrupted as collateral for exact `1.0` / `0.0` values.

Numeric members now coerce a revived `NumericValue` or `BigInt` back to `number`, which matches the generated types and the behaviour before `@aws-sdk/core@3.977.0`. Documents keep the revived representation, which is unchanged and appears intended.

Applied to both codecs (`JsonShapeDeserializer` and `JsonShapeDeserializer2`) so v1/v2 equivalence is preserved.

### Testing

Added a test to each codec spec asserting that a numeric member deserializes to `number` while the reviver is active (`1.0` → `1`, `0.0` → `0`, `1e3` → `1000`). Both fail without the change:

```
× JsonShapeDeserializer > deserializes numeric members to number when the reviver is active
  → expected { scalar: NumericValue{ …(2) } } to deeply equal { scalar: 1 }
```

Full protocols suite after the change: **32 files, 516 passed, 2 skipped**.

One existing expectation is updated. In `JsonShapeDeserializer2.spec.ts`, `fractional exponent with many digits does not throw in reviver` asserted that `{"scalar": 1.123456789012345678E16}` stays a `NumericValue`. `scalar` is a numeric member, and `JSON.parse` produced `11234567890123456` for that input before the reviver existed, so the expectation is changed to that value. The purpose of the test — that the reviver does not throw — is unaffected, and the neighbouring `v1/v2` equivalence cases still pass.

I could not run the repo's full `yarn` install locally, so verification was scoped to `packages-internal/core/src/submodules/protocols` with dependencies installed from npm.

### Checklist
  - [x] It's not a feature.
  - [x] No streams here.

---

### DIFF
--- a/packages-internal/core/src/submodules/protocols/json/codec-v1/JsonShapeDeserializer.spec.ts
+++ b/packages-internal/core/src/submodules/protocols/json/codec-v1/JsonShapeDeserializer.spec.ts
@@ -150,6 +150,15 @@ describe(JsonShapeDeserializer.name, () => {
     });
   });
 
+  it("deserializes numeric members to number when the reviver is active", async () => {
+    expect(await deserializer.read(widget, `{ "scalar": 1.0 }`)).toEqual({ scalar: 1 });
+    expect(await deserializer.read(widget, `{ "scalar": 0.0 }`)).toEqual({ scalar: 0 });
+    expect(await deserializer.read(widget, `{ "scalar": 1e3 }`)).toEqual({ scalar: 1000 });
+
+    const data = (await deserializer.read(widget, `{ "scalar": 1.0 }`)) as { scalar: unknown };
+    expect(typeof data.scalar).toBe("number");
+  });
+
   it("deserializes infinite and NaN numerics", async () => {
     expect(await deserializer.read(widget, JSON.stringify({ scalar: "Infinity" }))).toEqual({ scalar: Infinity });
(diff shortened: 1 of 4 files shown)


kuhe:
Thank you for reporting the issue and opening a PR. Because this change involves casting other data types to number, I went with https://github.com/aws/aws-sdk-js-v3/pull/8252 instead.

PR #8256, opened 2026-08-08
fix(lib-transfer-manager): optimize worker HTTP connections and file I/O

smilkuri:
### Description
Optimizes worker thread HTTP and I/O performance in the transfer manager by adding dns spread lookup, configuring `keepAlive: true` to reuse TCP connections and reuse the fd for all part reads.

### Testing
CI

### Checklist
  - [x] It's not a feature.
  - [x] No streams here.

---

### DIFF
--- a/lib/lib-transfer-manager/src/submodules/worker/http-request-worker.ts
+++ b/lib/lib-transfer-manager/src/submodules/worker/http-request-worker.ts
@@ -7,10 +7,48 @@
  */
 import { HttpRequest } from "@smithy/core/protocols";
 import { NodeHttpHandler } from "@smithy/node-http-handler";
-import { open } from "node:fs/promises";
+import dns from "node:dns";
+import { openSync, readSync, closeSync } from "node:fs";
 import { Agent as httpsAgent } from "node:https";
 import { parentPort } from "node:worker_threads";
 import { crc32 } from "node:zlib";
+
+/**
+ * Resolves all A records for a hostname, picks one at random per connection
+ * to spread load across S3's fleet. Falls back to dns.lookup for IPv6 or
+ * on resolution failure.
+ * @internal
+ */
+const DNS_TTL_MS = 1000;
+const dnsCache = new Map<string, { ips: string[]; ts: number }>();
+
+function spreadLookup(hostname: string, options: any, callback: any): void {
+  if (options && options.family === 6) {
+    return dns.lookup(hostname, options, callback);
+  }
+
+  const deliver = (ips: string[]) => {
+    const ip = ips[Math.floor(Math.random() * ips.length)];
+    if (options && options.all) {
+      callback(null, [{ address: ip, family: 4 }]);
+    } else {
+      callback(null, ip, 4);
+    }
+  };
+
+  const cached = dnsCache.get(hostname);
+  if (cached && cached.ips.length && Date.now() - cached.ts < DNS_TTL_MS) {
+    return deliver(cached.ips);
+  }
+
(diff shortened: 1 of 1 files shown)


lib/lib-transfer-manager/src/submodules/worker/http-request-worker.ts
```diff
@@ -7,11 +7,49 @@
...
 import { parentPort } from "node:worker_threads";
 import { crc32 } from "node:zlib";
 
+/**
+ * Resolves all A records for a hostname, picks one at random per connection
+ * to spread load across S3's fleet. Falls back to dns.lookup for IPv6 or
+ * on resolution failure.
+ * @internal
+ */
+const DNS_TTL_MS = 1000;
```
kuhe: tsdoc description placed on wrong symbol

lib/lib-transfer-manager/src/submodules/worker/http-request-worker.ts
```diff
@@ -7,11 +7,49 @@
...
+const dnsCache = new Map<string, { ips: string[]; ts: number }>();
+
+function spreadLookup(hostname: string, options: any, callback: any): void {
```
kuhe: why are these types `any`?

lib/lib-transfer-manager/src/submodules/worker/http-request-worker.ts
```diff
@@ -7,11 +7,49 @@
...
+    }
+  };
+
+  const cached = dnsCache.get(hostname);
+  if (cached && cached.ips.length && Date.now() - cached.ts < DNS_TTL_MS) {
+    return deliver(cached.ips);
+  }
+
+  dns.resolve4(hostname, (err, ips) => {
+    if (err || !ips || !ips.length) {
```
kuhe:
```js
err || !ips?.length
```

kuhe: is this equivalent?

lib/lib-transfer-manager/src/submodules/worker/http-request-worker.ts
```diff
@@ -195,6 +245,13 @@ if (parentPort) {
 
   port.on("message", (msg: HttpWorkerInboundMessage) => {
     if (msg.type === "done") {
+      for (const fd of fdCache.values()) {
+        try {
+          closeSync(fd);
+        } catch {
+          /* ignore */
```
kuhe: what could have happened to throw here and why is it safe to ignore?

smilkuri: process.exit(0) below will close everything anyways, so safe to ignore. If we throw here it wouldn't destroy the handler.

lib/lib-transfer-manager/src/submodules/worker/http-request-worker.ts
smilkuri: Yes

lib/lib-transfer-manager/src/submodules/worker/http-request-worker.ts
smilkuri: Will update the types.

lib/lib-transfer-manager/src/submodules/worker/http-request-worker.ts
kuhe: comment that here

kuhe (approved):

PR #8259, opened 2026-08-10
feat(lib-transfer-manager): add file based download api and worker thread based download.

smilkuri:
### Issue 
Internal JS-7054

### Description
- Adds `downloadToFile` API which takes a destination file path and downloads an S3 object to that location. It supports both PART based and RANGE based strategy. It uses a temp file with atomic rename, the destination file only appears once the download complete successfully. On failure or abort, the temp file is cleaned up.
- Adds `downloadByPartWithWorkers` and `downloadByRangeWithWorkers` for the `download()` API. When `workerThreadCount > 1`, download requests are routed through these methods based on the configured multipart strategy. Internally, Part 1 is fetched on the main thread to discover object metadata(PartsCount, ETag, total size). Remaining parts/range requests are dispatched to worker threads via WorkerHttpHandler, which assembles response bytes into dedicated ArrayBuffers and transfers ownership back to the main thread (zero-copy). An `OrderedPartQueue` reorders parts that arrive out-of-order, while a Readable stream pulls from the queue sequentially and is passed to joinStreams to deliver a single ordered body to the caller.

### Testing
How was this change tested?
```
$ yarn test
 ✓ src/submodules/transfer-manager/join-streams.spec.ts (10 tests) 578ms
     ✓ should handle backpressure when consumer is slow  554ms
 ✓ src/submodules/transfer-manager/S3TransferManager.spec.ts (101 tests) 377ms

 Test Files  2 passed (2)
      Tests  111 passed (111)
   Start at  14:39:19
   Duration  952ms (transform 418ms, setup 0ms, import 663ms, tests 954ms, environment 0ms)
   ```
 ```
   $ yarn test:e2e
    Test Files  1 passed (1)
      Tests  39 passed (39)
   Start at  14:40:53
   Duration  37.66s (transform 412ms, setup 0ms, import 616ms, tests 36.90s, environment 0ms)
```
   
### Checklist
  - [ ] It's not a feature.
  - [ ] No streams here.

---

### DIFF
--- a/lib/lib-transfer-manager/src/submodules/transfer-manager/S3TransferManager.ts
+++ b/lib/lib-transfer-manager/src/submodules/transfer-manager/S3TransferManager.ts
@@ -1,5 +1,6 @@
 import type {
   AbortMultipartUploadCommandInput,
+  ChecksumAlgorithm,
   CompletedPart,
   CompleteMultipartUploadCommandInput,
   CompleteMultipartUploadCommandOutput,
@@ -399,54 +400,35 @@ export class S3TransferManager implements IS3TransferManager {
       }
     };
 
+    const resolvedTransferOptions = transferOptions ?? {};
+    const downloadArgs = [
+      request,
+      resolvedTransferOptions,
+      streams,
+      requests,
+      metadata,
+      checksumValidationEnabled,
+    ] as const;
+
     try {
+      let responseMetadata: { totalSize: number | undefined; onStreamConsumed?: (index: number) => void };
+
       if (this.multipartDownloadType === "PART") {
         if (this.workerHttpHandler) {
-          const responseMetadata = await this.downloadByPartWithWorkers(
-            request,
-            transferOptions ?? {},
-            streams,
-            requests,
-            metadata,
-            checksumValidationEnabled
-          );
-          totalSize = responseMetadata.totalSize;
+          responseMetadata = await this.downloadByPartWithWorkers(...downloadArgs);
         } else {
-          const responseMetadata = await this.downloadByPart(
-            request,
-            transferOptions ?? {},
-            streams,
-            requests,
-            metadata,
-            checksumValidationEnabled
-          );
-          totalSize = responseMetadata.totalSize;
-          onStreamConsumed = responseMetadata.onStreamConsumed;
+          responseMetadata = await this.downloadByPart(...downloadArgs);
         }
       } else if (this.multipartDownloadType === "RANGE") {
         if (this.workerHttpHandler && !request.Range) {
-          const responseMetadata = await this.downloadByRangeWithWorkers(
-            request,
-            transferOptions ?? {},
-            streams,
-            requests,
-            metadata,
-            checksumValidationEnabled
-          );
-          totalSize = responseMetadata.totalSize;
+          responseMetadata = await this.downloadByRangeWithWorkers(...downloadArgs);
         } else {
-          const responseMetadata = await this.downloadByRange(
-            request,
-            transferOptions ?? {},
-            streams,
-            requests,
-            metadata,
-            checksumValidationEnabled
-          );
-          totalSize = responseMetadata.totalSize;
-          onStreamConsumed = responseMetadata.onStreamConsumed;
+          responseMetadata = await this.downloadByRange(...downloadArgs);
         }
       }
+
+      totalSize = responseMetadata!.totalSize;
+      onStreamConsumed = responseMetadata!.onStreamConsumed;
     } catch (error) {
       // On failure or abort, response bodies that were already received but not
       // yet handed to joinStreams are orphaned. Destroy them (attaching a no-op
@@ -1717,6 +1699,9 @@ export class S3TransferManager implements IS3TransferManager {
     };
 
     const releaseSlot = (): void => {
+      if (inFlight <= 0) {
+        throw new Error("releaseSlot called with no in-flight requests");
+      }
       inFlight--;
       if (nextDispatchResolve && inFlight < this.maxConcurrentDownloads) {
         const resolve = nextDispatchResolve;
@@ -1724,9 +1709,6 @@ export class S3TransferManager implements IS3TransferManager {
         resolve();
       }
     };
-
-    // Dispatch part GET requests via the SDK pipeline (WorkerHttpHandler intercepts them).
-    // Concurrency limiter provides backpressure — waitForSlot blocks when all slots are in use.
     const dispatchPromise = (async () => {
       for (let partNumber = 2; partNumber <= partsCount; partNumber++) {
         if (abortController.signal.aborted || queue.hasError()) break;
@@ -1772,7 +1754,6 @@ export class S3TransferManager implements IS3TransferManager {
             if (!transferResult) {
               queue.setError(new Error(`Missing download result for part ${partNumber}`));
               abortController.abort();
-              releaseSlot();
               return;
             }
 
@@ -1790,11 +1771,12 @@ export class S3TransferManager implements IS3TransferManager {
                 },
               })
             );
-            releaseSlot();
           })
           .catch((error) => {
             queue.setError(error);
             abortController.abort();
+          })
+          .finally(() => {
             releaseSlot();
           });
       }
@@ -1808,11 +1790,15 @@ export class S3TransferManager implements IS3TransferManager {
 
     // Create a Readable stream that pulls parts sequentially from the ordered queue.
     // Each chunk is a zero-copy Buffer view into the transferred ArrayBuffer.
+    let reading = false;
     const transferStream = new Readable({
       read() {
+        if (reading) return;
+        reading = true;
         queue
           .dequeue()
           .then((item) => {
+            reading = false;
             if (item === null) {
               // null means either all parts delivered OR an error was set.
               // If an error was set, destroy the stream with that error so
@@ -1831,6 +1817,7 @@ export class S3TransferManager implements IS3TransferManager {
             this.push(buf);
           })
           .catch((err) => {
+            reading = false;
             this.destroy(err as Error);
           });
       },
@@ -1936,6 +1923,9 @@ export class S3TransferManager implements IS3TransferManager {
     };
 
     const releaseSlot = (): void => {
+      if (inFlight <= 0) {
+        throw new Error("releaseSlot called with no in-flight requests");
+      }
       inFlight--;
       if (nextDispatchResolve && inFlight < this.maxConcurrentDownloads) {
         const resolve = nextDispatchResolve;
@@ -1994,7 +1984,6 @@ export class S3TransferManager implements IS3TransferManager {
             if (!transferResult) {
               queue.setError(new Error(`Missing download result for range ${start}-${end}`));
               abortController.abort();
-              releaseSlot();
               return;
             }
 
@@ -2012,11 +2001,12 @@ export class S3TransferManager implements IS3TransferManager {
                 },
               })
             );
-            releaseSlot();
           })
           .catch((error) => {
             queue.setError(error);
             abortController.abort();
+          })
+          .finally(() => {
             releaseSlot();
           });
       }
@@ -2028,11 +2018,15 @@ export class S3TransferManager implements IS3TransferManager {
 
     // Create a Readable stream that pulls ranges sequentially from the ordered queue.
     // Each chunk is a zero-copy Buffer view into the transferred ArrayBuffer.
+    let reading = false;
     const transferStream = new Readable({
       read() {
+        if (reading) return;
+        reading = true;
         queue
           .dequeue()
           .then((item) => {
+            reading = false;
             if (item === null) {
               if (queue.hasError()) {
                 this.destroy(queue.getError() as Error);
@@ -2048,6 +2042,7 @@ export class S3TransferManager implements IS3TransferManager {
             this.push(buf);
           })
           .catch((err) => {
+            reading = false;
             this.destroy(err as Error);
           });
       },
@@ -2472,7 +2467,7 @@ export class S3TransferManager implements IS3TransferManager {
     const { partSize } = this.calculatePartSize(contentLength);
     const filePath = (request.Body as any).path as string;
 
-    const buildDataSource = (checksumAlgorithm?: string, checksumHeader?: string): DataSource => ({
+    const buildDataSource = (checksumAlgorithm?: ChecksumAlgorithm, checksumHeader?: string): DataSource => ({
       type: "file",
       filePath,
       partSize,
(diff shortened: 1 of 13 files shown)


lib/lib-transfer-manager/src/submodules/transfer-manager/file-manager.ts
```diff
@@ -0,0 +1,136 @@
...
+   */
+  public async createTempFile(destination: string, totalSize: number): Promise<string> {
+    const tempFilePath = `${destination}.s3tmp.${this.generateUniqueId()}`;
+
+    // "wx" creates the file exclusively, so a name collision with a concurrent
+    // download to the same destination fails with EEXIST rather than truncating
+    // the other download's temp file.
+    const fd = await open(tempFilePath, "wx");
+    try {
+      await fd.truncate(totalSize);
```
smilkuri, line 31: Specification says not to pre-validate available space, we are just setting the logical file length here so it should be okay ig.

lib/lib-transfer-manager/src/submodules/transfer-manager/ordered-part-queue.ts
```diff
@@ -0,0 +1,134 @@
+/**
+ * Ordered delivery queue for stream downloads using ArrayBuffer ownership transfer.
+ *
+ * Workers download parts in parallel and transfer ArrayBuffers (zero copy) to main in
+ * arbitrary order. This queue holds out-of-order arrivals and delivers them
+ * sequentially to the Readable stream consumer.
+ *
+ * @internal
+ */
+
```
kuhe, line 2: this should not be in tsdoc format if it's module level, use `/* */` or `//`.

or attach it to the main export symbol

lib/lib-transfer-manager/src/submodules/transfer-manager/ordered-part-queue.ts
```diff
@@ -0,0 +1,134 @@
...
+   * The transferred ArrayBuffer containing the part data.
+   */
+  buffer: ArrayBuffer;
+  /**
+   * Actual bytes written (may be less than buffer.byteLength for the last part).
+   */
+  byteLength: number;
+}
+
+export class OrderedPartQueue {
```
kuhe, line 20: what does this do? ~2 sentences as class description

lib/lib-transfer-manager/src/submodules/transfer-manager/ordered-part-queue.ts
```diff
@@ -0,0 +1,134 @@
...
+
+    if (this.waitingConsumer) {
+      const consumer = this.waitingConsumer;
+      this.waitingConsumer = null;
+      consumer(null);
+    }
+  }
+
+  /**
+   * Returns whether an error has been set.
```
kuhe: I can tell from the code this is what is happening. Who is the caller and why do they need to know?

lib/lib-transfer-manager/src/submodules/transfer-manager/S3TransferManager.ts
```diff
@@ -171,7 +183,9 @@ export class S3TransferManager implements IS3TransferManager {
       };
 
       this.abortCleanupFunctions.set(signal, () => signal.removeEventListener("abort", removeListenerAfterAbort));
-      signal.addEventListener("abort", removeListenerAfterAbort, { once: true });
+      signal.addEventListener("abort", removeListenerAfterAbort, {
+        once: true,
+      });
```
kuhe, line 189: why is this reformatted?

lib/lib-transfer-manager/src/submodules/transfer-manager/S3TransferManager.ts
```diff
@@ -384,27 +401,51 @@ export class S3TransferManager implements IS3TransferManager {
...
+          );
+          totalSize = responseMetadata.totalSize;
+        } else {
+          const responseMetadata = await this.downloadByPart(
+            request,
+            transferOptions ?? {},
+            streams,
+            requests,
+            metadata,
+            checksumValidationEnabled
```
kuhe: are these parameters identical? prebuild in a variable

lib/lib-transfer-manager/src/submodules/transfer-manager/types.ts
```diff
@@ -106,6 +106,38 @@ export type DownloadRequest = GetObjectCommandInput;
...
+   * The local file path to which the response body will be written.
+   * Must be a non-empty string.
+   */
+  destination: string;
+  /**
+   * Whether to fail the request if a file already exists at the destination.
+   * When false, the existing file is overwritten.
+   * Defaults to false.
+   */
+  failIfExists?: boolean;
```
kuhe, line 126: this parameter is usually called `overwrite: boolean = true`. Spec driven?

lib/lib-transfer-manager/src/submodules/transfer-manager/S3TransferManager.ts
```diff
@@ -980,10 +2100,66 @@ export class S3TransferManager implements IS3TransferManager {
...
+    const metadata: Record<string, any> = {};
+    this.assignMetadata(metadata, initialResponse);
+    metadata.ContentLength = totalSize;
+    metadata.ContentRange = `bytes 0-${totalSize - 1}/${totalSize}`;
+
+    if (initialResponse.ChecksumType === "COMPOSITE") {
+      metadata.ChecksumCRC32 = undefined;
+      metadata.ChecksumCRC32C = undefined;
+      metadata.ChecksumSHA1 = undefined;
+      metadata.ChecksumSHA256 = undefined;
```
kuhe, line 2117: why do this? there are other checksums now too

lib/lib-transfer-manager/src/submodules/transfer-manager/S3TransferManager.ts
```diff
@@ -941,9 +1622,452 @@ export class S3TransferManager implements IS3TransferManager {
...
+      if (inFlight < this.maxConcurrentDownloads) {
+        return Promise.resolve();
+      }
+      return new Promise<void>((resolve) => {
+        nextDispatchResolve = resolve;
+      });
+    };
+
+    const releaseSlot = (): void => {
+      inFlight--;
```
kuhe, line 1705: should this throw if already at 0?

or, the decrement should be in the conditional with the work being done

lib/lib-transfer-manager/src/submodules/transfer-manager/S3TransferManager.ts
```diff
@@ -941,9 +1622,452 @@ export class S3TransferManager implements IS3TransferManager {
...
+                  totalBytes: totalSize,
+                },
+              })
+            );
+            releaseSlot();
+          })
+          .catch((error) => {
+            queue.setError(error);
+            abortController.abort();
+            releaseSlot();
```
kuhe, line 1780: move releaseslot to finally?

lib/lib-transfer-manager/src/submodules/transfer-manager/S3TransferManager.ts
```diff
@@ -941,9 +1622,452 @@ export class S3TransferManager implements IS3TransferManager {
...
+      // Signal the queue so the consumer stream surfaces the error.
+      if (!queue.hasError()) {
+        queue.setError(err);
+      }
+    });
+
+    // Create a Readable stream that pulls parts sequentially from the ordered queue.
+    // Each chunk is a zero-copy Buffer view into the transferred ArrayBuffer.
+    const transferStream = new Readable({
+      read() {
```
kuhe, line 1795: is this read method fine to call if it's still working?

lib/lib-transfer-manager/src/submodules/transfer-manager/S3TransferManager.ts
```diff
@@ -941,9 +1622,452 @@ export class S3TransferManager implements IS3TransferManager {
...
+
+    // Assign metadata from HeadObject response
+    Object.assign(metadata, {
+      ContentLength: totalSize,
+      ContentRange: `bytes 0-${totalSize - 1}/${totalSize}`,
+      ContentType: headResponse.ContentType,
+      ETag: headResponse.ETag,
+      LastModified: headResponse.LastModified,
+      $metadata: headResponse.$metadata,
+    });
```
kuhe, line 1899: do you have full ownership of the metadata object at this point where it is mutating?

lib/lib-transfer-manager/src/submodules/transfer-manager/worker-http-handler.ts
```diff
@@ -144,12 +245,113 @@ export interface WorkerHttpHandlerOptions extends HttpHandlerOptions {
...
+  offset: number;
+  /**
+   * Expected byte length of the part body. Used only as a fallback, for the same
+   * reason as {@link DownloadDataToFile.offset}.
+   */
+  expectedLength: number;
+  /**
+   * Optional CRC algorithm for inline checksum computation.
+   */
+  checksumAlgorithm?: string; // "CRC32" | "CRC32C" | "CRC64NVME"
```
kuhe: why only 3 options here?

lib/lib-transfer-manager/src/submodules/worker/http-request-worker.ts
```diff
@@ -95,23 +108,129 @@ interface HttpWorkerRAMRequestMessage extends BaseHttpWorkerRequestMessage {
...
+   */
+  expectedSize: number;
+  /**
+   * Sequential part/range index for ordered delivery on the main thread.
+   */
+  rangeIndex: number;
+  /**
+   * Optional CRC algorithm for inline checksum validation against S3 response headers.
+   */
+  checksumAlgorithm?: string;
```
kuhe: why not the ChecksumAlgorithm enum type from S3?

lib/lib-transfer-manager/src/submodules/worker/http-request-worker.ts
```diff
@@ -3,13 +3,21 @@
...
+ *
  * @internal
  */
 import { HttpRequest } from "@smithy/core/protocols";
 import { NodeHttpHandler } from "@smithy/node-http-handler";
 import dns from "node:dns";
 import { openSync, readSync, closeSync } from "node:fs";
 import type { LookupOptions } from "node:dns";
+import type { Checksum } from "@smithy/types";
+import { Crc32cJs, Crc64NvmeJs } from "@aws-sdk/checksums/crc";
```
kuhe: why import the JS implementations of the checksums and not the default?

lib/lib-transfer-manager/src/submodules/worker/http-request-worker.ts
```diff
@@ -130,6 +252,74 @@ function toBase64(n: number): string {
...
+    return new Crc32cJs();
+  }
+  // CRC32 — use node:zlib wrapper that conforms to Checksum interface
+  return new NodeCrc32();
+}
+
+/**
+ * Thin wrapper around node:zlib crc32 to conform to the Checksum interface.
+ */
+class NodeCrc32 implements Checksum {
```
kuhe: why is there a new checksum implementation here?

lib/lib-transfer-manager/src/submodules/transfer-manager/ordered-part-queue.ts
smilkuri: It reorders the parts and deliver them sequentially. Will update it.

lib/lib-transfer-manager/src/submodules/transfer-manager/ordered-part-queue.ts
smilkuri: Dispatch loop calls it to stop launching new part requests in case of failure.

lib/lib-transfer-manager/src/submodules/transfer-manager/S3TransferManager.ts
smilkuri: Formatted when ran `make format`

lib/lib-transfer-manager/src/submodules/transfer-manager/S3TransferManager.ts
smilkuri: Yes

lib/lib-transfer-manager/src/submodules/transfer-manager/types.ts
smilkuri: That's according to the specification.

lib/lib-transfer-manager/src/submodules/transfer-manager/S3TransferManager.ts
smilkuri: The reason for setting them to undefined is  s3 returns the checksum value for the whole multipart object (a composite of per-part checksums), not for the byte range being returned. Since we reassemble parts into a single stream, the per-part checksum from the initial response doesn't match to the joined output. But I agree we have to add that new checksums too.

lib/lib-transfer-manager/src/submodules/transfer-manager/S3TransferManager.ts
smilkuri: Yes will add a check here

lib/lib-transfer-manager/src/submodules/transfer-manager/S3TransferManager.ts
smilkuri: No we shouldn't. The queue only has one waitingConsumer slot, so a second `dequeue` call would overwrite the first one. Should have some kind of reading check here.

lib/lib-transfer-manager/src/submodules/transfer-manager/S3TransferManager.ts
smilkuri: Yes, we are creating metadata as empty object initially in `download()` right, nobody reads it before it reaches the download method, it is populated here. Instead should we return as a new object from the download method instead of mutating the parameter?

lib/lib-transfer-manager/src/submodules/transfer-manager/worker-http-handler.ts
smilkuri: Initially didn't add the new XXHASH algorithms because the SDK doesn't have local implementations. Now updated it to just skip validation for those.

lib/lib-transfer-manager/src/submodules/worker/http-request-worker.ts
smilkuri: Updated.

lib/lib-transfer-manager/src/submodules/transfer-manager/S3TransferManager.ts
kuhe: you can mutate if you have ownership of the object and its reference does not escape to the caller

lib/lib-transfer-manager/src/submodules/transfer-manager/worker-http-handler.ts
kuhe: it's not just XXHASH, there's sha256 & sha512, right?

lib/lib-transfer-manager/src/submodules/worker/http-request-worker.ts
```diff
@@ -246,47 +248,25 @@ interface HttpWorkerReadyMessage {
...
+    case "CRC32":
+      return new Crc32();
+    case "SHA1":
+      return new Sha1();
+    case "SHA256":
+      return new Sha256();
+    case "MD5":
+      return new Md5();
+    default:
+      // Algorithms without a local implementation (e.g. XXHASH64, SHA512) skip inline validation.
```
kuhe, line 268: unrecognized checksums should be supported, the client extension may introduce a checksum and it should be used if provided and specified by the caller

kuhe (approved):

PR #8270, opened 2026-08-18
feat(dynamic-client): create untyped clients from Smithy JSON AST

kuhe:
### Issue
n/a, experimental

### Description
Allows creation of a sigv4 dynamic client from Smithy JSON AST

### Testing
e2e using a handful of AWS services of various protocols.

### Checklist
  - [ ] It's not a feature.
  - [x] No streams here.

---


packages-internal/dynamic-client/src/endpoint/buildEndpointProvider.ts
```diff
@@ -0,0 +1,60 @@
...
+ * @returns an endpoint provider.
+ *
+ * @internal
+ */
+export function buildEndpointProvider(service: AstShape): EndpointProvider {
+  const traits = service.traits ?? {};
+
+  const ruleSet = traits["smithy.rules#endpointRuleSet"] as RuleSetObject | undefined;
+  if (ruleSet) {
+    const cache = new EndpointCache({ size: 50, params: ["endpoint"] });
```
trivikr: Use all declared endpoint ruleset parameters
```suggestion
    const cache = new EndpointCache({ size: 50, params: Object.keys(ruleSet.parameters) });
```

A regression test that resolves the same ruleset for two regions will demonstrate current cache collision.

packages-internal/dynamic-client/src/endpoint/buildEndpointProvider.ts
```diff
@@ -0,0 +1,60 @@
...
+  }
+
+  // TODO: the dynamic client should emplace a default endpointRuleSet when a
+  // model has none, rather than falling back to this passthrough provider.
+  // Mirror smithy-typescript's AddDefaultEndpointRuleSet (basic SDK::Endpoint
+  // ruleset) and AddDefaultAwsEndpointRuleSet (regional AWS flavor keyed on the
+  // service's endpointPrefix) so endpoint resolution always runs through the
+  // rules engine and honors the caller-supplied `Endpoint` builtin uniformly.
+  return (params) => {
+    const endpoint = params?.endpoint;
```
trivikr: Prefer the modeled parameter, if provided
```suggestion
    const endpoint = params?.Endpoint ?? params?.endpoint;
```

Regression test
```js
    const provider = buildEndpointProvider({ type: "service" });
    const resolved = provider({ Endpoint: "https://example.com" });
    expect(resolved.url.href).toBe("https://example.com/");
```

packages-internal/dynamic-client/src/schema/SchemaBuilder.ts
```diff
@@ -0,0 +1,440 @@
...
+    }
+    return [4, namespace, name, traits, names, refs];
+  }
+
+  /**
+   * @returns the schema ref for an aggregate member, wrapping it as a
+   *   `[ref, memberTraits]` pair only when the member carries its own traits.
+   */
+  private memberRef(member: AstMember): $SchemaRef {
+    const targetRef = this.ref(member.target);
```
trivikr, line 315: A member-level `timestampFormat` should override the target timestamp schema, while preserving any other member traits.

```suggestion
    let targetRef = this.ref(member.target);
    const targetShape = this.index.getShape(member.target);
    const timestampFormat = member.traits?.["smithy.api#timestampFormat"];

    if (
      timestampFormat !== undefined &&
      (member.target === "smithy.api#Timestamp" || targetShape?.type === "timestamp")
    ) {
      const timestampRef = sentinelForSimpleShape({ type: "timestamp", traits: member.traits })!;
      const targetTraits = targetShape ? this.index.getTraits(targetShape) : 0;
      if (targetShape && targetTraits !== 0) {
        const { namespace, name } = parseShapeId(member.target);
        targetRef = [0, namespace, name, targetTraits, timestampRef] as unknown as $SchemaRef;
      } else {
        targetRef = timestampRef as $SchemaRef;
      }
    }
```

A regression test can model all three Smithy timestamp formats at the member level and assert that each member reference uses the corresponding timestamp sentinel.

packages-internal/dynamic-client/src/schema/SchemaBuilder.ts
```diff
@@ -0,0 +1,440 @@
...
+      return undefined;
+    }
+    return simple;
+  }
+
+  /**
+   * Ensures a named static schema exists for the given shape, building it once.
+   */
+  private ensureNamed(shapeId: string, shape: AstShape): void {
+    const symbol = symbolOf(shapeId);
```
trivikr: Can internal schema storage use the full shape ID? `symbolOf` drops the namespace, so shapes such as `a#Shared` and `b#Shared` alias to whichever is built first.

A regression test can reference both from one input and assert that each thunk resolves to its namespace-qualified schema.

packages-internal/dynamic-client/src/schema/SchemaBuilder.ts
```diff
@@ -0,0 +1,440 @@
...
+    return baseException;
+  }
+
+  /**
+   * Registers an error schema in its namespace's registry, tracking the
+   * registry for later protocol error resolution.
+   */
+  private registerError(namespace: string, error: StaticErrorSchema): void {
+    const registry = TypeRegistry.for(namespace);
+    registry.registerError(error, ServiceException);
```
trivikr: Preserve modeled error names
```suggestion
    const exceptionName = error[2];
    const traits = error[3];
    const fault = typeof traits === "object" ? traits.error : undefined;
    const ModeledException = class extends ServiceException {
      public constructor(options: ConstructorParameters<typeof ServiceException>[0]) {
        super({
          ...options,
          name: exceptionName,
          ...(fault ? { $fault: fault } : {}),
        } as ConstructorParameters<typeof ServiceException>[0]);
        Object.setPrototypeOf(this, ModeledException.prototype);
      }
    };
    registry.registerError(error, ModeledException);
```

trivikr (changes requested):

trivikr (approved):

PR #8271, opened 2026-08-18
chore(build): typescript 7 multi-build mode

kuhe:
### Description
Using TS7's `--build` mode from https://devblogs.microsoft.com/typescript/announcing-typescript-7-0/, multiple packages can be built in the same memory/process context, speeding things up.


clients/client-accessanalyzer/tsconfig.build.json
```diff
@@ -0,0 +1,30 @@
+{
+  "extends": "./tsconfig",
+  "compilerOptions": {
+    "composite": true,
+    "outDir": "dist-es",
+    "rootDir": "src",
+    "declaration": true,
+    "declarationDir": "dist-types",
+    "tsBuildInfoFile": "dist-types/.tsbuildinfo",
```
kuhe: - [x] this should be moved into a central location and take on the package name, for easy deletion and avoiding accidental publishing of the buildinfo file

PR #8274, opened 2026-08-24
feat(lib-transfer-manager): add download directory functionality

smilkuri:
### Issue
Internal JS-7054

### Description
Implements the S3 Transfer Manager's `downloadDirectory` API.  It downloads all objects from a given S3 bucket under a given prefix to a local directory, with optional filtering and per-request customization. 

### Testing
Added unit tests and e2e tests

### Checklist
  - [x] It's not a feature.
  - [x] No streams here.

---


Akashbayzid (approved):

lib/lib-transfer-manager/src/submodules/transfer-manager/S3TransferManager.ts
```diff
@@ -1294,11 +1304,228 @@ export class S3TransferManager implements IS3TransferManager {
...
+    const failurePolicy = request.failurePolicy ?? ("terminate" as CannedFailurePolicy);
+    const semaphore = new Semaphore(maxConcurrency);
+    const abortController = new AbortController();
+    const inFlight = new Set<Promise<void>>();
+
+    let objectsDownloaded = 0;
+    let objectsFailed = 0;
+    let terminated = false;
+    let terminationError: unknown;
+
```
trivikr, line 1335: User cancellation currently only aborts the internal controller. When an in-flight `GetObject` then rejects, the task catch sees `abortController.signal.aborted`, increments `objectsFailed`, and returns without setting `terminated` or `terminationError`. The final branch therefore emits `transferComplete` and resolves even though the caller canceled the operation.

We should make user abort terminal and preserve the established `AbortError` contract.

```diff
+    const userSignal = transferOptions?.abortSignal as AbortSignal | undefined;
+    const onUserAbort = () => {
+      if (!terminated) {
+        terminated = true;
+        terminationError = Object.assign(new Error("Transfer aborted."), {
+          name: "AbortError",
+        });
+      }
+      abortController.abort();
+    };
+
     this.checkAborted(transferOptions);
     this.addEventListeners(transferOptions?.eventListeners);

     const removeLocalEventListeners = () => {
       this.removeEventListeners(transferOptions?.eventListeners);
+      userSignal?.removeEventListener("abort", onUserAbort);
       // existing abortCleanupFunctions cleanup
     };

-    const userSignal = transferOptions?.abortSignal as AbortSignal | undefined;
-    if (userSignal) {
-      userSignal.addEventListener("abort", () => abortController.abort(), {
-        once: true,
-      });
-    }
+    userSignal?.addEventListener("abort", onUserAbort, { once: true });
```

A regression test can use an abort-aware mock instead of a timer: return one listed object, make mocked `GetObject` stay pending after installing a listener on `options.abortSignal`, expose a promise indicating that `GetObject` started, then abort the user controller and assert rejection:

```ts
const download = tm.downloadDirectory(
  { bucket: "example-bucket", destination },
  { abortSignal: controller.signal }
);

await getObjectStartedPromise;
controller.abort();

await expect(download).rejects.toMatchObject({ name: "AbortError" });
```

The mock abort listener should reject the pending `GetObject` with an `AbortError`; on the current implementation this assertion instead observes a resolved `{ objectsDownloaded: 0, objectsFailed: 1 }` response.

lib/lib-transfer-manager/src/submodules/transfer-manager/S3TransferManager.ts
```diff
@@ -1294,11 +1304,228 @@ export class S3TransferManager implements IS3TransferManager {
...
+            let objectRequest: GetObjectCommandInput = {
+              Bucket: request.bucket,
+              Key: key,
+            };
+            try {
+              // Derive and validate the local destination path.
+              const localPath = deriveLocalPath(destination, key);
+
+              // Ensure all intermediate sub directories exist before writing.
+              await mkdir(dirname(localPath), { recursive: true });
```
trivikr: `deriveLocalPath` only verifies lexical containment, so an existing descendant symlink can redirect this write outside `destination`. Consider creating each parent component and checking its canonical location before proceeding:

```diff
+export async function createLocalParentDirectories(destination: string, localPath: string): Promise<void> {
+  const canonicalDestination = await realpath(destination);
+  const relativeParent = relative(destination, dirname(localPath));
+  let currentDirectory = destination;
+
+  for (const segment of relativeParent.split(sep)) {
+    if (!segment || segment === ".") continue;
+
+    currentDirectory = join(currentDirectory, segment);
+    await mkdir(currentDirectory, { recursive: true });
+
+    const canonicalDirectory = await realpath(currentDirectory);
+    const canonicalRelative = relative(canonicalDestination, canonicalDirectory);
+    if (canonicalRelative === ".." || canonicalRelative.startsWith(`..${sep}`) || isAbsolute(canonicalRelative)) {
+      throw new Error("Symbolic links outside the destination are not allowed.");
+    }
+  }
+}
```

Then replace this call with `await createLocalParentDirectories(destination, localPath)`.

Please also add a regression test that creates `destination/linked -> outsideDirectory`, downloads `linked/escaped.txt`, and verifies the outside file is not created and the object is reported as failed.

lib/lib-transfer-manager/src/submodules/transfer-manager/S3TransferManager.ts
```diff
@@ -1294,11 +1304,228 @@ export class S3TransferManager implements IS3TransferManager {
...
-  public downloadDirectory(
+  public async downloadDirectory(
     request: DownloadDirectoryRequest,
     transferOptions?: TransferOptions
   ): Promise<DownloadDirectoryResponse> {
-    throw new Error("Method not implemented.");
+    const destination = await createDestinationDirectory(request.destination);
+    const maxConcurrency = request.maxConcurrency ?? 64;
+    const failurePolicy = request.failurePolicy ?? ("terminate" as CannedFailurePolicy);
+    const semaphore = new Semaphore(maxConcurrency);
```
trivikr, line 1327: `maxConcurrency <= 0` causes the first semaphore acquisition to wait forever because no task can start and release a permit. Non-integer and non-finite values are also invalid concurrency limits. Validate before creating the destination or making S3 requests:

```suggestion
    const maxConcurrency = request.maxConcurrency ?? 64;
    if (!Number.isInteger(maxConcurrency) || maxConcurrency < 1) {
      throw new Error("maxConcurrency must be a positive integer");
    }

    const destination = await createDestinationDirectory(request.destination);
    const failurePolicy = request.failurePolicy ?? ("terminate" as CannedFailurePolicy);
    const semaphore = new Semaphore(maxConcurrency);
```

Please add a regression test using `maxConcurrency: 0`. An empty mocked listing avoids hanging the test runner while still verifying that the call rejects with a `maxConcurrency` error before any S3 request is sent.

lib/lib-transfer-manager/src/submodules/transfer-manager/S3TransferManager.ts
```diff
@@ -1294,11 +1304,228 @@ export class S3TransferManager implements IS3TransferManager {
...
    * @alpha
    */
-  public downloadDirectory(
+  public async downloadDirectory(
     request: DownloadDirectoryRequest,
     transferOptions?: TransferOptions
   ): Promise<DownloadDirectoryResponse> {
-    throw new Error("Method not implemented.");
+    const destination = await createDestinationDirectory(request.destination);
+    const maxConcurrency = request.maxConcurrency ?? 64;
```
trivikr: Should this be `100` as defined in https://github.com/aws/aws-sdk-js-v3/blob/e53a25aafbdd772c90d26471dc271e383f1daf71/lib/lib-transfer-manager/src/submodules/transfer-manager/types.ts#L518-L522?

```suggestion
    const maxConcurrency = request.maxConcurrency ?? 100;
```

lib/lib-transfer-manager/examples/download-directory.mjs
```diff
@@ -0,0 +1,32 @@
...
+});
+
+const result = await tm.downloadDirectory({
+  bucket: "s3-benchmarkrunner-bucket",
+  destination: "./downloaded",
+  // s3Prefix: "images/",
+  // // Only download .jpg objects.
+  // filter: (object) => object.Key?.endsWith(".jpg") ?? false,
+  // // Keep going if an individual object fails, instead of aborting everything.
+  failurePolicy: "terminate",
```
trivikr: Comment on line says "continue on error" failure policy is selected, but the code choose "terminate"

```suggestion
  failurePolicy: "continue",
```

trivikr (changes requested):

lib/lib-transfer-manager/examples/download-directory.mjs
smilkuri: This files under `examples/` directory was for local testing and tested with different policies between runs. They weren't meant to be committed will remove them.

kuhe (approved):

trivikr (approved):

PR #8278, opened 2026-08-25
fix(cloudfront-signer): preserve query spelling when signing urls

### LINKED ISSUE #8277: cloudfront-signer regression - change in signature from getSignedUrl when query params have spaces

### Describe the bug

In https://github.com/aws/aws-sdk-js-v3/issues/8033 it was reported that there was a regression in that URLs with special characters in the filenames now generated signed URLs which failed verification and errored with `403 AccessDenied` message.

This was correct, and fixed in https://github.com/aws/aws-sdk-js-v3/pull/7763 and released in 3.1056.0 (or possibly earlier but that's the earliest I can download via npm that exhibits the fix).

However, this introduced a second regression in that if the URL parameter has a space (+) in it, such as `response-content-disposition=inline; filename*=filename.pdf;`
or `response-content-disposition=inline; "My file.pdf";` then it breaks and produces a URL which exhibits the same 403 Request denied behaviour as described in the earlier issue.

### Regression Issue

### SDK version number

"@aws-sdk/cloudfront-signer": "3.1056.0"

Node.js

v24.13.0

### Reproduction Steps

This test if added to `packages/cloudfront-signer/src/sign.ts` passes in v3.1034.0 just before the regression and fails afterwards. The only change is in the signature part of it so I'm not sure what the internal issue is.
```
  it("should give me the same signature as it previously did", () => {
    const url = new URL("http://example.com/path/to/file.pdf");
    url.searchParams.set("response-content-disposition", "inline; filename*=filename.pdf");
    const href = url.toString();
    expect(href).toContain("inline%3B+filename*%3Dfilename.pdf");
    expect(href).toBe(
      "http://example.com/path/to/file.pdf?response-content-disposition=inline%3B+filename*%3Dfilename.pdf",
    );
    const result = getSignedUrl({
      url: href,
      keyPairId,
      dateLessThan,
      privateKey,
      passphrase,
    });
    expect(result).toBe(
      "http://example.com/path/to/file.pdf?response-content-disposition=inline%3B%20filename*%3Dfilename.pdf&Expires=1577836800&Key-Pair-Id=APKAEIBAERJR2EXAMPLE&Signature=JBs7DvOuSvgzWJQS~I03~90rFF-yVftF9gNpKd9TrRXrs6oLMITJ7htEnL2RU4XnZWx9-RY5uudGMIOCgkDGSmlmpk2-DXykCq00ZhRdawHHFIMOm~lc-e0ARIxWkcdRvyNBNsGJphyrPZIkcmWKHqlUUIi75gkjKCugcB96T09vjdEQ-rPVE-TYkF5RUjGVRBb8VLZHzC5nQsnf2nLi9-u4mCaq4obNWT1cGsIUJ3I4JbNd8N2nG5O3u0Y10Kye3HRkCstSBgqPX56CZqxFxUxIbcCrHgQUYTzQe60e76oqYPfk3ItRV4Ec3lan07gmO4e~zRp4xKN9wUmZm0sGMg__",
    );
  });
```

### Observed Behavior

I expect signed URLs to still be valid and for the signatures generated given the testcase input to be consistent over time.

### Expected Behavior

We experienced 403 errors and the test case above demonstrates an unexpected change in behaviour

### Possible Solution

I suspect its related to the URL encoding.

_No response_

shin4141:
### Issue
Fixes #8277

### Description
`getSignedUrl` normalized query values before building the canned-policy Resource. For query values containing spaces, this could change the Resource string used for signing even though the emitted URL remained validly encoded.

### Fix
Keep separate internal representations for:

- the encoded URL returned to the caller
- the Resource string used to build the signing policy

Path encoding remains unchanged. The output query continues to use the existing URL encoding, while the signing Resource preserves the input query spelling for raw spaces, `+`, `%20`, and `%2B`.

### Compatibility
The #8277 fixture now exactly matches the signed URL and signature produced by `@aws-sdk/cloudfront-signer` 3.1034.0. The emitted query still uses `%20`, while the signature is again computed from the historical pre-normalization `+` Resource representation.

Focused controls verify that raw space, `+`, `%20`, and `%2B` inputs remain distinct in the signing Resource and produce four distinct signatures.

### Testing
- #8277 exact regression: 3/3 focused runs passed
- canonicalization and selected controls: 8/8 passed
- `cloudfront-signer` native suite: 44/44 passed
- package types, ES/CJS build, Oxfmt, Oxlint, API Extractor, and focused Turbo build passed
- `git diff --check` passed

`yarn generate-clients` was not run because this change does not touch code generation, service models, generated clients, or package manifests; the command regenerates unrelated service-client artifacts.

### Claim boundary
The local cryptographic regression was reproduced and its query-canonicalization cause was proven. This verification did not call live CloudFront, so the reported HTTP 403 was not independently reproduced. Upstream CI and the upstream Node.js matrix have not run yet.

### AI assistance
This contribution was generated with assistance from Codex (an AI tool) and reviewed by Shin Nagata. I reviewed the final diff and remain responsible for the submission.

### Checklist
  - [x] It's not a feature.
  - [x] No streams here.

---

### DIFF
--- a/packages/cloudfront-signer/src/sign.spec.ts
+++ b/packages/cloudfront-signer/src/sign.spec.ts
@@ -973,6 +973,89 @@ describe("url component encoding", () => {
     expect(signedUrl.slice(0, target.length)).toBe(target);
   });
 
+  // https://github.com/aws/aws-sdk-js-v3/issues/8277
+  it("should preserve the last-known-good signature when a query value contains a space", () => {
+    const url = new URL("http://example.com/path/to/file.pdf");
+    url.searchParams.set("response-content-disposition", "inline; filename*=filename.pdf");
+    const href = url.toString();
+    const encodedHref =
+      "http://example.com/path/to/file.pdf?response-content-disposition=inline%3B%20filename*%3Dfilename.pdf";
+
+    expect(href).toBe(
+      "http://example.com/path/to/file.pdf?response-content-disposition=inline%3B+filename*%3Dfilename.pdf"
+    );
+
+    const policyFor = (resource: string) =>
+      JSON.stringify({
+        Statement: [
+          {
+            Resource: resource,
+            Condition: {
+              DateLessThan: {
+                "AWS:EpochTime": epochDateLessThan,
+              },
+            },
+          },
+        ],
+      });
+    const lastKnownGoodSignedUrl =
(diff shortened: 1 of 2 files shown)


rswheeldon (approved):

aBurmeseDev:
This was addressed in https://github.com/aws/aws-sdk-js-v3/pull/8283.

Issue #8113, opened 2026-06-18
CloudFront signed URLs fail when they include `filename*=UTF-8''`


MarquisLP:
### Checkboxes for prior research

- [x] I've gone through [Developer Guide](https://docs.aws.amazon.com/sdk-for-javascript/v3/developer-guide) and [API reference](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest)
- [x] I've checked [AWS Forums](https://forums.aws.amazon.com) and [StackOverflow](https://stackoverflow.com/questions/tagged/aws-sdk-js).
- [x] I've searched for [previous similar issues](https://github.com/aws/aws-sdk-js-v3/issues) and didn't find any solution.

### Describe the bug

As per the [Content-Disposition specification](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Disposition#as_a_response_header_for_the_main_body), a UTF-8 encoded filename can be set for links with the `attachment` disposition by including `filename*=UTF-8''{encodedFilename}` in the value for `response-content-disposition` in the query string.

However, when this addition is made to a URL and then passed to cloudfront-signer's `getSignedUrl()`, the resulting signed URL fails on load with an Access Denied error.

### Regression Issue

- [x] Select this option if this issue appears to be a regression.

### SDK version number

@aws-sdk/cloudfront-signer@3.1046.0

### Which JavaScript Runtime is this issue in?

Node.js

### Details of the browser/Node.js/ReactNative version

v24.4.1

### Reproduction Steps

Generate a signed URL that includes `filename*`:

```javascript
const signer = require('@aws-sdk/cloudfront-signer');

const filename = '文件';
const normalizedFilename = normalizeHelperFunction(filename);
const encodedFilename = encodeURIComponent(filename);

const signedUrl = signer.getSignedUrl({
    url: `https://example.com/path?response-content-disposition=attachment;filename=${normalizedFilename};filename*=UTF-8''${encodedFilename}`,
});

return signedUrl;
```

Take that signed URL and attempt to load it in a web browser.

### Observed Behavior

The signed URL redirects to an error page with an Access Denied error.

### Expected Behavior

The signed URL loads the requested file for download.

### Possible Solution

Add special handling of the `UTF-8''` part of the `filename*` option so that apostrophes are allowed to be included unencoded there. (See the Additional Information/Context below.)

### Additional Information/Context

There was a similar issue in the past: #7470.

[...]

PR #8282, opened 2026-08-27
fix(cloudfront-signer): preserve plus in query strings when signing URLs

aBurmeseDev:
### Issue
https://github.com/aws/aws-sdk-js-v3/issues/8277

### Description
`encodeUrlPath` [re-encoded](https://github.com/aws/aws-sdk-js-v3/blob/ba4e4498a7610ec0b5ad7af195db137f618e09bc/packages/cloudfront-signer/src/sign.ts#L316) query strings via `new URLSearchParams()` which applies form-encoding semantics where `+` means space, so `+` came back out as `%20`.

Since #8039, the policy `Resource` is built from that re-encoded URL so the signature covers `%20` when the caller asked for `+` and CloudFront 403s. Shows up most often with `response-content-disposition`, since `URL.searchParams.set()` emits `+` for spaces.

This PR replaces the `URLSearchParams` round-trip with `encodeQueryComponent` which splits on existing `%XX` escapes and `+`, passes those through untouched, and runs `encodeURIComponent` on the rest. Single quotes still become `%27` for #8113 and preserved hex is still uppercased which the old decode/re-encode did that implicitly. That keeps `+` the only behavior change here.

### Testing
- all unit tests pass, plus 2 new ones that check the URL output and verify the signature against explicit policy                             
- new e2e that fetches a signed URL with `+` in query from a live distribution                                                                                                                                  
- confirmed lowercase `%2b` still normalizes to `%2B`

---

### DIFF
--- a/packages/cloudfront-signer/src/cloudfront-signer.e2e.spec.ts
+++ b/packages/cloudfront-signer/src/cloudfront-signer.e2e.spec.ts
@@ -308,4 +308,23 @@ describe("@aws-sdk/cloudfront-signer e2e", () => {
     expect(response.status).toBe(200);
     expect(await response.text()).toBe(OBJECT_BODY);
   });
+
+  // https://github.com/aws/aws-sdk-js-v3/issues/8277
+  it("should access content when query contains + (space encoded as +)", async () => {
+    const url = new URL(`https://${domain}/${OBJECT_KEY}`);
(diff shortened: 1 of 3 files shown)


aBurmeseDev:
Reopened from upstream branch to fix failing CI.

PR #8283, opened 2026-08-27
fix(cloudfront-signer): preserve plus in query strings when signing URLs

aBurmeseDev:
### Issue
https://github.com/aws/aws-sdk-js-v3/issues/8277

### Description
`encodeUrlPath` [re-encoded](https://github.com/aws/aws-sdk-js-v3/blob/ba4e4498a7610ec0b5ad7af195db137f618e09bc/packages/cloudfront-signer/src/sign.ts#L316) query strings via `new URLSearchParams()` which applies form-encoding semantics where `+` means space, so `+` came back out as `%20`.

Since #8039, the policy `Resource` is built from that re-encoded URL so the signature covers `%20` when the caller asked for `+` and CloudFront 403s. Shows up most often with `response-content-disposition`, since `URL.searchParams.set()` emits `+` for spaces.

This PR replaces the `URLSearchParams` round-trip with `encodeQueryComponent` which splits on existing `%XX` escapes and `+`, passes those through untouched, and runs `encodeURIComponent` on the rest. Single quotes still become `%27` for #8113 and preserved hex is still uppercased which the old decode/re-encode did that implicitly. That keeps `+` the only behavior change here.

### Testing
- all unit tests pass, plus 2 new ones that check the URL output and verify the signature against explicit policy                             
- new e2e that fetches a signed URL with `+` in query from a live distribution                                                                                                                                  
- confirmed lowercase `%2b` still normalizes to `%2B`

---

### DIFF
--- a/packages/cloudfront-signer/src/sign.spec.ts
+++ b/packages/cloudfront-signer/src/sign.spec.ts
@@ -973,6 +973,115 @@ describe("url component encoding", () => {
     expect(signedUrl.slice(0, target.length)).toBe(target);
   });
 
+  // https://github.com/aws/aws-sdk-js-v3/issues/8277
+  it("should preserve + in query strings without converting to %20", () => {
+    const url = new URL("http://example.com/path/to/file.pdf");
+    url.searchParams.set("response-content-disposition", "inline; filename*=filename.pdf");
+    const href = url.toString();
+    // URL.searchParams.set encodes space as +
+    expect(href).toContain("inline%3B+filename*%3Dfilename.pdf");
+
+    const signedUrl = getSignedUrl({
+      url: href,
+      keyPairId,
+      privateKey,
+      passphrase,
+      dateLessThan: "2026-01-01",
+    });
+
+    // The + must be preserved in the signed URL, not converted to %20
+    const cfParamStart = signedUrl.search(/[?&]Expires=/);
+    expect(cfParamStart).toBeGreaterThan(0);
+    const baseUrl = signedUrl.slice(0, cfParamStart);
+    expect(baseUrl).toContain("inline%3B+filename*%3Dfilename.pdf");
+    expect(baseUrl).not.toContain("%20");
+
+    // Verify signature matches a policy over the original URL (with +)
+    const parsedUrl = parseUrl(signedUrl);
+    const signature = denormalizeBase64(parsedUrl.query!["Signature"] as string);
+    const policyStr = JSON.stringify({
+      Statement: [
+        {
+          Resource: href,
+          Condition: {
+            DateLessThan: {
+              "AWS:EpochTime": Math.round(new Date("2026-01-01").getTime() / 1000),
+            },
+          },
+        },
+      ],
+    });
+    expect(verifySignature(signature, policyStr)).toBeTruthy();
+  });
+
+  // https://github.com/aws/aws-sdk-js-v3/issues/8277
+  it("should preserve raw + as literal in query strings", () => {
+    const inputUrl = "http://example.com/file.pdf?q=a+b";
+    const signedUrl = getSignedUrl({
+      url: inputUrl,
+      keyPairId,
+      privateKey,
+      passphrase,
+      dateLessThan: "2026-01-01",
+    });
+
+    const cfParamStart = signedUrl.search(/[?&]Expires=/);
+    expect(cfParamStart).toBeGreaterThan(0);
+    const baseUrl = signedUrl.slice(0, cfParamStart);
+    expect(baseUrl).toBe("http://example.com/file.pdf?q=a+b");
+
+    // Verify signature matches a policy over the original URL (with +)
+    const parsedUrl = parseUrl(signedUrl);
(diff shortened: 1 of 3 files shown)


packages/cloudfront-signer/src/sign.ts
```diff
@@ -307,22 +307,54 @@ function encodeUrlPath(url: string): string {
...
+        return `${encodeQueryComponent(rawKey)}=${encodeQueryComponent(rawValue)}`;
+      });
+    encodedQuery = pairs.length > 0 ? `?${pairs.join("&")}` : "";
   }
 
   return origin + encodedPath + encodedQuery;
 }
 
+/**
+ * Percent-encodes a query component (key or value) preserving `+` literally
```
siddsriv: nit: succinct-ify this comment block if possible? personally, i want to see the rationale for things in comments more than the related issue history (which should be trackable from `blame`)

packages/cloudfront-signer/src/sign.spec.ts
```diff
@@ -973,6 +973,81 @@ describe("url component encoding", () => {
     expect(signedUrl.slice(0, target.length)).toBe(target);
   });
 
+  // https://github.com/aws/aws-sdk-js-v3/issues/8277
```
siddsriv, line 976: should we also have a test for a literal (unencoded) `%` in the query? another test case we could have is to test the hex-uppercasing normalization explicitly -- these should not be blocking i think

packages/cloudfront-signer/src/sign.ts
aBurmeseDev: trimmed it to one line

packages/cloudfront-signer/src/sign.spec.ts
aBurmeseDev: added both test cases

packages/cloudfront-signer/src/sign.ts
```diff
@@ -307,22 +307,43 @@ function encodeUrlPath(url: string): string {
...
   return origin + encodedPath + encodedQuery;
 }
 
+/**
+ * Percent-encodes a query component preserving `+` literally and uppercasing
+ * existing %xx hex. Single quotes become %27 for CloudFront.
+ * @internal
+ */
+function encodeQueryComponent(component: string): string {
+  return component
```
trivikr: `encodeURIComponent` throws on lone UTF-16 surrogates; `URLSearchParams` previously normalized them to U+FFFD.

```suggestion
  const wellFormedComponent = component.replace(/[\uD800-\uDFFF]/gu, "\uFFFD");
  return wellFormedComponent
```

Please add `it.each` cases for `\uD800`, `\uDC00`, and `\uD83D\uDE00`, expecting `%EF%BF%BD`, `%EF%BF%BD`, and `%F0%9F%98%80`, and verify each signature against the normalized policy `Resource`.

trivikr (changes requested):

packages/cloudfront-signer/src/sign.ts
```diff
@@ -307,22 +307,43 @@ function encodeUrlPath(url: string): string {
...
+    .split(/(%[0-9A-Fa-f]{2}|\+)/g)
```
trivikr, line 337: This is for reference. No action needed.

This also preserves all existing `%XX` escapes, so inputs such as `%41` and `%FF` no longer normalize to `A` and `%EF%BF%BD`. The live CloudFront E2E accepts both preserved forms, so this is not a correctness issue, but it does change generated URLs and signatures beyond the stated `+` fix.

I verified the behavior in an E2E test
```js
  it.each([
    ["percent-encoded unreserved characters", "%41%7e", "%41%7E"],
    ["malformed UTF-8", "%FF", "%FF"],
  ])("should access content with %s", async (_caseName, inputQuery, expectedQuery) => {
    const expectedBaseUrl = `https://${domain}/${OBJECT_KEY}?q=${expectedQuery}`;
    const signedUrl = getSignedUrl({
      url: `https://${domain}/${OBJECT_KEY}?q=${inputQuery}`,
      keyPairId,
      privateKey,
      dateLessThan: new Date(Date.now() + 60_000).toISOString(),
    });

    const cfParamStart = signedUrl.search(/[?&]Expires=/);
    expect(cfParamStart).toBeGreaterThan(0);
    expect(signedUrl.slice(0, cfParamStart)).toBe(expectedBaseUrl);

    const response = await fetch(signedUrl);
    expect(response.status).toBe(200);
    expect(await response.text()).toBe(OBJECT_BODY);
  });
```

trivikr: Resolving since this comment was just for reference.

packages/cloudfront-signer/src/sign.ts
aBurmeseDev: added lone surrogate normalization to U+FFFD before `encodeURIComponent`, and `it.each` cases for `\uD800`, `\uDC00`, and `\uD83D\uDE00` with signature verification against the encoded URL.

trivikr (approved):

PR #8285, opened 2026-09-02
fix(lib-transfer-manager): retry worker path download part failures

smilkuri:
### Description
In the worker download path, the worker writes response bytes directly to the file offset (or assembles them into the transfer buffer) and the response body does not go through the SDK middleware. So when S3 returns a non-2xx like 503 SlowDown, the worker was writing the small XML error body (SlowDown...) to the file as if it were object data. The download then failed downstream with a "part size doesn't match expected part size" error, and because the failure never reached the retry middleware, the transient error was not retried.

This change makes the worker check the response status before touching the file/buffer: on a non-2xx, instead of writing the body, it reads the error body and sends the raw status code, headers, and body back to the main thread. The main thread returns these to the SDK as a normal HttpResponse, so the SDK's own deserializer builds the correct typed error (e.g. SlowDown) with $metadata/$fault, and the retry middleware classifies and retries it.

### Testing
CI 

### Checklist
  - [x] It's not a feature.
  - [x] No streams here.

---

### DIFF
--- a/lib/lib-transfer-manager/src/submodules/worker/http-request-worker.ts
+++ b/lib/lib-transfer-manager/src/submodules/worker/http-request-worker.ts
@@ -28,6 +28,55 @@ const DNS_TTL_MS = 1000;
 const dnsCache = new Map<string, { ips: string[]; ts: number }>();
 
 /**
+ * Drain a small error-response body to text so its S3 error <Code> can be
+ * surfaced. Bounded to avoid buffering large bodies.
+ * @internal
+ */
+const readErrorBody = async (body: any): Promise<string | undefined> => {
+  if (!body) return undefined;
+  try {
+    const chunks: Buffer[] = [];
+    let total = 0;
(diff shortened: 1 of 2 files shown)


lib/lib-transfer-manager/src/submodules/worker/http-request-worker.ts
```diff
@@ -27,6 +27,48 @@ import { parentPort } from "node:worker_threads";
...
+  try {
+    const chunks: Buffer[] = [];
+    let total = 0;
+    for await (const chunk of body) {
+      const buf = typeof chunk === "string" ? Buffer.from(chunk) : Buffer.from(chunk);
+      chunks.push(buf);
+      total += buf.length;
+      if (total >= MAX_ERROR_BODY_BYTES) break;
+    }
+    return new Uint8Array(Buffer.concat(chunks));
```
kuhe: use concatBytes from smithy/core/serde

PR #8291, opened 2026-09-03
chore(lib-transfer-manager): add README file

smilkuri:
### Description
Adds a README for `@aws-sdk/lib-transfer-manager` covering the S3TransferManager overview and features, developer-preview notice, getting started (install and instantiation), and usage examples.

---

### DIFF
--- a/lib/lib-transfer-manager/README.md
+++ b/lib/lib-transfer-manager/README.md
@@ -1,0 +1,284 @@
+# @aws-sdk/lib-transfer-manager
+
+[![NPM version](https://img.shields.io/npm/v/@aws-sdk/lib-transfer-manager/latest.svg)](https://www.npmjs.com/package/@aws-sdk/lib-transfer-manager)
+[![NPM downloads](https://img.shields.io/npm/dm/@aws-sdk/lib-transfer-manager.svg)](https://www.npmjs.com/package/@aws-sdk/lib-transfer-manager)
+
+## Overview
+
+The S3 Transfer Manager is a high-level transfer utility built on top of the S3 client. It provides a simple API to allow you to transfer files and directories between your application and Amazon S3. The S3 Transfer Manager also enables you to monitor a transfer's progress in real-time.
+
+The following transfer operations are supported:
+
+- automatic [multipart upload](#upload-an-object) to S3
+- automatic [multipart download](#download-an-object) from S3
+- upload all files in a directory to an S3 bucket recursively or non-recursively (see [upload a directory](#upload-a-directory))
+- download all objects in a bucket to a local directory recursively or non-recursively (see [download a directory](#download-a-directory))
+- transfer progress listener (see [monitor transfer progress](#monitor-transfer-progress))
+
+## ⚠️ Developer Preview
+
+This library is currently in developer preview and is **NOT** recommended for production environments. It is meant for early access and feedback purposes at this time. We'd love to hear your feedback. See the AWS SDK and Tools [maintenance policy](https://docs.aws.amazon.com/sdkref/latest/guide/maint-policy.html#version-life-cycle)
+descriptions for more information.
+
+## Getting Started
+
+### Installation
+
+```
+npm install @aws-sdk/lib-transfer-manager
+```
+
+### Instantiate the S3 Transfer Manager
+
+You can instantiate the transfer manager easily using the default settings:
+
+```js
+import { S3TransferManager } from "@aws-sdk/lib-transfer-manager";
+
+const tm = new S3TransferManager();
+```
+
+Or pass a config to customize it, including your own `S3Client`:
+
+```js
+import { S3Client } from "@aws-sdk/client-s3";
+import { S3TransferManager } from "@aws-sdk/lib-transfer-manager";
+
+const tm = new S3TransferManager({
+  s3: new S3Client({ region: "us-west-2" }),
+  targetPartSizeBytes: 16 * 1024 * 1024, // 16 MB part size
+  maxConcurrentUploads: 256,
+});
+```
+
+For the full list of configuration options, see the documentation comments on
+[`S3TransferManagerConfig`](./src/submodules/transfer-manager/types.ts).
+
+## Usage Examples
+
+### Upload an object
+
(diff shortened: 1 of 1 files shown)


lib/lib-transfer-manager/README.md
```diff
@@ -0,0 +1,284 @@
+# @aws-sdk/lib-transfer-manager
+
+[![NPM version](https://img.shields.io/npm/v/@aws-sdk/lib-transfer-manager/latest.svg)](https://www.npmjs.com/package/@aws-sdk/lib-transfer-manager)
+[![NPM downloads](https://img.shields.io/npm/dm/@aws-sdk/lib-transfer-manager.svg)](https://www.npmjs.com/package/@aws-sdk/lib-transfer-manager)
+
+## Overview
+
+The S3 Transfer Manager is a high-level transfer utility built on top of the S3 client. It provides a simple API to allow you to transfer files and directories between your application and Amazon S3. The S3 Transfer Manager also enables you to monitor a transfer's progress in real-time.
```
kuhe: "transfer utility" -> "data transfer utility"

kuhe: "provides a simple API" -> "provides an API".

Also, isn't the product name "Transfer Manager for S3"?

kuhe: needs a line describing what the added value of using this is.

For example (modify as needed):

> The Transfer Manager internalizes multi-part data, parallel transfers, uses worker threads when possible, and optimizes DNS resolution to S3.

lib/lib-transfer-manager/README.md
```diff
@@ -0,0 +1,284 @@
...
+
+The following transfer operations are supported:
+
+- automatic [multipart upload](#upload-an-object) to S3
```
kuhe: "automatic" doesn't add much meaning here. Seamless transition between PutObject and a combined CreateMultipartUpload + [UploadPart, ...] + CompleteMultipartUpload.

lib/lib-transfer-manager/README.md
```diff
@@ -0,0 +1,284 @@
...
+- automatic [multipart download](#download-an-object) from S3
```
kuhe: what is "automatic" for the download side? Automatic parallel downloads of multiple objects or large objects by part or byterange.

lib/lib-transfer-manager/README.md
```diff
@@ -0,0 +1,284 @@
...
+- upload all files in a directory to an S3 bucket recursively or non-recursively (see [upload a directory](#upload-a-directory))
+- download all objects in a bucket to a local directory recursively or non-recursively (see [download a directory](#download-a-directory))
+- transfer progress listener (see [monitor transfer progress](#monitor-transfer-progress))
+
+## ⚠️ Developer Preview
+
+This library is currently in developer preview and is **NOT** recommended for production environments. It is meant for early access and feedback purposes at this time. We'd love to hear your feedback. See the AWS SDK and Tools [maintenance policy](https://docs.aws.amazon.com/sdkref/latest/guide/maint-policy.html#version-life-cycle)
```
kuhe, line 20: I don't think this is practical. The SDK overall is GA, the version of this pkg is in 3.x, and we are still responsible if someone uses it in production, even with this disclaimer. 

We should simply release to GA directly, when published.

lib/lib-transfer-manager/README.md
```diff
@@ -0,0 +1,284 @@
...
+
+### Installation
+
+```
+npm install @aws-sdk/lib-transfer-manager
+```
+
+### Instantiate the S3 Transfer Manager
+
+You can instantiate the transfer manager easily using the default settings:
```
kuhe:
```suggestion
You can instantiate the transfer manager using the default settings:
```

lib/lib-transfer-manager/README.md
```diff
@@ -0,0 +1,284 @@
...
+  failIfExists: true, // throws if ./downloads/some-key already exists (default: false, overwrite)
+});
+
+console.log(`Wrote ${result.bytesWritten} bytes`);
+```
+
+### Upload a directory
+
+Provide the destination `bucket` and the local `source` directory. By default the directory is not
+traversed recursively; set `recursive: true` to include subdirectories.
```
kuhe: directory separators are considered part of the S3 Object Key.

lib/lib-transfer-manager/README.md
```diff
@@ -0,0 +1,284 @@
...
+  source: "source/directory/to/upload",
+  recursive: true,
+  failurePolicy: async ({ objectRequest, error }) => {
+    console.warn(`Failed ${objectRequest.Key}: ${error}`);
+    return "continue";
+  },
+});
+```
+
+### Download a directory
```
kuhe: The method name may be required to be `downloadDirectory` by spec, but in all documentation language, it should be clear that you are downloading a set of Objects whose Keys match a Key prefix _into_ a directory, treating certain Key delimiters as directory markers, and never "downloading a directory". 


```suggestion
### Downloading Objects to a directory, mapping Key-delimiter `/` to subdirectories
```

kuhe (changes requested):

lib/lib-transfer-manager/README.md
smilkuri: It's referenced as S3 Transfer Manager everywhere but I agree according to recent discussion about marketing rules it should be "Transfer Manager for S3". Will update it.

lib/lib-transfer-manager/README.md
smilkuri: Let's discuss about this with the team.

lib/lib-transfer-manager/README.md
smilkuri: I kept the heading short as is and instead updated the description to make it clear.

smilkuri: Let me know if you have preference to update the heading.

kuhe: it should say: "Downloading Objects to a directory" or at least "Downloading to a directory".

kuhe (approved):

PR #8292, opened 2026-09-03
fix(lib-transfer-manager): update transfer progress events

smilkuri:
### Issue
- On multipart **downloads**, per-part fetches dispatched `bytesTransferred` events whose `transferredBytes` was a single part's size against the whole-object total, interleaving small per-part percentages with the cumulative stream-consumption values (e.g. `98% ... 0.2% ... 98%`).
- On **directory** transfers, the per-file `upload` / `downloadToFile` calls emitted single-object transfer events to listeners, mixed in with the aggregated directory-level snapshots.
  

### Description
- Report `bytesTransferred` as a cumulative running total. In the worker-based `download()` paths, drop the redundant per-part dispatch (progress already comes from `joinStreams` as the caller consumes the body). In `downloadToFile`, accumulate a running total across parts/ranges.
- Add an internal `emitEvents` flag (default `true`) to `upload`/`downloadToFile`. `uploadDirectory` and `downloadDirectory` pass `false` so only the aggregated directory-level snapshots are emitted.

### Testing
Added e2e tests.
```
$ yarn test:e2e
 Test Files  1 passed (1)
      Tests  44 passed (44)
   Start at  15:35:36
   Duration  43.66s (transform 405ms, setup 0ms, import 618ms, tests 42.91s, environment 0ms)
  ```

### Checklist
  - [x] It's not a feature.
  - [x] No streams here.

---

### CODE REVIEW

File: lib/lib-transfer-manager/src/submodules/transfer-manager/S3TransferManager.ts [hunk 1]
```diff
@@ -313,16 +313,22 @@ export class S3TransferManager implements IS3TransferManager {
    * @returns S3 PutObject or CompleteMultipartUpload response with transfer event dispatching.
    *
    */
-  public async upload(request: UploadRequest, transferOptions?: TransferOptions): Promise<UploadResponse> {
+  public async upload(
+    request: UploadRequest,
+    transferOptions?: TransferOptions,
+    emitEvents = true
```
kuhe, line 319: PR description says this flag is internal, why does it appear in the public API?

kuhe: it's missing from the tsdoc. who calls it with false (non-default)?

smilkuri: It shouldn't, will move this into a private method and will pass emitEvents flag there. Directory transfer method calls it with false.

kuhe (approved):

PR #8308, opened 2026-09-17
fix(recursion-detection): propagate only W3C baggage from InvokeStore

maxday:
Drop the traceparent/tracestate branch of the InvokeStore-driven W3C propagation added in #8156. 
Trace context (traceparent/tracestate) is delegated to the OTel SDK / user-provided propagators; the middleware now only forwards the W3C baggage header from InvokeStore.getBaggage().

- Middleware: remove TRACEPARENT/TRACESTATE constants and their block; propagate baggage independently, gated on an existing baggage header rather than on traceparent (symmetric to the removed logic).
- sanitizeTraceHeaders: only canonicalize casing of baggage.
- configuration.ts: rename the middleware tag TRACE_CONTEXT_PROPAGATION to BAGGAGE_PROPAGATION.
- Update unit and integ specs to cover baggage-only behavior; delete traceparent/tracestate assertions.

---


luben (approved):
LGTM, to add more context: The AWS SDK should care only about propagating Lineage for the recursion protection (and X-Ray header for the same reason). W3C trace context propagation should be delegate to the customer code, OTEL instrumentation, or the Lambda runtime.

This is not a breaking change as today W3C headers are not yet made available in Lambda and the `traceparent`/`tracestate` were recently added, but we reconsidered it.

